PR #1, opened 2016-08-27
Perf: Add volume name label

carlpett:
As per the TODO list on the wiki, I added volume labels to the perf collector. I also added a whitelist/blacklist functionality similar to what is used in other exporters, defaulting to excluding the `_Total` "meta-counter".

By the way, shouldn't the collector be named something like logical_disk rather than perf?

### DIFF
--- a/collectors/perf.go
+++ b/collectors/perf.go
@@ -4,10 +4,18 @@
 package collectors
 
 import (
+	"flag"
+	"fmt"
 	"log"
+	"regexp"
 
 	"github.com/StackExchange/wmi"
 	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	volumeWhitelist = flag.String("collector.perf.volume-whitelist", ".+", "Regexp of volumes to whitelist. Volume name must both match whitelist and not match blacklist to be included.")
+	volumeBlacklist = flag.String("collector.perf.volume-blacklist", "_Total", "Regexp of volumes to blacklist. Volume name must both match whitelist and not match blacklist to be included.")
 )
 
 // A PerfCollector is a Prometheus collector for WMI Win32_PerfRawData_PerfDisk_LogicalDisk metrics
@@ -46,6 +54,9 @@ type PerfCollector struct {
 	PercentIdleTime              *prometheus.Desc
 	PercentIdleTime_Base         *prometheus.Desc
 	SplitIOPerSec                *prometheus.Desc
+
+	volumeWhitelistPattern *regexp.Regexp
+	volumeBlacklistPattern *regexp.Regexp
 }
 
 // NewPerfCollector ...
@@ -289,6 +300,9 @@ func NewPerfCollector() *PerfCollector {
 			[]string{"volume"},
 			nil,
 		),
+
+		volumeWhitelistPattern: regexp.MustCompile(fmt.Sprintf("^(?:%s)$", *volumeWhitelist)),
+		volumeBlacklistPattern: regexp.MustCompile(fmt.Sprintf("^(?:%s)$", *volumeBlacklist)),
 	}
 }
 
@@ -387,6 +401,10 @@ func (c *PerfCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, 
 	}
 
 	for _, volume := range dst {
+		if c.volumeBlacklistPattern.MatchString(volume.Name) || !c.volumeWhitelistPattern.MatchString(volume.Name) {
+			continue
+		}
+
 		ch <- prometheus.MustNewConstMetric(
 			c.AvgDiskBytesPerRead,
 			prometheus.GaugeValue,
--- a/collectors/perf.go
+++ b/collectors/perf.go
@@ -55,238 +55,238 @@ func NewPerfCollector() *PerfCollector {
 		AvgDiskBytesPerRead: prometheus.NewDesc(
 			prometheus.BuildFQName(wmiNamespace, "perf", "avg_disk_bytes_per_read"),
 			"AvgDiskBytesPerRead",
-			nil,
+			[]string{"volume"},
 			nil,
 		),
 
 		AvgDiskBytesPerRead_Base: prometheus.NewDesc(
 			prometheus.BuildFQName(wmiNamespace, "perf", "avg_disk_bytes_per_read_base"),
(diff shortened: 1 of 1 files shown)


collectors/perf.go
```diff
@@ -4,12 +4,20 @@
...
 	"log"
+	"regexp"
 
 	"github.com/StackExchange/wmi"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+var (
+	volumeWhitelist = flag.String("collector.perf.volume-whitelist", ".+", "Regexp of volumes to whitelist. Volume name must both match whitelist and not match blacklist to be included.")
+	volumeBlacklist = flag.String("collector.perf.volume-blacklist", "_Total", "Regexp of volumes to blacklist. Volume name must both match whitelist and not match blacklist to be included.")
```
brian-brazil: I'd hardcode the _Total in the code itself, so that a user doesn't have to remember to explicitly exclude it.

collectors/perf.go
```diff
@@ -55,240 +66,243 @@ func NewPerfCollector() *PerfCollector {
...
 			prometheus.BuildFQName(wmiNamespace, "perf", "current_disk_queue_length"),
 			"CurrentDiskQueueLength",
-			nil,
+			[]string{"volume"},
 			nil,
 		),
 
 		DiskBytesPerSec: prometheus.NewDesc(
 			prometheus.BuildFQName(wmiNamespace, "perf", "disk_bytes_per_sec"),
 			"DiskBytesPerSec",
```
brian-brazil: These are all gauges, do raw counters exist that we could use instead?

collectors/perf.go
```diff
@@ -55,240 +66,243 @@ func NewPerfCollector() *PerfCollector {
...
 			prometheus.BuildFQName(wmiNamespace, "perf", "free_megabytes"), // XXX in bytes
 			"FreeMegabytes",
-			nil,
+			[]string{"volume"},
 			nil,
 		),
 
 		PercentDiskReadTime: prometheus.NewDesc(
 			prometheus.BuildFQName(wmiNamespace, "perf", "percent_disk_read_time"),
 			"PercentDiskReadTime",
```
brian-brazil: This should be converted to a ratio

collectors/perf.go
carlpett: Good point. If the user actually wants a total, then they can use `sum()`

collectors/perf.go
carlpett: Not from what I can find, at least. Most of the stuff that can be extracted from WMI is pre-aggregated in some way (some classes _only_ export running averages and percentages). I'll dig around a bit, but it doesn't look too good.

carlpett:
I started looking a bit more carefully at the data we're getting here, and after doing a bit more reading on the `PerfRaw*` classes (https://msdn.microsoft.com/en-us/windows/hardware/aa392761(v=vs.71)#how_to_interpret_property_qualifiers), it seems there needs to be a bit of post-processing involving the value and the "base" value. 
That will probably require a fair bit of refactoring of the collector, so maybe we should take that outside the scope of this PR (which was just about adding a label)?

carlpett:
After yet another look, the `_base` stuff only applies to the `avg*` metrics, and we shouldn't expose those anyway, right?

martinlindhe:
@carlpett: This is great!
So, will this expose all disks by default? I see you have added some configuration options.

I was also looking into this yesterday but ran out of time. My approach was to build the measure name from the drive letter, something like "perfc_disk_bytes_per_sec", "perfd_disk_bytes_per_sec" etc. How would one differentiate the measures with this patch?

Got to spin up a vm and try it out.
The "_Total" disk should not be exported.
Btw, patch greatly accepted. I agree we should go throughj the "perf" measures in detail in upcoming commits:
* several data points are probably unusable
* some need to be converted to base units (megabytes to bytes, etc)
* several should get names more in line with prometheus naming

martinlindhe:
Also, the name "perf" just came from "Win32_PerfRawData_PerfDisk_LogicalDisk", and was short. "logical_disk" is pretty long. perhaps "io" is better?

martinlindhe:
to summarize, additional points to be fixed mentioned here (also added to the wiki)
perf: hardcode "_Total" for always exlusion
perf: FreeMegabytes should be a ratio
perf: dont export Avg* and *_Base
 thanks @brian-brazil & @carlpett

carlpett:
Well yes, sort of :) This will expose all _volumes_ (which maps to partitions unless you use Windows Logical volumes/software raid). Here's a sample from my machine: 
```
wmi_perf_percent_free_space{volume="C:"} 62669
wmi_perf_percent_free_space{volume="D:"} 99522
wmi_perf_percent_free_space{volume="HarddiskVolume1"} 304
wmi_perf_percent_free_space{volume="HarddiskVolume3"} 95
```
`HarddiskVolume1` is the "System reserved" partition for recovery tools in my case, which is not mounted to a drive letter, while `HarddiskVolume3` is a secondary disk which I haven't had time to set up yet... I haven't tested drives mounted to a subpath (eg mounting a volume to d:\some\mountpoint), but I think that will still be called HarddiskVolumeX.
So it does not map to disks, as such. There is a PhysicalDisk family of counters for those (but then you don't get any mapping to drive labels, on the other hand).

I've got a fix for `_Total` already which I meant to add to the PR. Will submit as a separate patch in a minute.

martinlindhe:
Okay, yes it looks better than what I had in mind :)

brian-brazil:
On Unix anyway, the equivalent would only export mounted filesystems. I'd guess Windows should be the same.

How do you have a percentage free space over 100%?

carlpett:
I tested the subpath mount, and it turns out well:
```
wmi_logical_disk_size_bytes{volume="C:"} 1.365508096e+11
wmi_logical_disk_size_bytes{volume="D:"} 1.37435807744e+11
wmi_logical_disk_size_bytes{volume="D:\\Test-Mountpoint"} 6.0018393088e+10
```

@brian-brazil This should mean that we could by default filter out anything like `HarddiskVolume.+`. As far as I can tell from looking at the raw data, the `Name` property is the only thing that changes when a volume is mounted vs non-mounted.

As for the strange data in the initial post, that was before any of the other work done in the #5, so it just contained the raw data from WMI.

Issue #4, opened 2016-08-27
Clean up exposed metrics from LogicalDisk


carlpett:
Ensure correct units, remove unusable metrics etc. PR in progress.

martinlindhe:
:tada:

PR #5, opened 2016-08-27
Cleanup disk metrics

carlpett:
Addresses #4 by:
- Removing the pre-averaged metrics
- Removing metrics that aggregate read and write into a single metric (since those are exposed separately)
- Converting free_megabytes to free_bytes
- Removing the *_base metrics, and fixing the percent-metrics

Because of how the percent metrics are calculated (`percent_metric_name / percent_metric_name_base`), they get "fake precision" when getting close to zero, eg `wmi_perf_percent_disk_write_time{volume="HarddiskVolume3"} 2.931438995229291e-11`.
Possibly we should truncate to zero at some point? @brian-brazil - any recommendations here?


carlpett:
Also, just re-read the recommendations for writing exporters and about not exposing percentages. Fix coming.

martinlindhe:
Looks great. Waiting for @brian-brazil to comment before merging

carlpett:
I did some more cleanup, primarily cosmetic stuff on the metric names, as well as updating the help texts from `perfmon`.
I also changed from percentages to ratios. I'm not fully satisfied with the names, but not sure how to make them better. `idle_time` doesn't sound like a ratio, I'd expect seconds. Any ideas?

carlpett:
It also seems that some of the data we get might actually be counters rather than gauges - the ones with CounterType `272696320` (eg `reads_per_sec`) in the documentation seem to be strictly increasing, despite the help text indicating that it is a rate...

martinlindhe:
Regarding help texts, I think it's a good idea to keep the WMI names in there too, to make it meaningful for users to understand what the measure is derived from.

From the prometheus docs:
```
When you’re transforming metrics it’s useful for users to be able to track back to what the original was, and what rules were in play that caused that transform. Putting in the name of the collector/exporter, the id of any rule that was applied and the name/details of the original metric into the help string will greatly aid users.
```
https://prometheus.io/docs/instrumenting/writing_exporters/#help-strings

carlpett:
Yes, agreed. What do you think about this format? 
> The number of requests outstanding on the disk (derived from LogicalDisk.CurrentDiskQueueLength)

martinlindhe:
That would be great, also "The number of requests outstanding on the disk (LogicalDisk.CurrentDiskQueueLength)" works :)

carlpett:
Done!

martinlindhe:
By the way, feel free to rename perf.go to logical_disk.go in this PR

collectors/perf.go
```diff
@@ -63,240 +40,79 @@ type PerfCollector struct {
...
 			[]string{"volume"},
 			nil,
 		),
 
-		AvgDiskBytesPerTransfer: prometheus.NewDesc(
-			prometheus.BuildFQName(wmiNamespace, "perf", "avg_disk_bytes_per_transfer"),
-			"AvgDiskBytesPerTransfer",
+		ReadsPerSec: prometheus.NewDesc(
+			prometheus.BuildFQName(wmiNamespace, "perf", "reads_per_sec"),
+			"The rate of read operations on the disk (LogicalDisk.DiskReadsPerSec)",
```
brian-brazil: Over what time period?

collectors/perf.go
```diff
@@ -63,240 +40,79 @@ type PerfCollector struct {
...
+			"Fraction of time that the disk was busy servicing write requests (LogicalDisk.PercentDiskWriteTime)",
 			[]string{"volume"},
 			nil,
 		),
 
-		AvgDiskSecPerRead: prometheus.NewDesc(
-			prometheus.BuildFQName(wmiNamespace, "perf", "avg_disk_sec_per_read"),
-			"AvgDiskSecPerRead",
+		FreeSpace: prometheus.NewDesc(
+			prometheus.BuildFQName(wmiNamespace, "perf", "free_space"),
```
brian-brazil: free_space_ratio

This is to distinguish from what would be free_space_bytes.

collectors/perf.go
```diff
@@ -280,70 +224,28 @@ func (c *PerfCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc,
...
 			prometheus.GaugeValue,
-			float64(volume.PercentDiskWriteTime),
-			volume.Name,
-		)
-
-		ch <- prometheus.MustNewConstMetric(
-			c.PercentDiskWriteTime_Base,
-			prometheus.GaugeValue,
-			float64(volume.PercentDiskWriteTime_Base),
+			float64(volume.PercentDiskWriteTime)/float64(volume.PercentDiskWriteTime_Base)*100,
```
brian-brazil: By any chance are these the raw data?

collectors/perf.go
carlpett: That is not really specified in the documentation, sadly :( Would that make it useless to expose?

collectors/perf.go
carlpett: Yes, that is better, thanks

collectors/perf.go
brian-brazil: It makes it a lot less useful, but if it's all we have it's all we have.

brian-brazil: The docs we need are https://msdn.microsoft.com/en-us/library/ms803973.aspx and then to https://msdn.microsoft.com/en-us/library/ms804018.aspx

This is actually a Counter, as we're pulling from `PerfDataRaw`.

carlpett: I just found the same thing, but at a different part of msdn (it's a maze of a site...): https://msdn.microsoft.com/en-us/windows/hardware/aa389383(v=vs.71). I'm changing the metrics to counters, now. What would be good new names? `reads_total` etc?

collectors/perf.go
```diff
@@ -406,233 +180,72 @@ func (c *PerfCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc,
...
+			float64(volume.FreeMegabytes)*1024*1024,
 			volume.Name,
 		)
 
 		ch <- prometheus.MustNewConstMetric(
-			c.PercentFreeSpace,
+			c.ReadTime,
 			prometheus.GaugeValue,
-			float64(volume.PercentFreeSpace),
+			float64(volume.PercentDiskReadTime)/float64(volume.PercentDiskReadTime_Base),
```
brian-brazil: This is actually a Summary, but the denominator is a time advancing at normal.

I think we should expose just PercentDiskReadTime as a Counter. The units are 100ns, so that'll need converting.

collectors/perf.go
brian-brazil: Yes, `wmi_perf_logicaldisk_reads_total`.

Is the `perf` redundant?

collectors/perf.go
```diff
@@ -406,233 +180,72 @@ func (c *PerfCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc,
...
+			float64(volume.PercentDiskWriteTime)/float64(volume.PercentDiskWriteTime_Base),
 			volume.Name,
 		)
 
 		ch <- prometheus.MustNewConstMetric(
-			c.PercentIdleTime,
+			c.FreeSpace,
 			prometheus.GaugeValue,
-			float64(volume.PercentIdleTime),
+			float64(volume.PercentFreeSpace)/float64(volume.PercentFreeSpace_Base),
```
brian-brazil: We'll need to figure out what the units are experimentally as it's not documented, but this is two gauges.

brian-brazil:
Now that we've mostly figured out the type system, we should start a wiki page or something documenting the mapping from the various "Counter" types to Prometheus Metrics.

collectors/perf.go
carlpett: Yes, the perf part has been dropped  (not yet pushed, though). All the WMI classes we'll use are prefixed with perf.

brian-brazil: Can we add https://msdn.microsoft.com/en-us/library/ms803973.aspx to the PR too for future reference?

collectors/perf.go
carlpett: Hm, how do you mean? The values I'm getting from this division seem correct? Or did you mean to split it into total space and available space?

brian-brazil: It should be total and available space.

carlpett: Ok, I haven't been paying attention to the actual raw data... `PercentFreeSpace_Base` is just total disk size (in megabytes). Then splitting into total and available is very simple, at least :)

collectors/perf.go
```diff
@@ -319,78 +135,36 @@ func (c *PerfCollector) Collect(ch chan<- prometheus.Metric) {
...
 
 type Win32_PerfRawData_PerfDisk_LogicalDisk struct {
-	Name                         string
-	AvgDiskBytesPerRead          uint64
-	AvgDiskBytesPerRead_Base     uint32
-	AvgDiskBytesPerTransfer      uint64
-	AvgDiskBytesPerTransfer_Base uint32
-	AvgDiskBytesPerWrite         uint64
-	AvgDiskBytesPerWrite_Base    uint32
-	AvgDiskQueueLength           uint64
```
brian-brazil: It turns out that this is a Counter - which makes no sense for a queue length,

What it actually is is disk requests queued.

carlpett:
Hm, the rename wasn't very good, now all the comments are hidden away in other diffs :/

collectors/perf.go
carlpett: Aha, ok. Just convert to seconds and it's good, then?

brian-brazil:
I'm still perusing the docs, but I have a crazy idea.

It looks like it's possible to get to CounterType with WMI (though we'll need to use a different WMI library). Given that and presuming that metric naming is relatively consistent we may be able to fully automate the transformation of metric names and the user would only have to provide a WMI class name. We'll likely still need some hardcoded rules to fix e.g. bytes->megabytes.

I also note we cannot completely rely on documentation alone for CounterTypes - several of the ones in this PR changed in Windows 2000 (which means this won't quite work on XP).

collectors/perf.go
brian-brazil: Yes

collectors/perf.go
carlpett: Changed to `requests_queued_total`

carlpett:
@brian-brazil Interesting idea! We could just have a mapping list that depends on the OS version, for compatibility. Plus some utility functions for standard unit conversions in case they are not fully standardized.

Along with the discussion re IIS/SQL server metrics, if those should become separate exporters maybe such automatic transformation and conversion could form the base for a wmi exporter library that can be used to easily build new exporters?

carlpett:
@brian-brazil I _think_ I've addressed all the comments now, and the numbers look reasonable from what I can see

brian-brazil:
The counter types is what changed between versions.

What I'd see happening for IIS/SQL is that you'd run the WMI exporter but with different flags just pulling in IIS/SQL metrics.

collectors/logical_disk.go
```diff
@@ -0,0 +1,264 @@
...
+	for _, volume := range dst {
+		if volume.Name == "_Total" ||
+			c.volumeBlacklistPattern.MatchString(volume.Name) ||
+			!c.volumeWhitelistPattern.MatchString(volume.Name) {
+			continue
+		}
+
+		ch <- prometheus.MustNewConstMetric(
+			c.RequestsQueued,
+			prometheus.CounterValue,
```
brian-brazil: Gauge

collectors/logical_disk.go
```diff
@@ -0,0 +1,264 @@
...
+			c.WritesTotal,
+			prometheus.CounterValue,
+			float64(volume.DiskWritesPerSec),
+			volume.Name,
+		)
+
+		ch <- prometheus.MustNewConstMetric(
+			c.ReadTime,
+			prometheus.CounterValue,
+			float64(volume.PercentDiskReadTime)/1e7,
```
brian-brazil: We should make the 1e7 a constant for clarity.

collectors/logical_disk.go
```diff
@@ -0,0 +1,264 @@
...
+
+		TotalSpace: prometheus.NewDesc(
+			prometheus.BuildFQName(wmiNamespace, subsystem, "total_space_bytes"),
+			"Total space in bytes (LogicalDisk.PercentFreeSpace_Base)",
+			[]string{"volume"},
+			nil,
+		),
+
+		IdleTime: prometheus.NewDesc(
+			prometheus.BuildFQName(wmiNamespace, subsystem, "idle_time_total"),
```
brian-brazil: idle_time_seconds_total or idle_seconds_total

collectors/logical_disk.go
```diff
@@ -0,0 +1,264 @@
...
+
+		FreeSpace: prometheus.NewDesc(
+			prometheus.BuildFQName(wmiNamespace, subsystem, "free_space_bytes"),
+			"Free space in bytes (LogicalDisk.PercentFreeSpace)",
+			[]string{"volume"},
+			nil,
+		),
+
+		TotalSpace: prometheus.NewDesc(
+			prometheus.BuildFQName(wmiNamespace, subsystem, "total_space_bytes"),
```
brian-brazil: Usually this'd just be size_bytes (in unix anyway)

martinlindhe:
Hm, maybe need to rebase this branch. Appveyor is complaining it is unmergable without conflicts now.

carlpett:
Yep, just pushed :) (Very nice work adding it, btw!)

carlpett:
Need to squash before merging as well

martinlindhe:
@brian-brazil interesting idea, indeed. there is hundreds of WMI sources, but I'm not sure most of them is very usable. Let's discuss that in a separate issue #12

brian-brazil:
:+1: 

I'd suggest squashing commits.

carlpett:
Done!

martinlindhe:
Great work!

martinlindhe:
> collectors/logical_disk.go | 265 ++++++++++++++++++++
> collectors/perf.go         | 593 ---------------------------------------------

i love this diff :) thanks for all the work

PR #20, opened 2016-09-02
os: expose Win32_ComputerSystem and Win32_PerfRawData_PerfOS_System

martinlindhe:
Was looking over the bosun WMI collectors a bit and thought this might be useful.

### DIFF
--- a/collector/os.go
+++ b/collector/os.go
@@ -14,93 +14,82 @@ func init() {
 	Factories["os"] = NewOSCollector
 }
 
-// A OSCollector is a Prometheus collector for WMI Win32_OperatingSystem metrics
+// A OSCollector is a Prometheus collector for WMI metrics
 type OSCollector struct {
 	PhysicalMemoryFreeBytes *prometheus.Desc
 	PagingFreeBytes         *prometheus.Desc
 	VirtualMemoryFreeBytes  *prometheus.Desc
-	ProcessesMax            *prometheus.Desc
-	ProcessMemoryMaxBytes   *prometheus.Desc
+	ProcessesLimit          *prometheus.Desc
+	ProcessMemoryLimitBytes *prometheus.Desc
 	Processes               *prometheus.Desc
 	Users                   *prometheus.Desc
-	PagingMaxBytes          *prometheus.Desc
+	PagingLimitBytes        *prometheus.Desc
 	VirtualMemoryBytes      *prometheus.Desc
 	VisibleMemoryBytes      *prometheus.Desc
 }
 
 // NewOSCollector ...
 func NewOSCollector() (Collector, error) {
-
 	const subsystem = "os"
 
 	return &OSCollector{
-
-		PagingMaxBytes: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "paging_max_bytes"),
-			"SizeStoredInPagingFiles",
+		PagingLimitBytes: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "paging_limit_bytes"),
+			"OperatingSystem.SizeStoredInPagingFiles",
 			nil,
 			nil,
 		),
-
 		PagingFreeBytes: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "paging_free_bytes"),
-			"FreeSpaceInPagingFiles",
+			"OperatingSystem.FreeSpaceInPagingFiles",
 			nil,
 			nil,
 		),
-
 		PhysicalMemoryFreeBytes: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "physical_memory_free_bytes"),
-			"FreePhysicalMemory",
+			"OperatingSystem.FreePhysicalMemory",
 			nil,
 			nil,
 		),
-
 		Processes: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "processes"),
-			"NumberOfProcesses",
+			"OperatingSystem.NumberOfProcesses",
 			nil,
 			nil,
 		),
-
-		ProcessesMax: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "processes_max"),
-			"MaxNumberOfProcesses",
+		ProcessesLimit: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "processes_limit"),
+			"OperatingSystem.MaxNumberOfProcesses",
 			nil,
 			nil,
 		),
-
-		ProcessMemoryMaxBytes: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "process_memory_max_bytes"),
-			"MaxProcessMemorySize",
+		ProcessMemoryLimitBytes: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "process_memory_limix_bytes"),
+			"OperatingSystem.MaxProcessMemorySize",
 			nil,
 			nil,
 		),
-
 		Users: prometheus.NewDesc(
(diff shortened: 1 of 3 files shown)


martinlindhe:
Should probably also include some data from

```go
type Win32_PerfRawData_PerfOS_System struct {
	ContextSwitchesPersec     uint32
	ExceptionDispatchesPersec uint32
	Frequency_Object          uint64
	Processes                 uint32
	ProcessorQueueLength      uint32
	SystemCallsPersec         uint32
	SystemUpTime              uint64
	Threads                   uint32
	Timestamp_Object          uint64
}
```

carlpett:
I quite like the design where there is one collector per WMI class (and thus query), is there some strong incentive for coupling these two together? I think this especially goes for the `os` collector, which has a broad enough definition that it might otherwise over time sprout all sorts of features from different classes.
I'm not fully decided on what I think about multiple collectors sharing the same subsystem name (that is, let these new metrics keep the `os` subsystem tag, but move them into another collector). It might be a bit confusing, I think.

So, I think I'd propose creating a new collector for `Win32_ComputerSystem`.

carlpett:
Another thing comes to mind - is the same data available from some other class of the `PerfRawRata` type?

martinlindhe:
Sure, we can keep them separate. I'll update the patch, probably tomorrow.
Only motivation for keeping them in os.go was they are in the same logical group of stuff, some metrics even overlap. If they are in different files, we'd better come up with different subsystem names so they can be disabled/enabled on the cli.

martinlindhe:
Same data is possibly available in other places too, there are so many of the WMI classes, and many dont seem to be documented at all except on non-microsoft sites.

carlpett:
I added a little Powershell snippet that I came up with to list available WMI classes on the Wiki page: https://github.com/martinlindhe/wmi_exporter/wiki/Prometheus-and-WMI-CounterTypes#finding-available-counters. That at least give a listing, and it's then a manual job to go through them to find something useful, but most of them have at least somewhat transparent names.

Strangely, I couldn't find a `PerfRaw` class which gives total memory. `Win32_PerfRawData_PerfOS_Memory` sounded good, but as far as I can tell, it does not have this.

martinlindhe:
I'm not too surprised of the inconsistency in the WMI metrics. Seems to be produced over many years by different teams at Microsoft.

martinlindhe:
@carlpett I've updated the patch by separating the wmi collectors. the new ones are disabled by default at the moment. maybe they should be enabled by default, as they are likely available on every windows machine.

collector/os.go
```diff
@@ -30,77 +30,66 @@ type OSCollector struct {
...
-			"NumberOfProcesses",
+			"OperatingSystem.NumberOfProcesses",
 			nil,
 			nil,
 		),
-
 		ProcessesMax: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "processes_max"),
-			"MaxNumberOfProcesses",
+			"OperatingSystem.MaxNumberOfProcesses",
```
carlpett: Shouldn't we ditch this one? We've got `NumberOfProcesses` already, so we can use PromQL to get the max

collector/os.go
```diff
@@ -30,77 +30,66 @@ type OSCollector struct {
...
 			nil,
 			nil,
 		),
-
 		ProcessMemoryMaxBytes: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "process_memory_max_bytes"),
-			"MaxProcessMemorySize",
+			"OperatingSystem.MaxProcessMemorySize",
```
carlpett: Ditto this one, if we have the non-max metric (and if not, is this useful?)

collector/perfos.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	Factories["perfos"] = NewPerfOSCollector
+}
+
+// A PerfOSCollector is a Prometheus collector for WMI metrics
+type PerfOSCollector struct {
+	// XXX names
```
carlpett: Meant to be removed before merging? :)

collector/perfos.go
```diff
@@ -0,0 +1,133 @@
...
+	ContextSwitchesPersec     *prometheus.Desc
```
carlpett: Are these "per second" metrics actually per second? The ones in `logical_disk` were actually counters

carlpett:
Had a couple of comments. But yes, they should be on by default, I agree.

collector/os.go
brian-brazil: You want this one to catch peaks that happen between scrapes.

collector/os.go
martinlindhe: Hm, this is the physical upper limit of processes, not the max number seen (as you could get with promql), or am i misunderstanding?

collector/os.go
martinlindhe: Leaving it for now then

collector/perfos.go
martinlindhe: Yup, sorry!

collector/perfos.go
martinlindhe: Not sure at all (and there's no official documentation of these fields)
Here's an example output:

```
# TYPE wmi_perfos_context_switches_per_sec gauge
wmi_perfos_context_switches_per_sec 4.272205793e+09
# TYPE wmi_perfos_exception_dispatches_per_sec gauge
wmi_perfos_exception_dispatches_per_sec 1.2127682e+07
# TYPE wmi_perfos_system_calls_per_sec gauge
wmi_perfos_system_calls_per_sec 4.26178136e+08
```

martinlindhe:
removed a "XXX names" and squashed commit

collector/os.go
brian-brazil: processes_limit would be a better name then I think.

martinlindhe: renamed max to limit in the metrics

martinlindhe:
max -> limit in metric names

collector/perfos.go
```diff
@@ -0,0 +1,132 @@
...
+func (c *PerfOSCollector) Collect(ch chan<- prometheus.Metric) error {
+	if desc, err := c.collect(ch); err != nil {
+		log.Println("[ERROR] failed collecting os metrics:", desc, err)
+		return err
+	}
+	return nil
+}
+
+type Win32_PerfRawData_PerfOS_System struct {
+	ContextSwitchesPersec     uint32
```
carlpett: I think this is actually a total. This is the current value on my machine:
`ContextSwitchesPersec          : 666198955`

carlpett: This goes for the other Persec as well

collector/perfos.go
```diff
@@ -0,0 +1,132 @@
...
+	ExceptionDispatchesPersec *prometheus.Desc
+	ProcessorQueueLength      *prometheus.Desc
+	SystemCallsPersec         *prometheus.Desc
+	SystemUpTime              *prometheus.Desc
+	Threads                   *prometheus.Desc
+}
+
+// NewPerfOSCollector ...
+func NewPerfOSCollector() (Collector, error) {
+	const subsystem = "perfos"
```
carlpett: Sorry about my constant naming nitpicking... But shouldn't this be `system` or possibly `os_system` (which is a bit wierd, but still)?
The "base name" of the raw metrics is `Win32_PerfRawData_Perf`, so the specific class here is really `OS_System`

collector/perfos.go
martinlindhe: Thanks, updating these

collector/perfos.go
martinlindhe: No problem. Better get the naming right from the start, than having to change it later :)
I've renamed the file to system.go, the subsystem to "system". Patch updated.

carlpett:
Looks good to me! Great work! :cake:

PR #23, opened 2016-09-16
Remove branch requirement for deployments, since it seems broken in combination with tags

carlpett:
This fixes the fact that AppVeyor currently does not upload releases even though they are tagged. Apparently the branch name is replaced by the tag name (in contraditiction with their examples...)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -43,5 +43,4 @@ deploy:
     draft: false
     prerelease: false
     on:
-      branch: master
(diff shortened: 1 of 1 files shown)


carlpett:
Also proposing that we remove the two deployments I created while testing :)

martinlindhe:
:+1: 
Sounds good. Will you remove the releases?

carlpett:
Done! First real release released :smile:

PR #26, opened 2016-09-17
Cpu collector

carlpett:
Some initial work on the cpu collector. I chose the [`Win32_PerfRawData_PerfOS_Processor`](https://msdn.microsoft.com/en-us/library/aa394317(v=vs.90).aspx) over `Win32_PerfRawData_Counters_ProcessorInformation` since it is more widely available (the latter was introduced in Windows 2008R2). 

A few notes: 
- There is a "total" cpu time exposed (should probably be removed before merging anyway, but I added it for reference initially), but I can't get it to match the other timings:
```bash
wmi_cpu_time_total{core="0",mode="idle"} 27154.335649999997
wmi_cpu_time_total{core="0",mode="interrupt"} 409.140625
wmi_cpu_time_total{core="0",mode="privileged"} 4150.9375
wmi_cpu_time_total{core="0",mode="total"} 28108.4375
wmi_cpu_time_total{core="0",mode="user"} 3325.8125
```
My total uptime was ~76000 sec when this was run. From the documentation, I would have expected total to be the sum of interrupt, privileged and user, and that total+idle would be equal to the uptime.
- Are C-states interesting? Should we in that case include transition count as well?
- I kept `dpc_time_total` as a separate metric, since it is supposedly included in the `privileged` time (From the perfmon help text: "DPC Time is a component of Privileged Time"). Good or bad?

### DIFF
--- a/collector/cpu.go
+++ b/collector/cpu.go
@@ -1,0 +1,211 @@
+// returns data points from Win32_PerfRawData_PerfOS_Processor
+// https://msdn.microsoft.com/en-us/library/aa394317(v=vs.90).aspx - Win32_PerfRawData_PerfOS_Processor class
+package collector
+
+import (
+	"log"
+	"strings"
+
+	"github.com/StackExchange/wmi"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	Factories["cpu"] = NewCPUCollector
+}
+
+// A CPUCollector is a Prometheus collector for WMI Win32_PerfRawData_PerfOS_Processor metrics
+type CPUCollector struct {
+	CStateSecondsTotal *prometheus.Desc
+	TimeTotal          *prometheus.Desc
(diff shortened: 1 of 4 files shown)


carlpett:
@brian-brazil Any suggestions around this? There is a lot of possible data to expose, but eg power saving modes does not strike me as something I would use, at least. Probably not interrupts either, but that is something that I can at least imagine that some people could have some use for. 
My feeling is that we should probably just go for some minimal initial metric set (only `wmi_cpu_time_total`?) and expand as we go. 

But it would be good to have another set of eyes on the sample data above - does it seem reasonable?

brian-brazil:
I'd be inclined to include everything. Interrupts is an important one for example for network performance debugging.

carlpett:
After these fixes, I have no further planned work - any remaining comments?

martinlindhe:
LGTM!

carlpett:
Rebased on master and squashed

PR #28, opened 2016-09-24
Add a collector generator script

### LINKED ISSUE #12: Automatically convert WMI metrics to prometheus
@brian-brazil wrote in #5 
> I'm still perusing the docs, but I have a crazy idea.

> It looks like it's possible to get to CounterType with WMI (though we'll need to use a different WMI library). Given that and presuming that metric naming is relatively consistent we may be able to fully automate the transformation of metric names and the user would only have to provide a WMI class name. We'll likely still need some hardcoded rules to fix e.g. bytes->megabytes.

> I also note we cannot completely rely on documentation alone for CounterTypes - several of the ones in this PR changed in Windows 2000 (which means this won't quite work on XP).

> The counter types is what changed between versions.

> What I'd see happening for IIS/SQL is that you'd run the WMI exporter but with different flags just pulling in IIS/SQL metrics.

carlpett:
There is a fair amount of boilerplate involved in creating a new collector, so I've attempted to automate that part. This PR adds a script which queries a WMI class to pull its member list, and then creates a skeleton collector implementation.

One potential improvement would be if we could get the CounterType from the WMI query as well, then it would be possible to make a better guess on what type of metric should be exposed. Right now it is hardcoded to `GaugeValue`.

It doesn't really fix #12, but it at least removes some of the tedium...


martinlindhe:
very cool! I been meaning to learn "go generate", looks like i should study your impl. :)

tools/collector-generator/New-Collector.ps1
```diff
@@ -0,0 +1,18 @@
...
+    | Where-Object { $_.Definition -Match '^u?int' -and $_.Name -NotMatch '_' } `
+    | Select-Object Name, @{Name="Type";Expression={$_.Definition.Split(" ")[0]}}
+$input = @{
+    "Class"=$Class;
+    "CollectorName"=$CollectorName;
+    "Members"=$members
+} | ConvertTo-Json
+$outFileName = "..\..\collector\$CollectorName.go".ToLower()
+$input | .\collector-generator.exe | Out-File -NoClobber -Encoding UTF8 $outFileName
+go fmt $outFileName
```
martinlindhe: nitpick: text files should end with a new line

tools/collector-generator/collector.template
```diff
@@ -0,0 +1,62 @@
...
+    }
+    {{ range $m := .Members }}
+    ch <- prometheus.MustNewConstMetric(
+        c.{{ $m.Name }},
+        prometheus.GaugeValue,
+        float64(dst[0].{{ $m.Name }}),
+    )
+    {{ end }}
+    return nil, nil
+}
```
martinlindhe: nitpick: text files should end with a new line

carlpett:
Actually, this doesn't use `go generate`, should probably look into that... This just uses the `text/template` package

martinlindhe:
You're right of course. I commented before properly reading the code.

carlpett:
I had heard about it but it didn't cross my mind when doing this. I'll check it out some day, it is probably a better solution

carlpett:
Fixed and squashed!

martinlindhe:
great work!

Issue #21, opened 2016-09-02
Feature: network metrics


martinlindhe:
Expose as "net" and up naming with regards to prometheus conventions, could be based on https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/network_windows.go

* all network interfaces should be exposed
* should have white/blacklist like iis.go
* enabled by default?

martinlindhe:
I intend to work on this during the current week

martinlindhe:
Note: MSFT_NetAdapterStatisticsSettingData class seems useful, but is only available on Windows Server 2012+
https://msdn.microsoft.com/en-us/library/hh872390

PR #29, opened 2016-09-27
 add network collector

martinlindhe:
Implements #21

### DIFF
--- a/collector/net.go
+++ b/collector/net.go
@@ -1,0 +1,244 @@
+// returns data points from Win32_PerfRawData_Tcpip_NetworkInterface
+
+// https://technet.microsoft.com/en-us/security/aa394340(v=vs.80) (Win32_PerfRawData_Tcpip_NetworkInterface class)
+// https://msdn.microsoft.com/en-us/library/aa394216 (Win32_NetworkAdapter class)
+// https://msdn.microsoft.com/en-us/library/aa394353 (Win32_PnPEntity class)
+
+package collector
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"regexp"
+
+	"github.com/StackExchange/wmi"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	Factories["net"] = NewNetworkCollector
+}
+
+var (
+	nicWhitelist        = flag.String("collector.net.nic-whitelist", ".+", "Regexp of NIC:s to whitelist. NIC name must both match whitelist and not match blacklist to be included.")
+	nicBlacklist        = flag.String("collector.net.nic-blacklist", "", "Regexp of NIC:s to blacklist. NIC name must both match whitelist and not match blacklist to be included.")
+	nicNameToUnderscore = regexp.MustCompile("[^a-zA-Z0-9]")
+)
+
+// A NetworkCollector is a Prometheus collector for WMI Win32_PerfRawData_Tcpip_NetworkInterface metrics
+type NetworkCollector struct {
+	BytesReceivedTotal       *prometheus.Desc
+	BytesSentTotal           *prometheus.Desc
+	BytesTotal               *prometheus.Desc
+	PacketsOutboundDiscarded *prometheus.Desc
+	PacketsOutboundErrors    *prometheus.Desc
+	PacketsTotal             *prometheus.Desc
+	PacketsReceivedDiscarded *prometheus.Desc
+	PacketsReceivedErrors    *prometheus.Desc
+	PacketsReceivedTotal     *prometheus.Desc
+	PacketsReceivedUnknown   *prometheus.Desc
+	PacketsSentTotal         *prometheus.Desc
+
+	nicWhitelistPattern *regexp.Regexp
+	nicBlacklistPattern *regexp.Regexp
+}
+
+// NewNetworkCollector ...
+func NewNetworkCollector() (Collector, error) {
+	const subsystem = "net"
+
+	return &NetworkCollector{
+		BytesReceivedTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "bytes_received_total"),
+			"(Network.BytesReceivedPerSec)",
+			[]string{"nic"},
+			nil,
+		),
+		BytesSentTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "bytes_sent_total"),
+			"(Network.BytesSentPerSec)",
+			[]string{"nic"},
+			nil,
+		),
+		BytesTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "bytes_total"),
+			"(Network.BytesTotalPerSec)",
+			[]string{"nic"},
+			nil,
+		),
+		PacketsOutboundDiscarded: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "packets_outbound_discarded"),
+			"(Network.PacketsOutboundDiscarded)",
+			[]string{"nic"},
+			nil,
+		),
+		PacketsOutboundErrors: prometheus.NewDesc(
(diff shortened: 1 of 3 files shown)


carlpett:
Nice, I'll have a look! Did you use the generator for this?

collector/net.go
```diff
@@ -0,0 +1,178 @@
...
+	nicWhitelist = flag.String("collector.network.nic-whitelist", ".+", "Regexp of NIC:s to whitelist. NIC name must both match whitelist and not match blacklist to be included.")
+	nicBlacklist = flag.String("collector.network.nic-blacklist", "", "Regexp of NIC:s to blacklist. NIC name must both match whitelist and not match blacklist to be included.")
+)
+
+// A NetworkCollector is a Prometheus collector for WMI Win32_PerfRawData_Tcpip_NetworkInterface metrics
+type NetworkCollector struct {
+	// XXX names
+	BytesReceivedPerSec *prometheus.Desc
+	BytesSentPerSec     *prometheus.Desc
+	BytesTotalPerSec    *prometheus.Desc
```
carlpett: These are counters, as far as I can tell, so drop the `PerSec`

collector/net.go
```diff
@@ -0,0 +1,178 @@
...
+
+	CurrentBandwidth *prometheus.Desc
+
+	PacketsPerSec         *prometheus.Desc
+	PacketsReceivedPerSec *prometheus.Desc
+	PacketsSentPerSec     *prometheus.Desc
```
carlpett: Same here

collector/net.go
martinlindhe: Patch updated, renamed "PerSec" to Total

collector/net.go
```diff
@@ -0,0 +1,178 @@
...
+type Win32_PerfRawData_Tcpip_NetworkInterface struct {
+	CurrentBandwidth      uint32
+	BytesReceivedPerSec   uint32
+	BytesSentPerSec       uint32
+	BytesTotalPerSec      uint64
+	Name                  string
+	PacketsPerSec         uint32
+	PacketsReceivedPerSec uint32
+	PacketsSentPerSec     uint32
+}
```
carlpett: Some suggested additions, since they seem to provide a bit of error detection:
```
PacketsOutboundDiscarded
PacketsOutboundErrors
PacketsReceivedDiscarded
PacketsReceivedErrors
PacketsReceivedUnknown (?)
```

martinlindhe: Okay, let's expose these too

martinlindhe:
> Did you use the generator for this?

No, that slipped my mind entirely...

collector/net.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	Factories["net"] = NewNetworkCollector
+}
+
+var (
+	nicWhitelist = flag.String("collector.network.nic-whitelist", ".+", "Regexp of NIC:s to whitelist. NIC name must both match whitelist and not match blacklist to be included.")
+	nicBlacklist = flag.String("collector.network.nic-blacklist", "", "Regexp of NIC:s to blacklist. NIC name must both match whitelist and not match blacklist to be included.")
```
carlpett: Should we default to blacklisting the `isatap.*` as well as `Teredo` and other 6to4-tunneling interfaces?

carlpett:
> should network be enabled by default?
Yes, I think so

collector/net.go
martinlindhe: On my test box i have 3 "isatap.*" interfaces, but no Teredo ones (win7)

collector/net.go
```diff
@@ -0,0 +1,161 @@
...
+
+// NewNetworkCollector ...
+func NewNetworkCollector() (Collector, error) {
+	const subsystem = "net"
+
+	return &NetworkCollector{
+		BytesReceivedTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "bytes_received_total"),
+			"(Network.BytesReceivedPerSec)",
+			[]string{"nic"},
```
carlpett: On my local machine, I get this `Intel[R] Ethernet Connection I218-LM` - could this kind of labels make querying hard? Not sure what to do about it if so, but we might want to think about it.

martinlindhe: There is ways of mapping it to something else, they are doing it in this code: https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/network_windows.go

Didn't dig deeper into that, yet.

martinlindhe:
Patch updated:

* enabled "net" by default
* exclude "isatap.+" by default
* added more metrics

collector/net.go
brian-brazil: I'd default to not filtering anything

martinlindhe:
Applied changes to expose interface name differently, mapping Win32_NetworkAdapter.PNPDeviceID into Win32_PnPEntity.DeviceID and using the resulting Name. I have not tested, but suspect this is done to provide a stable device id (consistent over reboots)

On my test box, it looks like:

```
Interface11 = Intel[R] Ethernet Connection I219-LM          
Interface12 = Intel[R] Dual Band Wireless-AC 8260           
Interface23 = VMware Virtual Ethernet Adapter for VMnet1    
Interface24 = VMware Virtual Ethernet Adapter for VMnet8
```

collector/net.go
martinlindhe: @brian-brazil I missed your comment. Patch updated to remove default blacklist "isatap.+"

martinlindhe:
Ironically, now with the adapter name to interface mapping, the isatap stuff gets excluded anyway. Seems there is no corresponding matches for them and they are skipped  by

```go

		iface := mNicInstanceNameToInterfaceIndex[nic.Name]
		if iface == "" {
			continue
		}
```

carlpett:
Hm, from my experiments here, the `Win32_PnPEntity` has the same name as given in `Win32_PerfRawData_Tcpip_NetworkInterface`?

But are we sure this is a problem? @brian-brazil - are there any caveats/usability issues with having labels with brackets etc?

If we do make them different, we should ensure that it is at least something that is easy to figure out how it corresponds to the interfaces seen on the machine. We could use one of `InterfaceIndex`, `DeviceID` or `Index`  from the intermediary `Win32_NetworkAdapter` that you already query. I'm not really sure of the difference between them - `DeviceID` and `Index` seem to always be the same, while `InterfaceIndex` looks more arbitrary. I'd still vouch for `InterfaceIndex`, since that is the parameter you'd send to `Get-NetAdapter` in Powershell. Then also probably change the label name to `interface_index`?

carlpett:
Hm, I misread the code, that's why I couldn't get it to line up with your results.

martinlindhe:
If all we want is a bracket-less identifier, we could just mangle the interface name, say "Intel[R] Ethernet Connection I218-LM" into "Intel_R__Ethernet_Connection_I218_LM" (non-alphanumeric letters turns into _).

martinlindhe:
Patch updated:
* Renamed label to "interface_index"

carlpett:
Yup, that would work too. Would reduce the scope of the collector a bit as well, which is always nice.
Something that just popped into my mind, but I think might not be a problem after all, is if the interface name is guaranteed to be unique. I'm _pretty_ sure Windows makes it so, and all the machines I have access to supports this. I'll google a bit.

carlpett:
I tried setting a duplicate name on an interface and got an error. Phew.

martinlindhe:
I think too Windows makes them unique (by appending # 2, # 3 etc on the end, without spaces)

martinlindhe:
Okay, reworking the patch to use mangled interface names, and calling the label "nic" again, right?

martinlindhe:
Patch updated:
* drop the interface_index stuff
* mangle interface names (with a test)

carlpett:
Looks good to me! And the first test, as well! :tada:

PR #30, opened 2016-09-27
exporter: add /health endpoint. fixes #25

### LINKED ISSUE #25: / should serve some welcome page instead of redirecting
Currently we do a redirect from `/` to `/metrics`. Drawback to this is that there is no request that can be done to check that everything is alright (http server is working, at least) without also incurring the cost of running the wmi queries. 

Our use case is that we have `/` as a http healthcheck endpoint for exporters which we register in Consul. If we want to keep the redirect, we should have _some_ endpoint which indicates that things are ok, at least.

martinlindhe:
@carlpett Will this work, with respect to #25 ?

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -3,6 +3,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"sort"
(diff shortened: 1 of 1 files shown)


carlpett:
Yup!

martinlindhe:
Great, merging

PR #31, opened 2016-09-27
simplify appveyor

martinlindhe:
It seems appveyor has been upgraded to have go 1.7.1 out-of-box.

so this patch simplifies appveyor.yml to remove the custom "install go 1.7" build step. should speed up CI by a few seconds


carlpett:
Nice!

PR #32, opened 2016-09-29
add boilerplate for promu

martinlindhe:
For this to be usable, https://github.com/prometheus/promu/pull/47 needs to land first

This excludes setting BuildUser. I dont see it is useful. Thought?


carlpett:
Hm. Kluven.
Here are some potential benefits, though their actual value can be contested:
1. It would help point out that a binary was built somewhere else than AppVeyor, ie may not be consistent with any particular state in the repo
2. The version output will still contain a user field, so not populating it will "look odd"
3. "Might as well for consistency"

Yup, not very good ones, but it's what I've got.

martinlindhe:
Okay all good points!
I re-added it and squashed commit.
Personally i feel it's too much info (user@host), but no big deal since appveyor will build binaries for us anyway.

carlpett:
Now that the promu change is merged, this just needs a bump to govendor and then we're good to release 0.1, right?

martinlindhe:
yep, think so! should we try to finish up 0.1 today (merge this pr and bump vendor), and maybe post a little announcement tomorrow or so? any objections?

carlpett:
@martinlindhe Sorry, been away a while. Yes, I think that sounds excellent!

martinlindhe:
@carlpett no worries, welcome back!

PR #34, opened 2016-10-13
bumped govendor

martinlindhe:
updated dependencies.

please note: we recently improved govendor compat for windows, but additional windows fixes has landed so be sure to update govendor if reproducing this


carlpett:
Not very easy to review, anything in particular that needs looking at?

martinlindhe:
dont think so, it should just be a straightforward dependency bump, tested without and with this change, no differences

carlpett:
Ok!

PR #35, opened 2016-10-13
appveyor: build using promu to embed branch and revision info into binary

### LINKED ISSUE #24: Version info is not compiled into binary
Noticed that version info is not compiled into the binary, so log lines don't show much. I'm guessing we're not doing all the things that is expected from prometheus/common? Since most prometheus tools use promu, maybe there is some magic there? For instance, the promu.yml file adds some ldflags: https://github.com/prometheus/promu/blob/master/.promu.yml

martinlindhe:
resolves #24


martinlindhe:
ping @carlpett

carlpett:
Sorry, been away a while. Looks good!

PR #39, opened 2016-10-18
Setup GitVersion

carlpett:
So, this works fine on my machine, but AppVeyor logs this:
`command failed: fork/exec c:\go\bin\go.exe: invalid argument`

Probably related to it also seeming to mangle the parameters sent to promu. Haven't figured out why yet. 
@martinlindhe Is there some way to share access to the AppVeyor account? I can't do anything much more sophisticated than adding echo-commands to the build script and rerunning.


carlpett:
Also somewhat alarming is that AppVeyor _does not fail the build_ when promu fails. Should also be investigated.

carlpett:
You can look at the build triggered to see the look of the version, btw. This build was `0.1.1-PullRequest.39+4`

martinlindhe:
looking into if i can add you to appveyour...

martinlindhe:
> Also somewhat alarming is that AppVeyor does not fail the build when promu fails. Should also be investigated.

this is because promu is not signaling failure here: https://github.com/prometheus/promu/blob/master/cmd/build.go#L112
I sent a patch about this just now: https://github.com/prometheus/promu/pull/50, which also outputs the "go" parameters that was failing.

it still doesnt resolve the root cause, but will give us chance to dig further.

carlpett:
Nice find! I'll wait for it to get merged, then

martinlindhe:
Regarding users on appveyor. They are added at https://ci.appveyor.com/team, and need your email. 

I added the gmail you used to commit to this repository with as an administrator, hope that is the same you used to sign up there with.

/m

carlpett:
@martinlindhe Ah, excellent, I just got the email as well. Thanks, that'll hopefully make it possible to investigate from both ends then

martinlindhe:
That's great!

carlpett:
Fixed (turns out that the file was written with some weird encoding when using `Tee-Object`)

martinlindhe:
nice work! maybe the promu project is interested in your solution as it can remove the need to manually update a VERSION file

carlpett:
Will probably need a rewrite for cross-platform compatibility then - GitVersion is a C#-project delivered as a Windows executable which doesn't seem to even run on Mono... I'll give their algorithm a look though, might be simple enough to reimplement

martinlindhe:
Ah, didn't think of that! Nevermind then :)

martinlindhe:
PS, https://github.com/prometheus/promu/pull/50 is now merged

PR #41, opened 2016-11-14
system: fix uptime calculation

### LINKED ISSUE #40: Make wmi_system_system_up_time format more Prometheus friendly
The metric _wmi_system_system_up_time_ is exported as Pico-seconds since system-start.
To visualize it you have to use something like

`wmi_system_system_up_time{instance="$instance"}/1000000000000`

It would be more convenient to export the system start as Unix timestamp. See [Prometheus-Timestamp-Doc](https://prometheus.io/docs/practices/instrumentation/#timestamps-not-time-since).
This should result in a query like: 
`time() - wmi_system_system_up_time{instance="$instance"}`
Prometheus would also be able to better compress the time series.

martinlindhe:
…raw data doesnt change. fixes #40

### DIFF
--- a/collector/system.go
+++ b/collector/system.go
@@ -55,7 +55,7 @@ func NewSystemCollector() (Collector, error) {
 		),
 		SystemUpTime: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "system_up_time"),
-			"PerfOS_System.SystemUpTime",
+			"SystemUpTime/Frequency_Object",
 			nil,
 			nil,
 		),
@@ -81,10 +81,12 @@ func (c *SystemCollector) Collect(ch chan<- prometheus.Metric) error {
 type Win32_PerfRawData_PerfOS_System struct {
 	ContextSwitchesPersec     uint32
 	ExceptionDispatchesPersec uint32
+	Frequency_Object          uint64
 	ProcessorQueueLength      uint32
 	SystemCallsPersec         uint32
 	SystemUpTime              uint64
 	Threads                   uint32
+	Timestamp_Object          uint64
 }
 
 func (c *SystemCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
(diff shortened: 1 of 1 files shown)


martinlindhe:
Would like input from @carlpett  or @brian-brazil before merging this.

Also, this change makes existing exporters confused I guess since the value is radically changed, what would be best way to handle this. I would prefer semver (eg release 0.2.0 with this change?)

martinlindhe:
Second, would it make more sense to convert this to a unix timestamp? [1]

1: https://prometheus.io/docs/practices/instrumentation/#timestamps-not-time-since

carlpett:
Had a look at this - the way they calculate it in the formatted data is by using `TimestampObject`: `(TimestampObject-SystemUpTime)/TicksToSeconds` comes out correct. So I'd suggest we do that instead, so we don't have to query a new object.

Regarding the continuity of the metric, I'm not sure. Since the previous value was broken, I'd consider it a bugfix rather than a "feature", so I don't think it needs a point release from a semver point of view.

brian-brazil:
Generally we should always export raw data. For example there's probably a race in here in somewhere that can cause the value to fluctuate slightly.

martinlindhe:
Reworked the patch:
* dropped usage of preformatted data
* calc uptime as carl suggested: (Timestamp_Object-SystemUpTime)/Frequency_Object

collector/system.go
```diff
@@ -81,46 +81,49 @@ func (c *SystemCollector) Collect(ch chan<- prometheus.Metric) error {
...
 	SystemUpTime              uint64
 	Threads                   uint32
+	Timestamp_Object          uint64
 }
 
 func (c *SystemCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
-	var dst []Win32_PerfRawData_PerfOS_System
-	if err := wmi.Query(wmi.CreateQuery(&dst, ""), &dst); err != nil {
+	var raw []Win32_PerfRawData_PerfOS_System
+	if err := wmi.Query(wmi.CreateQuery(&raw, ""), &raw); err != nil {
```
carlpett: Minor nitpick: Now that you aren't pulling in multiple counters, might as well revert this name to `dst`, since that is what we use everywhere else (in truth, I'm not sure why we use that name, though? 😄  )

martinlindhe:
Reworked. Now it converts the Windows timestamp into a unix timestamp.

brian-brazil:
:+1:

PR #42, opened 2016-11-14
Add DNS server collector

carlpett:
This PR adds a collector for the Windows DNS Server.

There's not much documentation for this one, sadly. I've tested it on my work environment, and also set up a pair of VMs to test the zone transfer stuff (we don't use that) and it all _seems_ reasonable, but there may be some metrics that are actually overlapping (for instance `IXFRSuccessReceived` is _likely_ the sum of `IXFRTCPSuccessReceived` and `IXFRUDPSuccessReceived`, so I skipped it, but I was unable to make my systems actually generate non-zero values for those counters).

Naming is up for debate, I'm not super-pleased with a lot of the names. I've created a gist [here](https://gist.github.com/carlpett/f22ed6e8ea9dd7aa4f7291eaaa45b3ad) where the metrics, their labels and the included counters as well as some values from my production environment are shown.

### DIFF
--- a/collector/dns.go
+++ b/collector/dns.go
@@ -1,0 +1,492 @@
+// returns data points from Win32_PerfRawData_DNS_DNS
+// https://msdn.microsoft.com/en-us/library/ms803992.aspx?f=255&MSPPError=-2147217396
+// https://technet.microsoft.com/en-us/library/cc977686.aspx
+package collector
+
+import (
+	"log"
+
+	"github.com/StackExchange/wmi"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	Factories["dns"] = NewDNSCollector
+}
+
+// A DNSCollector is a Prometheus collector for WMI Win32_PerfRawData_DNS_DNS metrics
+type DNSCollector struct {
+	ZoneTransferRequestsReceived  *prometheus.Desc
+	ZoneTransferRequestsSent      *prometheus.Desc
+	ZoneTransferResponsesReceived *prometheus.Desc
+	ZoneTransferSuccessReceived   *prometheus.Desc
+	ZoneTransferSuccessSent       *prometheus.Desc
+	ZoneTransferFailures          *prometheus.Desc
+	MemoryUsedBytes               *prometheus.Desc
+	DynamicUpdatesQueued          *prometheus.Desc
+	DynamicUpdatesReceived        *prometheus.Desc
+	DynamicUpdatesFailures        *prometheus.Desc
+	NotifyReceived                *prometheus.Desc
+	NotifySent                    *prometheus.Desc
+	SecureUpdateFailures          *prometheus.Desc
+	SecureUpdateReceived          *prometheus.Desc
+	Queries                       *prometheus.Desc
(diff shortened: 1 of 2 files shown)


collector/dns.go
```diff
@@ -0,0 +1,492 @@
...
+}
+
+// NewDNSCollector ...
+func NewDNSCollector() (Collector, error) {
+	const subsystem = "dns"
+	return &DNSCollector{
+		ZoneTransferRequestsReceived: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "zone_transfer_requests_received_total"),
+			"Number of zone transfer requests (AXFR/IXFR) received by the master DNS server",
+			[]string{"type"},
```
brian-brazil: "type" is very generic. I'd suggest "rr" here.

brian-brazil: You'll have to forgive me, it's been about 15 years since I read the DNS RFCs. qtype would be better too.

collector/dns.go
```diff
@@ -0,0 +1,492 @@
...
+		ZoneTransferFailures: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "zone_transfer_failures_total"),
+			"Number of failed zone transfers of the master DNS server",
+			nil,
+			nil,
+		),
+		MemoryUsedBytes: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "memory_used_bytes_total"),
+			"Total memory used by DNS server, per type",
+			[]string{"type"},
```
carlpett: Here's another `type` label which should be changed. Possible values are caching/database_node/nbstat/record_flow/tcp_message/udp_message. 
`pool` is a common label name, and while probably not very correct it might still be better. Other suggestions? Area?

collector/dns.go
```diff
@@ -0,0 +1,492 @@
...
+		RecursiveQuerySendTimeouts: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "recursive_query_send_timeouts_total"),
+			"Number of recursive query sending timeouts",
+			nil,
+			nil,
+		),
+		WinsQueries: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "wins_queries_total"),
+			"Number of WINS lookup requests received by the server",
+			[]string{"type"},
```
carlpett: More `type`. Here it can be forward or reverse.

collector/dns.go
brian-brazil: Pool or area are both common for this sort of thing.

collector/dns.go
brian-brazil: Direction maybe?

collector/dns.go
carlpett: Sorry if that came across as rude. I'm not entierly fluent in the standard, myself...

carlpett:
@brian-brazil Updated with the new name suggestions. Anything else that could be better?
@martinlindhe Any comments?

brian-brazil:
:+1:

martinlindhe:
looks great, merging!

PR #56, opened 2017-03-03
Add Win32_Service collector

### LINKED ISSUE #55: Feature: Services states
Hi Martin,

It would be useful to get Windows Service states exposed in a similar fashion that node_exporter exposes systemd states.

i.e. (example from node_exporter 
```
node_systemd_unit_state{name="syslog.service",state="activating"} 0
node_systemd_unit_state{name="syslog.service",state="active"} 0
node_systemd_unit_state{name="syslog.service",state="deactivating"} 0
node_systemd_unit_state{name="syslog.service",state="failed"} 0
node_systemd_unit_state{name="syslog.service",state="inactive"} 1
```

Some good information at the below urls:

https://msdn.microsoft.com/en-us/library/aa394418(v=vs.85).aspx
https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/processes_windows.go

carlpett:
Fixes #55

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -25,7 +25,7 @@ type WmiCollector struct {
 }
 
 const (
-	defaultCollectors = "cpu,cs,logical_disk,net,os,system"
+	defaultCollectors = "cpu,cs,logical_disk,net,os,service,system"
 	serviceName       = "wmi_exporter"
 )
 
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ iis | [Win32_PerfRawData_W3SVC_WebService](https://msdn.microsoft.com/en-us/libr
 logical_disk | [Win32_PerfRawData_PerfDisk_LogicalDisk](https://msdn.microsoft.com/en-us/windows/hardware/aa394307(v=vs.71)) metrics (disk I/O) | &#10003;
 net | [Win32_PerfRawData_Tcpip_NetworkInterface](https://technet.microsoft.com/en-us/security/aa394340(v=vs.80)) metrics (network interface I/O) | &#10003;
 os | [Win32_OperatingSystem](https://msdn.microsoft.com/en-us/library/aa394239) metrics (memory, processes, users) | &#10003;
+service | [Win32_Service](https://msdn.microsoft.com/en-us/library/aa394418(v=vs.85).aspx) metrics (service states) | &#10003;
 system | Win32_PerfRawData_PerfOS_System metrics (system calls) | &#10003;
 
 The HELP texts shows the WMI data source, please see MSDN documentation for details.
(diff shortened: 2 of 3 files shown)


exporter.go
```diff
@@ -25,7 +25,7 @@ type WmiCollector struct {
 }
 
 const (
-	defaultCollectors = "cpu,cs,logical_disk,net,os,system"
+	defaultCollectors = "cpu,cs,logical_disk,net,os,service,system"
```
martinlindhe: @carlpett Do you feel this should be enabled by default? If so, patch looks good to me. If not, please change this :)

carlpett: I'd say so, yes, but I'm open to changing it. It adds ~400 metrics on my systems, give or take, which doesn't feel that heavy. The IIS collector creates around ~500 on our web servers. On the other hand, it would still be an increase of the total with about 40%. 

I know I'll enable it everywhere, but that doesn't mean everyone else wants it :) I'm fine either way.

martinlindhe: Let's enable it for now. I have the feeling this is still a relative lightweight monitoring solution compared to the alternatives I looked at.

martinlindhe:
AppVeyor failed, I'm not 100% it is related to this PR, or some random AppVeyor issue? 

`Exception calling ".ctor" with "3" argument(s): "End of Central Directory record could not be found."`
 
https://ci.appveyor.com/project/martinlindhe/wmi-exporter/build/job/ah1eyp67r4arwccq

carlpett:
Weird, but yeah, looks like AppVeyor. All the steps work out fine locally, including the 7z invocation.

martinlindhe:
I posted the issue over here: http://help.appveyor.com/discussions/problems/6200-merge-build-failed-where-pr-build-succeded-2-hours-ago

martinlindhe:
The build still fails. But I got a reply from them:


> Hi Martin,

> Problem is that PR (and normal build as well) does not go to [this](https://github.com/martinlindhe/wmi_exporter/blob/master/appveyor.yml#L28) code path. Only tagged build does.
>
> I did small research and I see that it is Expand-Archive throws insde build.ps1. Probably because wix-binaries.zi did not download properly.
>
> I believe you can debug from here. Useful tool is [RDP](http://help.appveyor.com/discussions/problems/6200/r?go=aHR0cHM6Ly93d3cuYXBwdmV5b3IuY29tL2RvY3MvaG93LXRvL3JkcC10by1idWlsZC13b3JrZXIv). However note that RDP session does not contain the same environment variables as build session so you may need to enter some absolute values or set $Version and $env:MSI_ARCH to debug.
>
> Please let us know if you need more help.

PR #58, opened 2017-03-04
Always expose all possible service states/start modes

carlpett:
This will make sure will always expose all the states/start modes that a service can be in, as discussed in #55 after closing.

However, implementing this has made me reconsider the discussion in the original pull-request regarding if this should be enabled by default, since it seems I made some miscalculation regarding the number of metrics. I get ~2500 on my machine, not ~400. 

This might still not be such a big deal (our Prometheus server currently ingests ~700k metrics/s, and I'm guessing it wouldn't be called a "big" installation by the core team), but still might warrant taking the discussion again. @martinlindhe, what do you think?

### DIFF
--- a/collector/service.go
+++ b/collector/service.go
@@ -55,6 +55,26 @@ type Win32_Service struct {
 	StartMode string
 }
 
+var (
+	allStates = []string{
+		"stopped",
+		"start pending",
+		"stop pending",
+		"running",
+		"continue pending",
+		"pause pending",
+		"paused",
+		"unknown",
+	}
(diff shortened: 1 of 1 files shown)


martinlindhe:
I'm ok with providing a good default out-of-the box experience so we should leave this enabled by default I think. If people want more control they can always disable unwanted metrics.
Unless this turns out to significantly increase cpu time or such, I'm ok with enabling it by default.

martinlindhe:
@carlpett feel free to tag a new release if you feel this is good-to-go

PR #59, opened 2017-03-04
Add [defaults] placeholder support to collectors.enabled

### LINKED ISSUE #44: Enable collector in addition to defafults
is there a option, to install all the default collectors + the IIS as the optional ?

carlpett:
Fixes #44.

By adding the placeholder `[defaults]` in the list sent to `-collectors.enabled`, the user can add extra collectors without having to keep the defaults enabled. This makes updates when having extra collectors enabled, since you no longer need to update the list with new default collectors. For example, our web servers currently have `-collectors.enabled "cpu,cs,iis,logical_disk,net,os,system"`, just to enable `iis`. When we update to the latest release which supports the `service´ collector, we need to update our configuration to benefit from it.

I wrote a few tests for the merging logic, but I'm not all that pleased with the amount of string juggling needed in the tests themselves. Any ideas for improvements?

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -25,8 +25,9 @@ type WmiCollector struct {
 }
 
 const (
-	defaultCollectors = "cpu,cs,logical_disk,net,os,service,system"
-	serviceName       = "wmi_exporter"
+	defaultCollectors            = "cpu,cs,logical_disk,net,os,service,system"
+	defaultCollectorsPlaceholder = "[defaults]"
+	serviceName                  = "wmi_exporter"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


martinlindhe:
Nice feature! LGTM

brian-brazil:
Over in the node exporter, we're thinking of making each collector enablement a flag - like it's done in the mysqld exporter.

martinlindhe:
Not sure we can use flags in our setup, the parameters are passed to the msi installer and limitations in there is beyond our reach: https://github.com/martinlindhe/wmi_exporter#installation

martinlindhe:
> I wrote a few tests for the merging logic, but I'm not all that pleased with the amount of string juggling needed in the tests themselves. Any ideas for improvements?

No ideas atm, looks fine

PR #67, opened 2017-03-31
Added local_time metric with time_zone label to OS collector

atezs82:
Added a new metric called wmi_os_local_time to the exporter, with which OS time on Windows can be also scraped from Prometheus (and, for instance alerts can be generated when it does not match the time of the server). The metric also contains a label called time_zone in which the time zone on Windows is stored (and can be compared accordingly with any value in Prometheus).

### DIFF
--- a/collector/os.go
+++ b/collector/os.go
@@ -5,7 +5,8 @@ package collector
 
 import (
 	"log"
-
+	"time"
+	
 	"github.com/StackExchange/wmi"
 	"github.com/prometheus/client_golang/prometheus"
 )
@@ -26,6 +27,8 @@ type OSCollector struct {
 	PagingLimitBytes        *prometheus.Desc
 	VirtualMemoryBytes      *prometheus.Desc
 	VisibleMemoryBytes      *prometheus.Desc
+	Time			*prometheus.Desc
+	Timezone		*prometheus.Desc
 }
 
 // NewOSCollector ...
@@ -49,6 +52,18 @@ func NewOSCollector() (Collector, error) {
 			prometheus.BuildFQName(Namespace, subsystem, "physical_memory_free_bytes"),
 			"OperatingSystem.FreePhysicalMemory",
 			nil,
+			nil,
+		),
+		Time: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "time"),
+			"OperatingSystem.LocalDateTime",
+			nil,
+			nil,
+		),
+		Timezone: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "timezone"),
+			"OperatingSystem.LocalDateTime",
+			[]string{"timezone"},
 			nil,
 		),
 		Processes: prometheus.NewDesc(
@@ -117,6 +132,7 @@ type Win32_OperatingSystem struct {
 	SizeStoredInPagingFiles uint64
 	TotalVirtualMemorySize  uint64
 	TotalVisibleMemorySize  uint64
+	LocalDateTime		time.Time
 }
 
 func (c *OSCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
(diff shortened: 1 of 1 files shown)


collector/os.go
```diff
@@ -131,6 +140,16 @@ func (c *OSCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, er
 		float64(dst[0].FreePhysicalMemory*1024), // KiB -> bytes
 	)
 
+	time := dst[0].LocalDateTime
+	timeZoneName, _ := time.Zone()
+	
+	ch <- prometheus.MustNewConstMetric(
+		c.LocalDateTime,
+		prometheus.GaugeValue,
+		float64(time.Unix()),
```
brian-brazil: Can you clarify, is this local time or is it unix time? If local time, what does that mean?

atezs82: This is Unix time (as described in https://golang.org/pkg/time/#Time.Unix).

brian-brazil: Local time implies that timezones are involved, which is not the case for unixtime. This should be just called "time".

The timezone should go on a separate 1-valued metric.

atezs82: You are right, see the latest updated code below. Can you do a review for that as well?

collector/os.go
```diff
@@ -51,6 +54,18 @@ func NewOSCollector() (Collector, error) {
...
 			nil,
 		),
+		Time: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "time"),
+			"OperatingSystem.LocalDateTime",
+			nil,
+			nil,
+		),
+		TimeZone: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "time_zone"),
```
brian-brazil: it'd spell it as timezone

brian-brazil:
This seems fine from a metrics standpoint, but I'll leave on of the WMI people to review and merge.

collector/os.go
```diff
@@ -26,6 +27,8 @@ type OSCollector struct {
 	PagingLimitBytes        *prometheus.Desc
 	VirtualMemoryBytes      *prometheus.Desc
 	VisibleMemoryBytes      *prometheus.Desc
+	Time			*prometheus.Desc
+	TimeZone		*prometheus.Desc
```
martinlindhe: Timezone

collector/os.go
martinlindhe: I agree with brian

collector/os.go
```diff
@@ -51,6 +54,18 @@ func NewOSCollector() (Collector, error) {
...
+		Time: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "time"),
+			"OperatingSystem.LocalDateTime",
+			nil,
+			nil,
+		),
+		TimeZone: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "time_zone"),
+			"OperatingSystem.LocalDateTime",
+			[]string{"time_zone"},
```
martinlindhe: timezone

carlpett:
Implementation-wise, this looks fine to me. Thanks @atezs82! 
One point of discussion would be if it could possibly be better to use the UTC offset in minutes rather than the name of the timezone? Seems less likely to cause localization issues and similar.

brian-brazil:
> UTC offset in minutes 

Seconds is the convention in Prometheus.

> Seems less likely to cause localization issues and similar.

There are multiple timezones that can result in the same offsets at various times of the year. For example, I'm in UTC for half the year. The full timezone name is what you need.

carlpett:
>>UTC offset in minutes
>
>Seconds is the convention in Prometheus.

Right, sloppy phrasing on my part. What I meant was that what WMI exposes is the offset in minute. It should be exposed as seconds, as you say.

>>Seems less likely to cause localization issues and similar.
>
>There are multiple timezones that can result in the same offsets at various times of the year. For example, I'm in UTC for half the year. The full timezone name is what you need.

After re-reading the code and trying a few zone/dst changes locally, I can agree to this. I had two concerns, which both seem to have been wrong:

- I initially misread as the timezone name being read from WMI. Windows localizes the timezone name, so that seemed like it would be likely to create issues for deployments in multiple languages. Go's `time.zone()` does the reasonable thing, of course.
- In the same vein of experience with poor implementations of time zones, I wasn't sure that DST changes would result in different zones. Again, Go seems to do the right thing.

So, in short, I'm fine with this.

atezs82:
Thanks for all the comments! I did my best to incorporate them. 
Also squashed the commit not to mess up the repository.
Let me know if anything else is required!

martinlindhe:
@atezs82 Thank you very much for this contribution!

PR #71, opened 2017-04-29
Use the new prometheus http API

juergenhoetzel:
https://github.com/prometheus/client_golang/pull/214

Thanks for this package! 👍

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -189,9 +189,7 @@ func main() {
 
 	go func() {
 		log.Infoln("Starting server on", *listenAddress)
-		if err := http.ListenAndServe(*listenAddress, nil); err != nil {
-			log.Fatalf("cannot start WMI exporter: %s", err)
-		}
+		log.Fatalf("cannot start WMI exporter: %s", http.ListenAndServe(*listenAddress, nil))
 	}()
 
 	for {
--- a/exporter.go
+++ b/exporter.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/martinlindhe/wmi_exporter/collector"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/version"
 )
@@ -177,7 +178,7 @@ func main() {
 	nodeCollector := WmiCollector{collectors: collectors}
 	prometheus.MustRegister(nodeCollector)
 
-	http.Handle(*metricsPath, prometheus.Handler())
+	http.Handle(*metricsPath, promhttp.Handler())
 	http.HandleFunc("/health", healthCheck)
 	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		http.Redirect(w, r, *metricsPath, http.StatusMovedPermanently)
--- a/vendor/github.com/prometheus/client_golang/prometheus/counter.go
+++ b/vendor/github.com/prometheus/client_golang/prometheus/counter.go
@@ -30,16 +30,8 @@ type Counter interface {
 	Metric
 	Collector
 
-	// Set is used to set the Counter to an arbitrary value. It is only used
-	// if you have to transfer a value from an external counter into this
-	// Prometheus metric. Do not use it for regular handling of a
-	// Prometheus counter (as it can be used to break the contract of
-	// monotonically increasing values).
-	//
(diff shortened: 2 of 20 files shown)


carlpett:
Hi @juergenhoetzel, thanks for the PR! The build fails because of missing vendoring, could you add promhttp as a vendored package?

martinlindhe:
@juergenhoetzel You'll also need to commit the changes in vendor, pr is still failing on missing promhttp

exporter.go
```diff
@@ -188,9 +189,7 @@ func main() {
 
 	go func() {
 		log.Infoln("Starting server on", *listenAddress)
-		if err := http.ListenAndServe(*listenAddress, nil); err != nil {
-			log.Fatalf("cannot start WMI exporter: %s", err)
-		}
+		log.Fatalf("cannot start WMI exporter: %s", http.ListenAndServe(*listenAddress, nil))
```
martinlindhe: Not sure this is correct.
log.Fatal() simply exits the program. You'll still need to do as before, that is checking for error.

juergenhoetzel: Hi @martinlindhe 
`http.ListenAndServer` only returns when an *error* occured. I.e. it never returns *nil*. This style is also used in the official [Go http documentation](https://golang.org/pkg/net/http/)

.gitignore
```diff
@@ -1,2 +1,3 @@
 *.exe
-VERSION
\ No newline at end of file
+VERSION
+vendor/*/
```
martinlindhe: This is wrong. We do version the vendor folder.

appveyor.yml
```diff
@@ -21,6 +21,7 @@ install:
 
 build_script:
   - ps: gitversion /output json /showvariable FullSemVer | Set-Content VERSION -PassThru
+  - govendor sync
```
martinlindhe: This is wrong. We commit the vendor folder for reproducible builds.

juergenhoetzel:
> @juergenhoetzel Good work! For the pr to land, you'll also need to commit the changes in vendor, because it is still failing on missing promhttp. Appveyour output: https://ci.appveyor.com/project/martinlindhe/wmi-exporter/build/258/job/x5ucqsbxku8huu96

@martinlindhe  Thanks for the hint. I didn't used *govendor* before. I removed all *vendored* sources and and used the *sync* functionality instead: https://github.com/kardianos/govendor .

exporter.go
martinlindhe: Ah, that's all good then! I was under the wrong impression it was non-blocking

martinlindhe:
@juergenhoetzel For vendoring, all you need to do is, manually

govendor sync
git add vendor
git commit -am "updated vendored dependencies"

This gives us a working snapshot of correct vendored dependencies,
so we have reproducible builds.
We don't want this step to happen in appveyor at a later time, when deps might have changed, and we don't want the vendor folder gitignored.

juergenhoetzel:
> This gives us a working snapshot of correct vendored dependencies,
> so we have reproducible builds.
> We don't want this step to happen in appveyor at a later time, when deps might have changed, and we don't want the vendor folder gitignored.

Make sense. I will revert/amend the change.

martinlindhe:
Feel free to just commit additional changes to this branch, we can merge the PR as a single commit when done

martinlindhe:
I seem to have forgotten how to use govendor.

I think the proper usage is something like:

```
govendor fetch github.com/prometheus/client_golang/prometheus/promhttp
git add vendor
git commit -am "updated vendored dependencies"
```

juergenhoetzel:
> I seem to have forgotten how to use govendor.
> 
> I think the proper usage is something like:
> 
> govendor fetch github.com/prometheus/client_golang/prometheus/promhttp
> git add vendor
> git commit -am "updated vendored dependencies"

Thanks for your feedback. I actually used 

```bash
govendor.exe fetch github.com/prometheus/client_golang/prometheus/...
```

to update all referenced packages under the `prometheus` path.

What started as an small `prometheus` API update ended in an excursion to the `go` package dependency nightmare 😨 

😆

martinlindhe:
For updating vendoring of a single package, you'd specify that package. I think that what all that is needed in this patch.

for updating all of the prometheus deps, you would run

    govendor.exe fetch github.com/prometheus/...

The command you used only updated the deps under github.com/prometheus/client_golang/prometheus
But I don't think that is needed.

Thanks for your work, looks good!

PR #72, opened 2017-04-30
update vendored files

martinlindhe:
Sync vendored libs. Not required, but keeps us up-to-date with prometheus changes for now.


carlpett:
Nice! Great to get some of the maintenance work fixed.
Reviewing is a bit hard, but I'll take AppVeyor's word for it being ok! :)

Issue #80, opened 2017-06-06
Feature Request:  VM Performance Counters


MattMencel:
VMware Tools adds "VM Processor" and "VM Memory" counters to the Performance Monitor.

The full list can be found here: https://rvdnieuwendijk.com/2012/06/19/use-performance-monitor-to-get-vm-performance-statistics/

I looked up the path to some of these in Performance Monitor and they appear like this:

`<PARAM NAME="Counter00002.Path" VALUE="\VM Memory\Memory Active in MB"/>`

Can these be polled with the WMI exporter?

carlpett:
Hi,
Thanks for the idea! It is indeed possible. I did a quick dig on a vmware machine I have access to, and the classes are `Win32_PerfRawData_vmGuestLib_VMem` and `Win32_PerfRawData_vmGuestLib_VCPU`. Its getting a bit late here to do it tonight, but I'll see if have some time tomorrow!

carlpett:
Pull request #81 created. The new collector is not part of the default collectors, so you'll need to enable it using the `-collectors.enabled` flag, but please try it out!

martinlindhe:
@MattMencel Please reopen if you discover issues with the vmware collector. I've pushed a new release v0.2.4 to supply binaries with this feature, which can be downloaded at https://github.com/martinlindhe/wmi_exporter/releases/tag/v0.2.4

PR #81, opened 2017-06-07
Add Vmware collector

carlpett:
This adds a Vmware collector as suggested in #80.


martinlindhe:
Very nice! Looks good to me

PR #82, opened 2017-06-26
Fix memory leak

### LINKED ISSUE #77: memory leak?
I use the latest version and it's using more ram over time. Once it was at about 500 megabytes. I'm not sure if it's a leak. Maybe the memory is reserved but not really used.

Not sure which go_memstats_  I could check for that.

carlpett:
Fixes #77


martinlindhe:
Looks good, great work!

PR #84, opened 2017-06-27
Update WiX download url

carlpett:
WiX is now on GitHub, which means stable download urls. Yay!


martinlindhe:
Yay, nicely done!

Issue #89, opened 2017-07-03
CPU collector blocks every ~17 minutes on call to wmi.Query


higels:
We've deployed the wmi-exporter on a number of our Windows systems running Server 2012r2 standard. This was a version built several months ago with Go 1.8.

These systems typically push 20-30 Gbps of traffic and have 24 hyperthreads. There is generally plenty of room on half of these threads.

Having deployed the exporter, we observed that every 17-18 minutes (independent of when the exporter was started), the traffic being served by them would drop significantly for roughly 120 seconds and CPU consumption on one or two cores would spike. The "WMI Provider Host" would also increase in CPU utilisation during this time.

Killing the exporter did not cause the load to drop off immediately.

We also noticed that calls to the exporter would block during this time.

We set GOMAXPROCS to 2, to see if there was any concurrency issues that might be overloading WMI, but that didn't improve matters.

We built a new version from master on Friday June 30th with Go 1.8.1. The issue persisted.

We started to disable collectors to isolate where the problem was occurring and found that it only happened when the CPU collector was enabled. We put some printfs around the calls inside the collect function and found that it was blocking on the call to wmi.Query.

Interestingly, we did not see any pauses or breaks in the data while monitoring that data in perfmon while the query was running.

We dialled testing back to a 2 CPU VM running Windows 10 not doing anything particularly difficult or interesting and polling using "curl $blah ; sleep 60" in a loop.

The results suggest that something is going wrong every 17-18 minutes (the lines with *** were added by me):

```
time="2017-07-03T10:31:38-07:00" level=debug msg="OK: cpu collector succeeded after 0.032996s." source="exporter.go:90"
2017/07/03 10:32:38 *** Starting CPU Collector run...
2017/07/03 10:32:38 *** Created WMI Query for CPU...
2017/07/03 10:32:40 *** Ran WMI Query for CPU...
2017/07/03 10:32:40 *** Sending data for 0 to exporter...
2017/07/03 10:32:40 *** Sending data for 1 to exporter...
time="2017-07-03T10:32:40-07:00" level=debug msg="OK: cpu collector succeeded after 2.324834s." source="exporter.go:90"
2017/07/03 10:33:40 *** Starting CPU Collector run...
2017/07/03 10:33:40 *** Created WMI Query for CPU...
2017/07/03 10:33:40 *** Ran WMI Query for CPU...
2017/07/03 10:33:40 *** Sending data for 0 to exporter...
[...]

PR #90, opened 2017-07-05
Change scrape_duration_seconds to gauge

carlpett:
Currently the `scrape_duration_seconds` metric is a summary. When trying to reason about #89, it is hard to look back in time to see if others have this problem as well, since we do not have the buckets. I see no real benefit in using a summary here, and think we should be using a histogram instead.

Probably we should be setting the `Buckets` option as well, rather than using the defaults. My lowest 0.5 quantile with the current summary is ~0.04s, so I would make an initial guess that something like `[0.01, 0.015, 0.03, 0.1, 0.15, 0.3, 1.0, 3.0, Inf]` would capture the interesting ranges with acceptable precision?

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -33,31 +33,25 @@ const (
 )
 
 var (
-	scrapeDurations = prometheus.NewGaugeVec(
-		prometheus.GaugeOpts{
-			Namespace: collector.Namespace,
-			Subsystem: "exporter",
-			Name:      "collector_duration_seconds",
-			Help:      "wmi_exporter: Duration of a collection.",
-		},
+	scrapeDurationDesc = prometheus.NewDesc(
+		prometheus.BuildFQName(collector.Namespace, "exporter", "collector_duration_seconds"),
+		"wmi_exporter: Duration of a collection.",
 		[]string{"collector"},
+		nil,
 	)
-	scrapeSuccess = prometheus.NewGaugeVec(
-		prometheus.GaugeOpts{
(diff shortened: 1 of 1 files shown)


martinlindhe:
So, the bucket gets configured by the user and not by us?

Reading up on https://prometheus.io/docs/practices/histograms/

brian-brazil:
This should be a gauge, which contains the information for that scrape.

This was fixed a while back on the node exporter, I'm guessing this was copied from before then.

carlpett:
@brian-brazil Ah, that makes sense. Yes, the summary is probably from the initial commit, more or less. 
I changed to a gauge now, as well as split the result into a separate gauge.

exporter.go
```diff
@@ -81,16 +92,15 @@ func execute(name string, c collector.Collector, ch chan<- prometheus.Metric) {
 	begin := time.Now()
 	err := c.Collect(ch)
 	duration := time.Since(begin)
-	var result string
 
 	if err != nil {
 		log.Errorf("ERROR: %s collector failed after %fs: %s", name, duration.Seconds(), err)
-		result = "error"
+		scrapeSuccess.WithLabelValues(name).Set(0)
```
brian-brazil: This should use ConstMetrics, so there's no interference from any scrapes happening at the same time

carlpett: Thanks. Fixed!

brian-brazil:
:+1:

martinlindhe:
LGTM!

PR #93, opened 2017-07-15
First implementation of AD collector

### LINKED ISSUE #43: Feature: Active Directory
Already in progress, using the `Win32_PerfRawData_DirectoryServices_DirectoryServices` counters. This will just cover the "Active Directory - Directory Services" part of the Active Directory, not the web services etc (there are over 150 counters exposed by this class alone...)

carlpett:
So, I started this way back in November, but then priorities were shifted and the branch has been collecting dust for a while. But I had an opportunity to finish it up today, so here it is. 

Fixes #43

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ Prometheus exporter for Windows machines, using the WMI (Windows Management Inst
 
 Name     | Description | Enabled by default
 ---------|-------------|--------------------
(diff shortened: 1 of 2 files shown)


martinlindhe:
Very nice! I see what you mean with need for documenting the values better, we have collected quite some now.

Merging this so it can be exposed for wider testing

PR #95, opened 2017-07-18
Add hint in usage about quoting flags in powershell

### LINKED ISSUE #91: Wrong flags in Usage helper text
I was trying to manually run `wmi_exproter.exe` with custom options, because I wanted the port to listen at 9100 (like node_exporter).

However, the flag listed in the command-line usage wasn't working, issuing the error `flag provided but not defined: -addr`:

```
PS C:\Program Files\wmi_exporter> .\wmi_exporter.exe -addr ":9100"
flag provided but not defined: -addr
Usage of C:\Program Files\wmi_exporter\wmi_exporter.exe:
  -collector.iis.app-blacklist string
        Regexp of apps to blacklist. App name must both match whitelist and not match blacklist to be included.
  -collector.iis.app-whitelist string
        Regexp of apps to whitelist. App name must both match whitelist and not match blacklist to be included. (default
 ".+")
  -collector.iis.site-blacklist string
        Regexp of sites to blacklist. Site name must both match whitelist and not match blacklist to be included.
  -collector.iis.site-whitelist string
        Regexp of sites to whitelist. Site name must both match whitelist and not match blacklist to be included. (defau
lt ".+")
  -collector.logical_disk.volume-blacklist string
        Regexp of volumes to blacklist. Volume name must both match whitelist and not match blacklist to be included.
  -collector.logical_disk.volume-whitelist string
        Regexp of volumes to whitelist. Volume name must both match whitelist and not match blacklist to be included. (d
efault ".+")
  -collector.net.nic-blacklist string
        Regexp of NIC:s to blacklist. NIC name must both match whitelist and not match blacklist to be included.
  -collector.net.nic-whitelist string
        Regexp of NIC:s to whitelist. NIC name must both match whitelist and not match blacklist to be included. (defaul
t ".+")
  -collectors.enabled string
        Comma-separated list of collectors to use. Use '[default]' as a placeholder for all the collectors enabled by de
fault (default "cpu,cs,logical_disk,net,os,service,system")
  -collectors.print
        If true, print available collectors and exit.
  -log.format value
        Set the log target and format. Example: "logger:syslog?appname=bob&local=7" or "logger:stdout?json=true" (defaul
t "logger:stderr")
  -log.level value
        Only log messages with the given severity or above. Valid levels: [debug, info, warn, error, fatal] (default "in
fo")
  -telemetry.addr string
        host:port for WMI exporter. (default ":9182")
  -telemetry.path string
        URL path for surfacing collected metrics. (default "/metrics")
  -version
        Print version information.
```

So I found out that it needed to use two hyphens (`--addr ":9100"`) instead of one (`-addr ":9100"`), for every option.

So I think a fix is needed in the Usage text.

carlpett:
Fixes #91, as a temporary measure.

@martinlindhe I started looking at just changing the flags, but of course nothing is as simple as could be expected... I'll create a new issue to discuss it.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -157,6 +157,13 @@ func initWbem() {
 	wmi.DefaultClient.SWbemServicesClient = s
 }
 
+func usage() {
+	fmt.Fprintf(os.Stderr, "Usage of %s:\n", os.Args[0])
+	flag.PrintDefaults()
+	fmt.Fprintf(os.Stderr, "\nNote: If executing from Powershell, the flags need to quoted. For example:\n%s\n",
+		"\twmi_exporter \"-collectors.enabled\" iis")
+}
+
 func main() {
 	var (
 		showVersion       = flag.Bool("version", false, "Print version information.")
@@ -165,6 +172,7 @@ func main() {
 		enabledCollectors = flag.String("collectors.enabled", filterAvailableCollectors(defaultCollectors), "Comma-separated list of collectors to use. Use '[default]' as a placeholder for all the collectors enabled by default")
 		printCollectors   = flag.Bool("collectors.print", false, "If true, print available collectors and exit.")
 	)
+	flag.Usage = usage
 	flag.Parse()
 
 	if *showVersion {


martinlindhe:
iirc the flag parsing was inherited from other prometheus projects when I started.

Personally, I've had much more success using the stdlib https://golang.org/pkg/flag/, or for more advanced usage I like https://github.com/alecthomas/kingpin

PR #100, opened 2017-08-03
Add DisplayName and Description to Windows service

### LINKED ISSUE #99: The wmi_exporter service should have a description
Hi,

The wmi_exporter service should have a description, and the display name could possibly be changed to something more friendly, for example "Prometheus WMI exporter".

Thank you

carlpett:
Fixes #99.

I opted to set the display name to just "WMI Exporter" rather than the arguable clearer "Prometheus WMI exporter" to avoid throwing people off who might be used to it being called wmi_exporter in the list. Opinions? Maybe too change-averse? :)

@lubyou

### DIFF
--- a/installer/wmi_exporter.wxs
+++ b/installer/wmi_exporter.wxs
@@ -36,7 +36,7 @@
         <File Id="wmi_exporter.exe" Name="wmi_exporter.exe" Source="Work\wmi_exporter.exe" KeyPath="yes">
           <fw:FirewallException Id="MetricsEndpoint" Name="WMI Exporter (HTTP [LISTEN_PORT])" Description="WMI Exporter HTTP endpoint" Port="[LISTEN_PORT]" Protocol="tcp" Scope="any" IgnoreFailure="yes" />
         </File>
-        <ServiceInstall Id="InstallExporterService" Name="wmi_exporter" ErrorControl="normal" Start="auto" Type="ownProcess" Arguments="-log.format logger:eventlog?name=wmi_exporter [CollectorsFlag] [ListenFlag] [MetricsPathFlag]" />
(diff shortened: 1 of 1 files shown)


martinlindhe:
LGTM!

lubyou:
Hi,

I personally prefer "Prometheus WMI Exporter", as I believe that prefixing services with the org/company name makes it easier to differentiate them from the MS services (which are not prefixed).

martinlindhe:
All of Microsoft's services have the Company name set, which allows you to find this out.

Screenshot lacks this PR, but shows the company name field of their services:

untitled

lubyou:
Hi,

What I meant is to have the company name as part of the display name, as HP, Symantec, Veeam, VMware and others seem to do for their services, as opposed to the company name in the StringFileInfo.

On numerous occasions, I sorted services in services.msc by display name to look for services from a specific vendor, where the fact that the vendor included its name in the display name was helpful.

Ultimately I am fine with either "WMI Exporter" or "Prometheus WMI Exporter", it is just a name, and ultimately the functionality is what counts, but I still believe that including "Prometheus" into the display name is useful to people who use services.msc.

martinlindhe:
> What I meant is to have the company name as part of the display name, as HP, Symantec, Veeam, VMware and others seem to do for their services, as opposed to the company name in the StringFileInfo.

Yes, I understand that.
However this does not appear to be standard as I mentioned earlier, and you would be better off sorting on the Company Name.

The thing is, we are a separate project and currently not part of Prometheus more than compatible by them. We love prometheus but we also have no right to use their name.

brian-brazil:
Using Prometheus as a descriptor is not a problem, so you could do "WMI Exporter for Prometheus" for example.

martinlindhe:
@brian-brazil thanks. That sounds like a good solution

What do you think @lubyou ?

lubyou:
I am still in favour of "Prometheus WMI Exporter", because of the sorting thing (I can sort processes by company name,  but not services)

Your call :)

martinlindhe:
"WMI exporter for Prometheus" implies we are compatible with Prometheus.
"Prometheus WMI exporter" implies we ARE Prometheus.

This is starting to become major bikeshedding...

Let @carlpett decide on final wording

Issue #105, opened 2017-08-09
Doc Request: Cannot get "-collector.process.processes-where" flag to work


vbeausoleil:
Hi,

Would it be possible to have a filter example for the "-collector.process.processes-where" flag? I can't seem to get it to work.

I get the following errors when I open localhost:9182/metrics

time="2017-08-09T16:32:16-04:00" level=error msg="[ERROR] failed collecting process metrics: <nil> Exception occurred. (Invalid query )" source="process.go:142"

time="2017-08-09T16:32:16-04:00" level=error msg="ERROR: process collector failed after 0.008049s: Exception occurred. (Invalid query )" source="exporter.go:90"

Thank you!

martinlindhe:
Indeed, this feature is pretty low level so an example would be good.

You can do something like

`.\wmi_exporter.exe "-collectors.enabled" "process" "-collector.process.processes-where" "Name='firefox'"`
or
`.\wmi_exporter.exe "-collectors.enabled" "process" "-collector.process.processes-where" "IDProcess=1128"`

The available columns are documented in https://msdn.microsoft.com/en-us/library/aa394323(v=vs.85).aspx

martinlindhe:
@vbeausoleil I hope the added example are useful, if not please reach out again!

vbeausoleil:
It works! Thank you for the quick reply @martinlindhe

Are there any plans to implement a similar filter feature for the service collector? If not, I might give it a shot when I have some free time.

martinlindhe:
Nothing planned at the moment, but if it is a useful feature then it would be very welcome!
It should be quite similar to what's already implemented in collector/process.go

PR #109, opened 2017-08-10
Service Collector Filter Flag

vbeausoleil:
Follow up to my comment in issue #105 

* Added flag to filter services (like the one for the process collector).
* Added example for said flag in README.md

### DIFF
--- a/collector/service.go
+++ b/collector/service.go
@@ -3,6 +3,8 @@
 package collector
 
 import (
+	"bytes"
+	"flag"
 	"log"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


martinlindhe:
Looks perfect, thank you very much for your contribution!

collector/service.go
```diff
@@ -14,15 +16,30 @@ func init() {
...
 // NewserviceCollector ...
 func NewserviceCollector() (Collector, error) {
 	const subsystem = "service"
+
+	var wc bytes.Buffer
+	if *serviceWhereClause != "" {
+		wc.WriteString("WHERE ")
+		wc.WriteString(*serviceWhereClause)
+	} else {
+		log.Println("warning: No where-clause specified for process collector. This will generate a very large number of metrics!")
```
carlpett: I don't think we need this warning for the service collector, there won't be that many metrics.

carlpett:
Bit late to the party here, sorry. @vbeausoleil would you be up for addressing the review point I added?

collector/service.go
martinlindhe: I think this was simply carried over from https://github.com/martinlindhe/wmi_exporter/blob/master/collector/process.go#L52

vbeausoleil:
@carlpett Sure, I can address it if you wish. 

I felt like it was still a relevant warning since every service on a machine has 8 different states that are reported and exposed as metrics so it's still feels like a lot. Initially, that is the reason why I wanted to have a filter flag in order to reduce the noise/amount of metrics I scraped.

Just let me know if you want me to open another PR to remove the warning.

carlpett:
Yes, on closer inspection you are right, there's a lot of them. So never mind then, and thanks for contributing! :)

vbeausoleil:
Great! I'm currently looking into a way to enable service filters when the exporter is installed to make mass deployment simpler. I currently have a Powershell script that stops the service, changes the binary path to include the flags and starts it back up again.

I've been thinking about adding an install parameter.. Do you guys have any opinion on the subject?

carlpett:
Sounds like a good approach. I think we might just want a blanket "extra flags" parameter to the installer rather than having a specialized one for service filters, disk filters, process filters, etc., though.

PR #112, opened 2017-08-12
Use Kingpin for flags

### LINKED ISSUE #96: Change flags to be Windows friendlier
The dotted flags we currently use (`-collectors.enabled` etc) don't play nice with how Powershell tokenizes stuff, so users need to quote the flags (`"-collectors.enabled"`). (See also #91)
Since starting the exporter from Powershell is a pretty standard use-case, platform coherence and compatibility seems more important than imitating other (primarily Linux-focused) parts of the ecosystem.

One stumbling block here is that the logging library we use (`github.com/prometheus/common/log`) defines its own flags as part of initialization. So just changing our flags would lead to a mix, which would be worse.

I'm not aware of any way to change the flags already set by another package, but maybe it is possible? Another possibility would be to have a patch on the log library, but I'm not too keen on diverging there.

carlpett:
The main Prometheus project is moving from `flag` to [Kingpin](https://github.com/alecthomas/kingpin/) for flags. For similarity, we should do so too. In addition, this fixes #96.

I have tested updating an existing installation, and we do the right thing in updating the existing service with the new flag definitions.


martinlindhe:
Looks great!
Should we merge this now and bump v1.0, or hold on a little more?
Might there be any other incomatible changes we want to include for 1.0? I can't think of any..

carlpett:
Yeah, it's quite a step up over all the previous libraries I've used :)
I think we should hold this one until we're ready to do a 1.0. I marked a few more issues with the label, which are basically "polish stuff" that I think would be good to have in place before we mark it as done. Agree?

martinlindhe:
Agree!

carlpett:
As discussed in #169, this has become urgent, so I have rebased this on master and we'll have it for 0.3 instead of 1.0 :)

carlpett:
Tested the resulting WMI, it works too, including logging to Windows Event log.

carlpett:
@martinlindhe Ok to merge? I think getting this in, along #174 and then releasing a 0.3.0 sounds like a plan?

martinlindhe:
Yes, sounds great!

PR #108, opened 2017-08-10
wip: add SMART disk monitor collector

martinlindhe:
First part to resolve #78 

In order to continue, we need to get https://github.com/StackExchange/wmi/pull/30 accepted,
because the SMART data is in a byte array and StackExchange/wmi currently doesnt support []uint8.

Next issue is that this requires to be ran as admin, or the query will return empty result.
You can check manually what data you get by issuing `gwmi -namespace root\wmi -class MSStorageDriver_ATAPISmartData`

Third, while developing the class stopped returning values multiple times, which I had to resolve by rebooting the box. I hope that was only due to my negligence. Will keep it running for a while to find out.

The current data exported looks like this:
```

# HELP wmi_smart_capability Smart capability (SMART.SmartCapability)
# TYPE wmi_smart_capability gauge
wmi_smart_capability{volume="SCSI\\Disk&Ven_Corsair&Prod_Performance_Pro\\4&26e76114&0&050000_0"} 2
wmi_smart_capability{volume="SCSI\\Disk&Ven_WDC&Prod_WD10EACS-00D6B1\\4&26e76114&0&030000_0"} 3
wmi_smart_capability{volume="SCSI\\Disk&Ven_WDC&Prod_WD20EARS-00MVWB0\\4&26e76114&0&040000_0"} 3
# HELP wmi_smart_selftest_status The self test status code (SMART.SelfTestStatus)
# TYPE wmi_smart_selftest_status gauge
wmi_smart_selftest_status{volume="SCSI\\Disk&Ven_Corsair&Prod_Performance_Pro\\4&26e76114&0&050000_0"} 0
wmi_smart_selftest_status{volume="SCSI\\Disk&Ven_WDC&Prod_WD10EACS-00D6B1\\4&26e76114&0&030000_0"} 0
wmi_smart_selftest_status{volume="SCSI\\Disk&Ven_WDC&Prod_WD20EARS-00MVWB0\\4&26e76114&0&040000_0"} 0
# HELP wmi_smart_total_time Total time used (SMART.TotalTime)
# TYPE wmi_smart_total_time gauge
wmi_smart_total_time{volume="SCSI\\Disk&Ven_Corsair&Prod_Performance_Pro\\4&26e76114&0&050000_0"} 10
wmi_smart_total_time{volume="SCSI\\Disk&Ven_WDC&Prod_WD10EACS-00D6B1\\4&26e76114&0&030000_0"} 23400
[...]

PR #113, opened 2017-08-12
Collector documentation

### LINKED ISSUE #94: Improve documentation/help strings
Currently the only documentation we have (apart from the code) is at best a link to MSDN with the fields exposed by WMI. While documentation can be hard to keep up to date, I believe it would make it easier for new users to get started with the exporter. 

There are three areas I think we can improve:
- Collector docs in the repo, linked from the main README. Initially this could simply be the TYPE/HELP strings that are generated as part of the exposition format (like [memcached_exporter](https://github.com/prometheus/memcached_exporter) for example). Which brings the next point...
- We don't have help strings for quite a large number of the metrics (I'm responsible for quite a few of those myself). This is mostly drudge-work with copy-pasting from perfmon's descriptions. 
- Some getting-started stuff, useful queries, dashboards, alerts... There is a discussion about this sort of thing [here](https://github.com/prometheus/node_exporter/pull/590) and on the dev-maillist. 

Thoughts? There is of course an initial cost of doing this, but I think the benefits and the quite low upkeep cost makes it reasonable.

carlpett:
This is a work in progress, and discussions are needed before merging. Fixes #94.

I used the `service` collector as a sample to discuss around. I included a short description of what the collector does, flags that can be used to alter its behaviour, a list of metrics with types and labels, and a section for queries and alerts. Anything else that would be useful? 

I would like to see some dashboards in a later stage, as well, but the `service` collector might not be a good case for that.


martinlindhe:
I think this is a great push in the right direction! I also added something similar in #108

martinlindhe:
Looks fine to merge I think

carlpett:
If we can agree to this format, then I'd like to add it for all the existing collectors before merge. 
One thing that did not come up in this collector but which is mentioned in #94 is the help strings. As you noted there, there isn't really a good source for the help text, since we don't always have docs from MSDN, and sometimes the data isn't actually what their docs indicate (mostly because there is another layer of post-processing in perfmon that we bypass), and sometimes we manipulate or even merge multiple fields. I think the ideal would be for us to have something which is actually helpful (well, obviously...) and indicates what it is in the context that we present it. Might be a lot of work to figure out, but would increase quality quite a bit.

martinlindhe:
The format is fine, I'll update #108 docs with it before merging

martinlindhe:
Also, I think we can incrementally improve HELP strings over the next couple weeks/months.

carlpett:
Year-and-a-bit later, finally added docs for all collectors. Anything to adjust?

martinlindhe:
Looks great! Lets merge this finally :-)

PR #126, opened 2017-10-12
Add .Net Framework Perf Collectors

gboucher90:
I used the collector generator tool + a bit of manual edit to add the process name label.
Those metrics are quite useful when working with .Net, especially the memory and exceptions ones.

### DIFF
--- a/.promu.yml
+++ b/.promu.yml
@@ -1,6 +1,8 @@
 repository:
     path: github.com/martinlindhe/wmi_exporter
 build:
+    binaries:
+        - name: wmi_exporter
     ldflags: |
         -X {{repoPath}}/vendor/github.com/prometheus/common/version.Version={{.Version}}
         -X {{repoPath}}/vendor/github.com/prometheus/common/version.Revision={{.Revision}}
--- a/Makefile
+++ b/Makefile
@@ -2,4 +2,4 @@ fmt:
 	gofmt -l -w -s .
 
 build:
-	promu build -v .
+	promu build -v
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,7 +22,7 @@ install:
 build_script:
   - ps: gitversion /output json /showvariable FullSemVer | Set-Content VERSION -PassThru
   - govendor test -v +local
-  - promu build -v .
+  - promu build -v
   - ps: |
       $ErrorActionPreference = "Stop"
       if($env:APPVEYOR_REPO_TAG -eq "True") {


carlpett:
Hi @gboucher90,
Thank you very much! Mostly it looks very good, there are just a few naming/metric type things I think should be aligned with the rest of the exporter. I'll leave the comments inline in a moment!

collector/netframework_clrexceptions.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/StackExchange/wmi"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	Factories["netframework_clrexceptions"] = NewWin32_PerfRawData_NETFramework_NETCLRExceptionsCollector
+}
+
+// A Win32_PerfRawData_NETFramework_NETCLRExceptionsCollector is a Prometheus collector for WMI Win32_PerfRawData_NETFramework_NETCLRExceptions metrics
+type Win32_PerfRawData_NETFramework_NETCLRExceptionsCollector struct {
```
carlpett: You can leave out the `Win32_PerfRawData_` bit in the beginning.

collector/netframework_clrexceptions.go
```diff
@@ -0,0 +1,132 @@
...
+	NumberofExcepsThrown       *prometheus.Desc
+	NumberofExcepsThrownPersec *prometheus.Desc
+	NumberofFiltersPersec      *prometheus.Desc
+	NumberofFinallysPersec     *prometheus.Desc
+	ThrowToCatchDepthPersec    *prometheus.Desc
```
carlpett: The struct field names can have some more descriptive names, rather than just the WMI field name, so `NumberOfExceptionsThrown`, `ThrowToCatchDepth`, and so on.

The `PerSec` suffix can always be removed, and is even misleading (in the PerfRaw counters, they are actually the number of times the counter has been incremented, not the rate)

collector/netframework_clrexceptions.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// NewWin32_PerfRawData_NETFramework_NETCLRExceptionsCollector ...
+func NewWin32_PerfRawData_NETFramework_NETCLRExceptionsCollector() (Collector, error) {
+	const subsystem = "win32_perfrawdata_netframework_netclrexceptions"
```
carlpett: Leave out the `win32_perfrawdata_` here as well

carlpett: Perhaps even shorten to just `netframework_exceptions`?

collector/netframework_clrexceptions.go
```diff
@@ -0,0 +1,132 @@
...
+	return &Win32_PerfRawData_NETFramework_NETCLRExceptionsCollector{
+		NumberofExcepsThrown: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "numberof_exceps_thrown"),
```
carlpett: Just like the field names, the metric name should have an as clear name as possible, as well as conform to the Prometheus [naming best practices](https://prometheus.io/docs/practices/naming/). So `exceptions_thrown_total` would be one potential name here.

collector/netframework_clrexceptions.go
```diff
@@ -0,0 +1,132 @@
...
+
+	for _, process := range dst {
+
+		if process.Name == "_Global_" {
+			continue
+		}
+
+		ch <- prometheus.MustNewConstMetric(
+			c.NumberofExcepsThrown,
+			prometheus.GaugeValue,
```
carlpett: These are probably Counters, not Gauges.

collector/netframework_clrexceptions.go
gboucher90: Yes good point! I changed it to "netframework_clrexceptions"

collector/netframework_clrexceptions.go
gboucher90: yes ok done

carlpett:
Nice progress! Maybe I wasn't very clear, I only added a comment on the first occurrence of each type of change needed. So metric types and namings are changes I'd like to see through the entire PR. Let me know if it is not clear what these changes are!

gboucher90:
No it was clear =) I just haven't had time to make more progress so I decided to push what I already did (fixing collector/struct names for all the files and the for loop) and then continue later (probably on Monday). Sorry for the delay, I will let you know when I think I'm finished ;)

PR #128, opened 2017-10-16
Add .Net Performance counters

gboucher90:


gboucher90:
Ok I think I have taken into considerations all the remarks you made. Let me know. Thank you!

gboucher90:
Hello @carlpett, did you get a chance to look at the pr?

carlpett:
@gboucher90 Sorry, been travelling the last couple of days and had very little free time. I've got a short window until my next meeting now though, let's see if it is enough :) Stay tuned!

collector/netframework_clrmemory.go
```diff
@@ -0,0 +1,388 @@
...
+			nil,
+		),
+		Gen0heapsize: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "gen0heap_size_bytes"),
+			"Displays the maximum bytes that can be allocated in generation 0; it does not indicate the current number of bytes allocated in generation 0.",
+			[]string{"process"},
+			nil,
+		),
+		Gen0PromotedBytesPerSec: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "gen0_promoted_bytes_per_sec"),
```
carlpett: This and the corresponding Gen1 are per_sec-style counters that also need new names, probably `gen0_promoted_bytes_total`?

collector/netframework_clrjit.go
```diff
@@ -0,0 +1,119 @@
...
+	const subsystem = "netframework_clrjit"
+	return &NETFramework_NETCLRJitCollector{
+		NumberofMethodsJitted: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "jit_methods_total"),
+			"Displays the total number of methods JIT-compiled since the application started. This counter does not include pre-JIT-compiled methods.",
+			[]string{"process"},
+			nil,
+		),
+		PercentTimeinJit: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "jit_timein_percent"),
```
carlpett: Did you check if this is actually reported in percent? The raw data of most other things named percent have actually reported the amount of time spent, and then perfmon calculates the percentage for display.

collector/netframework_clrmemory.go
```diff
@@ -0,0 +1,388 @@
...
+			[]string{"process"},
+			nil,
+		),
+		NumberofSinkBlocksinuse: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "number_sink_blocksinuse"),
+			"Displays the current number of synchronization blocks in use. Synchronization blocks are per-object data structures allocated for storing synchronization information. They hold weak references to managed objects and must be scanned by the garbage collector.",
+			[]string{"process"},
+			nil,
+		),
+		NumberTotalcommittedBytes: prometheus.NewDesc(
```
carlpett: Nitpicking (and probably a fault in the generator?): Capital C in Committed (and same for reserved below)

carlpett:
In general, _really_ good! I left a few minor things I found which would be nice to fix.
Also, maybe it could be a good thing to have the possibility to filter which processes you want? Take a look at the `Process` collector for an example.

gboucher90:
Thank you for this feedback =)
I will update the PR early this week to fix all these issues!

collector/netframework_clrmemory.go
gboucher90: Ah missed that one! Thank you

collector/netframework_clrjit.go
gboucher90: Yes indeed it represent the amount of time spent and not the percent? I'm fixing this one and the GC Time one

collector/netframework_clrmemory.go
gboucher90: Yes simply removed the "per_sec" suffix to keep it coherent with what I choosed with the similar perf counters

gboucher90:
Ok it is updated ;)
I completely agree for the per process filtering! Though do you mind if we do it in another PR?

carlpett:
Super, I'll take a look later today! 
Also, adding it in a later PR is fine :)

collector/netframework_clrjit.go
```diff
@@ -0,0 +1,119 @@
...
+	const subsystem = "netframework_clrjit"
+	return &NETFramework_NETCLRJitCollector{
+		NumberofMethodsJitted: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "jit_methods_total"),
+			"Displays the total number of methods JIT-compiled since the application started. This counter does not include pre-JIT-compiled methods.",
+			[]string{"process"},
+			nil,
+		),
+		TimeinJit: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "jit_timein"),
```
carlpett: Should have a unit, so `jit_time_seconds_total`?

collector/netframework_clrmemory.go
```diff
@@ -0,0 +1,349 @@
...
+			nil,
+		),
+		NumberTotalreservedBytes: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "reserved_bytes"),
+			"Displays the amount of virtual memory, in bytes, currently reserved by the garbage collector. Reserved memory is the virtual memory space reserved for the application when no disk or main memory pages have been used.",
+			[]string{"process"},
+			nil,
+		),
+		TimeinGC: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "gc_timein"),
```
carlpett: Like the JIT counter, `gc_time_seconds_total`

collector/netframework_clrsecurity.go
```diff
@@ -0,0 +1,118 @@
...
+	const subsystem = "netframework_clrsecurity"
+	return &NETFramework_NETCLRSecurityCollector{
+		NumberLinkTimeChecks: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "link_time_checks_total"),
+			"Displays the total number of link-time code access security checks since the application started.",
+			[]string{"process"},
+			nil,
+		),
+		TimeinRTchecks: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "rt_checks_timein"),
```
carlpett: `rt_checks_time_seconds_total`

carlpett:
This is really good, many thanks for a very nice contribution!

Just one more thing that struck me... I did a comparison with some Java examples, and in that case they put the heap/gc stats in the same metric separated by a label `area`. I think that could make the metrics easier to use in Grafana dashboards, splitting memory usage by generation in a single query rather than having one per gen+loh. What do you think?

Sorry that this is taking so long to get merged and that I'm bringing up new things along the way!

collector/netframework_clrmemory.go
gboucher90: Yes but this one seems to be the time of the last GC and not the total time spent doing GC so I would remove the "total" suffix. I need to double check.

collector/netframework_clrjit.go
gboucher90: I need to double check what this one represents, I though it was the time spent during the last window and not the total time spent since the beginning of the app.

collector/netframework_clrsecurity.go
gboucher90: Similarly to the the others I need to double check the meaning.

gboucher90:
That's perfectly fine, there is no rush, quality first!

You mean like those:
java_memory_usage_committed_bytes{area="Heap",} 1.489731584E9
java_memory_usage_committed_bytes{area="NonHeap",} 3.9223296E8
java_memory_usage_max_bytes{area="Heap",} 4.294770688E9
java_memory_usage_max_bytes{area="NonHeap",} 4.17333248E8

We would have for example:
heap_size_bytes{area="Gen0"}
heap_size_bytes{area="Gen1"}
heap_size_bytes{area="Gen2"}
heap_size_bytes{area="LOH"}
heap_size_bytes{area="All"}

collections_total{area="Gen0"}
collections_total{area="Gen1"}
collections_total{area="Gen2"}

It seems like a good idea to me! Our dashboards often split graphs per generation for a specific metric. I do not see any down side doing it this way. I'm in favor.

brian-brazil:
Those looks like JMX metrics to me, you should look at the Prometheus Java client and JMX exporter.

carlpett:
```heap_size_bytes{area="Gen0"}
heap_size_bytes{area="Gen1"}
heap_size_bytes{area="Gen2"}
heap_size_bytes{area="LOH"}
heap_size_bytes{area="All"}```
Yes, except not the "All" one - that one will be available from `sum(heap_size_bytes)`.

collector/netframework_clrmemory.go
gboucher90: It is not monotonically increasing, so definitely a gauge, the value is > 100 so not a percentage and too high to be seconds. It could be microseconds or simply a number of cycles.

gboucher90:
So I introduced the "area" label for GC related metrics.

I'm still not sure about the "time_in" kind of metric. Unless you have an idea, maybe I should switch to the Win32_PerfFormattedData_NETFramework_X ?

carlpett:
I found this description in [the docs](https://msdn.microsoft.com/en-us/library/x2tyfybc(v=vs.110).aspx):
> Displays the percentage of elapsed time that was  spent performing a garbage collection since the last garbage collection cycle. This counter usually indicates the work done by the garbage collector to collect and compact memory on behalf of the application.  This counter is updated only at the end of every garbage collection.  This counter is not an average; its value reflects the last observed value.

Based on this, I think the returned value will be the amount of time spent in GC (or JIT/RT for the others), but you need to normalize on the "time since last cycle". This value is returned in the `Frequency_Object` field on the WMI class (you can just add it to your structs if you don't currently query for it. So the value of the metric would be something like `TimeInGC / Frequency_Object`.
Does the data make more sense then?

I would prefer if we don't use the formatted classes, as there have been _some_ processing applied to them, but it is not always clear what (Perfmon still does additional computations to display them). So it is easier to use them wrong.

gboucher90:
Thank you!

I finally made it work using "Frequency_PerfTime" ("Frequency_Object" seemed to be always equal to 0). The data now makes sense based on my GC simulation.

carlpett:
Hurray! Great work 🥇

PR #132, opened 2017-11-11
fix: IIS Collector fails, when Application has more than one worker processes

icanit:
This pull request fixes issue described here: https://github.com/martinlindhe/wmi_exporter/issues/131

### DIFF
--- a/collector/iis.go
+++ b/collector/iis.go
@@ -909,7 +909,7 @@ type Win32_PerfRawData_W3SVCW3WPCounterProvider_W3SVCW3WP struct {
 	OutputCacheTotalFlushes        uint64
 	OutputCacheTotalHits           uint64
 	OutputCacheTotalMisses         uint64
-	TotalFilesCached               	uint64
+	TotalFilesCached               uint64
 	TotalFlushedFiles              uint64
 	TotalFlushedMetadata           uint64
 	TotalFlushedURIs               uint64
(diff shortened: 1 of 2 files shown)


collector/iis.go
```diff
@@ -909,7 +909,7 @@ type Win32_PerfRawData_W3SVCW3WPCounterProvider_W3SVCW3WP struct {
 	OutputCacheTotalFlushes        uint64
 	OutputCacheTotalHits           uint64
 	OutputCacheTotalMisses         uint64
-	TotalFilesCached               uint64
+	TotalFilesCached               	uint64
```
carlpett: Please run `go format` to avoid these accidental indents

collector/iis.go
```diff
@@ -991,7 +991,7 @@ var ApplicationStates = map[uint32]string{
 }
 
 // W3SVCW3WPCounterProvider_W3SVCW3WP returns names prefixed with pid
-var workerProcessNameExtractor = regexp.MustCompile(`^\d+_(.+)$`)
+var workerProcessNameExtractor = regexp.MustCompile(`^(\d+)$_(.+)$`)
```
carlpett: This looks strange - the regex seems like it couldn't work? There are two "end of string" (`$`), so it shouldn't ever match

carlpett:
Have a look at the regexp, other than that it looks pretty good.
As I mentioned in the original issue, we could also aggregate across the workers instead. Most of them should be simple, just sum them up. There seem to be one or two where it would be max instead. 
I don't have a strong opinion either way - @martinlindhe any thoughts?

collector/iis.go
martinlindhe: A tip about regex: Check out https://regex101.com/, in the left menu select "Golang" flavor.

martinlindhe:
I am also ok with either way of solving this for now. The way this PR does it is fine by me

carlpett:
Nice!

PR #142, opened 2017-12-18
Add TCP collector

JorritSalverda:
For retrieving metrics about TCP connections.

### DIFF
--- a/collector/tcp.go
+++ b/collector/tcp.go
@@ -124,7 +124,7 @@ func NewTCPCollector() (Collector, error) {
 			nil,
 		),
 		TimestampSys100NS: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "frequency_sys100ns"),
+			prometheus.BuildFQName(Namespace, subsystem, "timestamp_sys100ns"),
 			"(TCP.TimestampSys100NS)",
 			nil,
 			nil,
--- a/collector/tcp.go
+++ b/collector/tcp.go
@@ -143,17 +143,14 @@ func (c *TCPCollector) Collect(ch chan<- prometheus.Metric) error {
 }
 
 type Win32_PerfRawData_Tcpip_TCPv4 struct {
-	Caption                     string
 	ConnectionFailures          uint64
 	ConnectionsActive           uint64
 	ConnectionsEstablished      uint64
 	ConnectionsPassive          uint64
 	ConnectionsReset            uint64
-	Description                 string
 	Frequency_Object            uint64
 	Frequency_PerfTime          uint64
 	Frequency_Sys100NS          uint64
-	Name                        string
 	SegmentsPersec              uint64
 	SegmentsReceivedPersec      uint64
 	SegmentsRetransmittedPersec uint64
--- a/collector/tcp.go
+++ b/collector/tcp.go
@@ -1,0 +1,252 @@
+// returns data points from Win32_PerfRawData_Tcpip_TCPv4
+
+// https://msdn.microsoft.com/en-us/library/aa394341(v=vs.85).aspx (Win32_PerfRawData_Tcpip_TCPv4 class)
+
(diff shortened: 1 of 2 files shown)


collector/tcp.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// A TCPCollector is a Prometheus collector for WMI Win32_PerfRawData_Tcpip_TCPv4 metrics
+type TCPCollector struct {
+	ConnectionFailures          *prometheus.Desc
+	ConnectionsActive           *prometheus.Desc
+	ConnectionsEstablished      *prometheus.Desc
+	ConnectionsPassive          *prometheus.Desc
+	ConnectionsReset            *prometheus.Desc
+	FrequencyObject             *prometheus.Desc
```
carlpett: These three Frequency* fields should be removed. They are not valuable as metrics in themselves, they are just useful to calculate the values of other metrics.

collector/tcp.go
```diff
@@ -0,0 +1,249 @@
...
+	FrequencyPerfTime           *prometheus.Desc
+	FrequencySys100NS           *prometheus.Desc
+	SegmentsPerSec              *prometheus.Desc
+	SegmentsReceivedPerSec      *prometheus.Desc
+	SegmentsRetransmittedPerSec *prometheus.Desc
+	SegmentsSentPerSec          *prometheus.Desc
+	TimestampObject             *prometheus.Desc
```
carlpett: Same as with Frequency*, the Timestamp* ones are also just for correlating, and can be removed from here

collector/tcp.go
```diff
@@ -0,0 +1,249 @@
...
+type TCPCollector struct {
+	ConnectionFailures          *prometheus.Desc
+	ConnectionsActive           *prometheus.Desc
+	ConnectionsEstablished      *prometheus.Desc
+	ConnectionsPassive          *prometheus.Desc
+	ConnectionsReset            *prometheus.Desc
+	FrequencyObject             *prometheus.Desc
+	FrequencyPerfTime           *prometheus.Desc
+	FrequencySys100NS           *prometheus.Desc
+	SegmentsPerSec              *prometheus.Desc
```
carlpett: *PerSec is actually a misleading name in WMI - these count the total number, not the per-second rate (The reason for the name is probably that they can be _used_ to calculate the rate). So replacing "PerSec" with "Total" is better here.

collector/tcp.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// Counters
+	ch <- prometheus.MustNewConstMetric(
+		c.ConnectionFailures,
+		prometheus.CounterValue,
+		float64(dst[0].ConnectionFailures),
+	)
+	ch <- prometheus.MustNewConstMetric(
+		c.ConnectionsActive,
+		prometheus.CounterValue,
```
carlpett: Is this actually a counter? I would have expected a gauge?

carlpett:
Thanks @JorritSalverda!
There are a few minor fixes needed, but overall looks very good! We should probably document them, as they are not all that intuitive

collector/tcp.go
JorritSalverda: Strangely it is a counter. I graphed them overnight as a gauge, but that resulted in eternally climbing values. Or is that because I indicated it to be a counter in code?

JorritSalverda: When reading https://msdn.microsoft.com/en-us/library/aa394341(v=vs.85).aspx the Connection*** values seem to be counters, but it's open for debate. I copied the use of CounterValue from the `net` collector.

Treating them as counters results in a graph like this, which seems about right:

tcp-counters

carlpett: Huh... Oh well, it is not the first time that things have names that don't really match. Good that you checked!

carlpett:
Very nice, thanks a lot!

PR #166, opened 2018-03-13
Use dep instead of govendor

### LINKED ISSUE #121: tooling: use dep
I've started the process of using [dep](https://github.com/golang/dep) for vendoring in some of my golang projects, and want to bring it up here.

Pros for using `dep` as compared to other solutions:
* Is on the way to becoming merged in `go`, possibly around go 1.10.
* `dep prune` removes unnessecary files from the `vendor` folder such as text files, making it smaller
* Many others, see the project homepage

Cons:
* The `dep` tool requires go 1.8+, but is only required to use for updating vendored dependencies.

We currently use govendor iirc.

The `vendor` folder is automatically used as of go 1.6 (or 1.5 using the GOVENDOREXPERIMENT=1 env)

Any thoughts about this @carlpett or @brian-brazil ?

For example, does any parts of the rest of prometheus ecosystem embrace `dep`, or any other vendoring tools?

hairyhenderson:
Fixes #121 

#121 wasn't super clear on whether a decision had been made, but the transition is quite simple, so this PR converts from `govendor` to `dep`.

I basically just:
- ran `dep init` and tweaked `Gopkg.toml` to be less full of default comments
- edited `README.md` and `appveyor.yml` to swap mentions of `govendor` with `dep`

The changes in `vendor` should only be cosmetic, due to the slight difference in which files are considered for vendoring between `dep` and `govendor`. The only code that got updated was for `github.com/go-ole/go-ole` - I'm not sure why but it's not breaking 🤷‍♂️.

In a later PR, if this is accepted, I can submit updates to packages (from a `dep ensure -update` command).

/cc @martinlindhe 

Signed-off-by: Dave Henderson <dhenderson@gmail.com>


Gopkg.toml
```diff
@@ -0,0 +1,3 @@
+[prune]
+  go-tests = true
```
martinlindhe: If we're going with go dep (see other comment), I would prefer the config `non-go = true` to reduce the weight of the vendor folder.

martinlindhe:
Nice work!
But I'm not sure we decided on using `dep`.

Also, due to recent developments in golang vendoring (https://research.swtch.com/vgo), perhaps we should wait a little bit longer until things settle.

(I am not sure the "dep" tool will be the official vendoring tool).

Gopkg.toml
hairyhenderson: I'd done that at first, but the changes were more severe 😉.

I'll alter that for now.

hairyhenderson:
@martinlindhe yeah, `vgo` complicates things to be sure. But from what I understand the message is still "migrate to `dep` first, then it'll be much easier to migrate to `vgo` later". I'll see if I can find a reference for this...

hairyhenderson:
Just to clarify, from https://github.com/golang/go/wiki/vgo#current-state:

> For any production workloads, use dep, or migrate to it if you have not done so already.

Gopkg.toml
martinlindhe: No worries about that :)

martinlindhe:
Sounds good enough to me. Lets get this merged then. Could you just address the code review comment first?

hairyhenderson:
@martinlindhe done! PTAL 🙂

martinlindhe:
Looks good. Will update appveyor.yml to run tests again after this merge.

martinlindhe:
I followed up the appveyor testing with https://github.com/martinlindhe/wmi_exporter/commit/bda7dd18cfab45a12b8ef054118efec70ddd0ea7

thanks for your contribution!

hairyhenderson:
You're welcome! 😉 I've updated the deps in #167

Issue #169, opened 2018-03-14
0.2.11 failed to start


karl-johan-grahn:
When installing the latest release https://github.com/martinlindhe/wmi_exporter/releases/download/v0.2.11/wmi_exporter-0.2.11-amd64.msi I get this error:
![image](https://user-images.githubusercontent.com/6355577/37391678-2de4c62c-276c-11e8-8b3d-80d1526bc8a1.png)

martinlindhe:
Can confirm issue. In the meantime you could use 0.2.10, there should be no functional changes.

martinlindhe:
Quick update:

- wmi_exporter.exe is correctly built.

- event viewer logs the following:
    - A timeout was reached (30000 milliseconds) while waiting for the WMI exporter service to connect.
    - The WMI exporter service failed to start due to the following error: 
The service did not respond to the start or control request in a timely fashion.

Currently investigating wix things.

carlpett:
I'm away on a business trip so limited ability to assist, but some thoughts to investigate:
- Does it work if launched "manually" from the command line? 
- Any interesting vendor changes in the WBEM libs?

martinlindhe:
@carlpett Hey, I was just about to reach out to you! :-)

wmi_exporter.exe runs correctly if launched directly.
I thought at first the issue was the dep bump yesterday, but I have now rebuilt the .msi on older commits with the same result.

Example:

    git checkout b9b8cfd1ca08892a16a37cfab22931b0e134db38   # the commit for tag v0.2.10
    .\installer\build.ps1 -PathToExecutable .\wmi_exporter.exe -Version "0.999.999" -Arch "amd64"

the resulting .msi fails the same. So I think we can rule out the dep bump.

In order to get anywhere I had to update build.ps1 due to https://blog.github.com/2018-02-23-weak-cryptographic-standards-removed/, commit https://github.com/martinlindhe/wmi_exporter/commit/191debeed6b34eb595f1c2ca234536ceee5e00e8

martinlindhe:
More details:

the wix-binaries.zip sha256 is confirmed to be unchanged.
My local download has hash da034c489bd1dd6d8e1623675bf5e899f32d74d6d8312f8dd125a084543193de, which can also be found in https://github.com/golang/build/blob/master/cmd/release/releaselet.go#L298

martinlindhe:
[...]

PR #174, opened 2018-03-20
Bring the textfile collector over from node_exporter

poblahblahblah:
This adds a slightly modified textfile collector from the official
node_exporter project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ process | [Win32_PerfRawData_PerfProc_Process](https://msdn.microsoft.com/en-us/
 service | [Win32_Service](https://msdn.microsoft.com/en-us/library/aa394418(v=vs.85).aspx) metrics (service states) | &#10003;
 system | Win32_PerfRawData_PerfOS_System metrics (system calls) | &#10003;
 tcp | [Win32_PerfRawData_Tcpip_TCPv4](https://msdn.microsoft.com/en-us/library/aa394341(v=vs.85).aspx) metrics (tcp connections) |
+textfile | Read prometheus metrics from a text file | &#10003;
 vmware | Performance counters installed by the Vmware Guest agent |
 
 The HELP texts shows the WMI data source, please see MSDN documentation for details.
@@ -39,6 +40,7 @@ Name | Description
 `LISTEN_ADDR` | The IP address to bind to. Defaults to 0.0.0.0
 `LISTEN_PORT` | The port to bind to. Defaults to 9182.
 `METRICS_PATH` | The path at which to serve metrics. Defaults to `/metrics`
+`TEXTFILE_DIR` | As the `-collector.textfile.directory` flag, provide a directory to read text files with metrics from
 
 Parameters are sent to the installer via `msiexec`. Example invocation:
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Name | Description
 `LISTEN_ADDR` | The IP address to bind to. Defaults to 0.0.0.0
 `LISTEN_PORT` | The port to bind to. Defaults to 9182.
 `METRICS_PATH` | The path at which to serve metrics. Defaults to `/metrics`
-`TEXTFILE_DIR` | As the `-collector.textfile.directory` flag, provide a directory to read text files with metrics from
+`TEXTFILE_DIR` | As the `--collector.textfile.directory` flag, provide a directory to read text files with metrics from
 
 Parameters are sent to the installer via `msiexec`. Example invocation:
 
--- a/collector/textfile.go
+++ b/collector/textfile.go
@@ -1,0 +1,253 @@
+// Copyright 2015 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !notextfile
+
+package collector
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/log"
+)
+
+var (
+	textFileDirectory = flag.String("collector.textfile.directory", "C:\\Program Files\\wmi_exporter\\textfile_inputs", "Directory to read text files with metrics from.")
+	textFileAddOnce   sync.Once
+	mtimeDesc         = prometheus.NewDesc(
+		"wmi_textfile_mtime_seconds",
+		"Unixtime mtime of textfiles successfully read.",
+		[]string{"file"},
+		nil,
+	)
+)
+
+type textFileCollector struct {
+	path string
+	// Only set for testing to get predictable output.
+	mtime *float64
+}
+
+func init() {
+	Factories["textfile"] = NewTextFileCollector
+}
+
+// NewTextFileCollector returns a new Collector exposing metrics read from files
+// in the given textfile directory.
+func NewTextFileCollector() (Collector, error) {
+	return &textFileCollector{
+		path: *textFileDirectory,
+	}, nil
+}
+
+func convertMetricFamily(metricFamily *dto.MetricFamily, ch chan<- prometheus.Metric) {
+	var valType prometheus.ValueType
+	var val float64
+
+	allLabelNames := map[string]struct{}{}
+	for _, metric := range metricFamily.Metric {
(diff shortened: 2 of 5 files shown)


poblahblahblah:
I am not sure how to test creating and using the MSI installer, so I am fairly confident that it's not correct. Any tips on how to test this locally would be much appreciated and I can follow up with a PR for the README.

collector/textfile.go
```diff
@@ -0,0 +1,252 @@
...
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/log"
+)
+
+var (
+	textFileDirectory = flag.String("collector.textfile.directory", "C:\\Program Files\\prometheus\\textfile_inputs", "Directory to read text files with metrics from.")
```
poblahblahblah: Is this a sane default?

exporter.go
```diff
@@ -27,7 +27,7 @@ type WmiCollector struct {
 }
 
 const (
-	defaultCollectors            = "cpu,cs,logical_disk,net,os,service,system"
+	defaultCollectors            = "cpu,cs,logical_disk,net,os,service,system,textfile"
```
poblahblahblah: Perhaps we don't want this as a default?

collector/textfile.go
carlpett: We default the installation to `C:\Program Files\wmi_exporter` so it would be nice to use that, instead of requiring a new directory here.

collector/textfile.go
```diff
@@ -0,0 +1,252 @@
...
+	textFileAddOnce   sync.Once
+	mtimeDesc         = prometheus.NewDesc(
+		"node_textfile_mtime_seconds",
```
carlpett: I don't think we should be putting metrics inside the `node` namespace, even ones that are supposed to be compatible? Seems like a dangerous route to go down...
That is, I think this should be `wmi_...` instead.

collector/textfile.go
```diff
@@ -0,0 +1,252 @@
...
+					*metricFamily.Name,
+					metricFamily.GetHelp(),
+					names, nil,
+				),
+				metric.Histogram.GetSampleCount(),
+				metric.Histogram.GetSampleSum(),
+				buckets, values...,
+			)
+		default:
+			panic("unknown metric type")
```
carlpett: I know this probably comes from the node_exporter code, but should we really panic? I don't like the idea of crashing the exporter just because one of the data sources are bad. 
I would prefer just logging this, possibly increasing some diagnostic metric, and then `continue`.

collector/textfile.go
```diff
@@ -0,0 +1,252 @@
...
+			convertMetricFamily(mf, ch)
+		}
+	}
+
+	c.exportMTimes(mtimes, ch)
+
+	// Export if there were errors.
+	ch <- prometheus.MustNewConstMetric(
+		prometheus.NewDesc(
+			"node_textfile_scrape_error",
```
carlpett: As mentioned previously, I don't think this should be in the `node` namespace.

collector/textfile_test.go
```diff
@@ -0,0 +1,126 @@
...
+	for i, test := range tests {
+		mtime := 1.0
+		c := &textFileCollector{
+			path:  test.path,
+			mtime: &mtime,
+		}
+
+		// Suppress a log message about `nonexistent_path` not existing, this is
+		// expected and clutters the test output.
+		log.AddFlags(kingpin.CommandLine)
```
carlpett: We haven't switched the wmi_exporter to kingpin for flags yet, so this probably doesn't make sense

collector/textfile_test.go
```diff
@@ -0,0 +1,126 @@
...
+		panic(fmt.Sprintf("failed to update collector: %v", err))
+	}
+}
+
+func TestTextfileCollector(t *testing.T) {
+	tests := []struct {
+		path string
+		out  string
+	}{
+		{
```
carlpett: I believe these files need to exist for the tests to pass?

README.md
```diff
@@ -22,6 +22,7 @@ process | [Win32_PerfRawData_PerfProc_Process](https://msdn.microsoft.com/en-us/
 service | [Win32_Service](https://msdn.microsoft.com/en-us/library/aa394418(v=vs.85).aspx) metrics (service states) | &#10003;
 system | Win32_PerfRawData_PerfOS_System metrics (system calls) | &#10003;
 tcp | [Win32_PerfRawData_Tcpip_TCPv4](https://msdn.microsoft.com/en-us/library/aa394341(v=vs.85).aspx) metrics (tcp connections) |
+textfile | Read prometheus metrics from a text file |
```
carlpett: Since the collector is added to the defaults further down, it should be marked as such here as well (or not be enabled by default)

README.md
```diff
@@ -39,6 +40,7 @@ Name | Description
 `LISTEN_ADDR` | The IP address to bind to. Defaults to 0.0.0.0
 `LISTEN_PORT` | The port to bind to. Defaults to 9182.
 `METRICS_PATH` | The path at which to serve metrics. Defaults to `/metrics`
+`TEXTFILE_DIR` | As the `--collector.textfile.directory` flag, provide a directory to read text files with metrics from
```
carlpett: Currently should just be one `-`, ie `-collector.textfile.directory`, since we haven't switched to Kingpin yet

carlpett:
There's a compilation failure in the tests, see here: https://ci.appveyor.com/project/martinlindhe/wmi-exporter/build/402/job/2l6kquvy0c8s6ln7.

carlpett:
Regarding the MSI, assuming you have compiled the exporter already, you can build it by running `./installer/build.ps1 -PathToExecutable ./wmi_exporter.exe -Version 0.2.99`. This will build an msi that goes into `./installer/Output/`, and you can test it from there. It should be fairly well-behaved, so just uninstall as usual if something isn't working, and try again.

poblahblahblah:
It looks like wmi_exporter and node_exporter are more different than I had assumed when I just cargo culted this over, so the tests were fairly incompatible with wmi_exporter. This is not ideal, but for now I've removed the tests.

poblahblahblah:
The MSI that gets built doesn't work - will troubleshoot that.

poblahblahblah:
I guess the good news is that even if I build the installer from master it still doesn't work, so I am guessing this is a local issue. I'll try to find another wandows computer to test on.

poblahblahblah:
@carlpett it looks like `-log.format` is not a valid argument for wmi_exporter.exe, which is why my MSI is failing to install.

Here is what the wmi_exporter.wxs file configures the service to use:

```
"C:\Program Files\wmi_exporter\wmi_exporter.exe" -log.format logger:eventlog?name=wmi_exporter
```

Here's what happens when I try that command line flag against the compiled exe:

```
PS C:\Users\Patrick O'Brien\src\go\src\github.com\martinlindhe\wmi_exporter> ./wmi_exporter.exe -log.format logger:eventlog?name=wmi_exporter
2018/03/21 14:26:06 warning: Couldn't open registry to determine IIS version: The system cannot find the file specified.
flag provided but not defined: -log
```

Any guidance here?

martinlindhe:
@poblahblahblah due to powershell parsing arguments with dots in them, you need to specify them like this:

    "-log.format"

poblahblahblah:
@martinlindhe I guess my point is I didn't change anything with the wxs file so it looks like any time i build an MSI with that wxs file the MSI installer fails. Are you building the official packages with a different wxs file? I am pretty confused

VR6Pete:
This is an interesting feature!

Do you have any examples of where this could be used in a Windows world? Cheers Pete

carlpett:
@poblahblahblah So sorry for missing this! Due to #169, current master doesn't build working msi:s. I've been pretty busy the last few days, but I'll see if I can fix it today. 
If you just want to test your msi, you could make a temporary change which removes the logging flags in the .wxs
Again, apologies for wasting your time on this!

poblahblahblah:
@carlpett no worries!

I've updated the wxs file to work correctly. Everything works and my change looks good.

poblahblahblah:
@carlpett Any chance this can get a re-review?

carlpett:
@poblahblahblah Yes, of course! 
I did a fix for the log-flags in #112, which will also affect this PR a bit (changing the flag declaration syntax), so I would like to merge that first. Sounds ok?

collector/textfile.go
```diff
@@ -0,0 +1,253 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+	"github.com/prometheus/common/log"
+)
+
+var (
+	textFileDirectory = flag.String("collector.textfile.directory", "C:\\Program Files\\wmi_exporter\\textfile_inputs", "Directory to read text files with metrics from.")
+	textFileAddOnce   sync.Once
```
carlpett: This doesn't seem to be used?

poblahblahblah:
@carlpett ok - I think this is ready to go. Tested locally and everything worked like expected.

carlpett:
Great! 
Regarding the variable - yes, that is the idea, but the piece of code which only adds it once doesn't seem to have come along with the translation? Also, I don't think we have the feature which made node_exporter need that part anyway?

poblahblahblah:
Ah - you are correct. Fix incoming.

carlpett:
LGTM!

carlpett:
Thanks @poblahblahblah!

PR #178, opened 2018-03-25
Fix logrus vendoring (capitalization of import path changed)

carlpett:
The recommended `logrus` import path changed from title-cased to lower-cased a while ago. We fixed the imports, but the actual files in vendor/ remained in the wrong directory. 
That means it works to build on Windows (which doesn't care about the casing in paths), but breaks on Linux.


carlpett:
Since this is a very non-interesting change, and blocks my work on #112, I'll fast-track this and merge it myself.

martinlindhe:
Nice work, thanks for fixing this!

PR #185, opened 2018-04-05
Create textfile_inputs directory during install

### LINKED ISSUE #184: textfile collector logs error because default directory does not exist
The textfile collector sets the default directory to the `textfile_inputs` subdirectory of the installation. The directory is not created, however, so each scrape results in an error logged.
We should ensure the default directory is created.

carlpett:
Fixes #184.

It does not handle the case where the user overrides the textfile directory, but I think we can leave that as the users' responsibility for now.


poblahblahblah:
My bad :/

martinlindhe:
LGTM!

PR #186, opened 2018-04-05
Set service recovery to restart on crash

### LINKED ISSUE #175: Windows Service Recovery
Would it be possible to make a suggestion to set the option to automatically restart the wmi exporter service action upon the installation of the MSI? This should just be as simple as setting a reg key within the MSI registry table. 

Cheers

Pete

carlpett:
Fixes #175.


martinlindhe:
LGTM!

PR #194, opened 2018-04-29
Add process_start_time_seconds for the exporter itself

### LINKED ISSUE #165: Missing process_* metrics
It'd quite useful to be able to monitor generic process metrics of the exporter itself, and I'm used to exporters having at least `process_start_time_seconds` exposed.

Obviously this is done by default on Linux by the code in https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector.go, and that can't work because Windows has no `/proc` filesystem.

I'd rather not have to configure wmi_exporter to gather process metrics on itself, though that is a possible workaround.

Ultimately the fix belongs here - https://github.com/prometheus/client_golang/issues/376 - but I'm logging this issue in this repo in case there's some interest in using some of the WMI exporter code for this.

Thanks!

carlpett:
Fixes #165

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -44,6 +44,16 @@ var (
 		[]string{"collector"},
 		nil,
 	)
+
+	// This can be removed when client_golang exposes this on Windows
+	// (See https://github.com/prometheus/client_golang/issues/376)
+	startTime = float64(time.Now().Unix())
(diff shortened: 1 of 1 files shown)


martinlindhe:
LGTM!

PR #198, opened 2018-05-12
Add status metric for service-collector

### LINKED ISSUE #171: Windows Services - Report against Degraded State
Hi,

How can I query windows services in a "degraded" state?  Nagios is able to do this which I am trying to replace with Prometheus :) Not found any values that would tell me a windows service is running in a degraded status.

Thanks :) 

Pete

carlpett:
Fixes #171.
On my test system this results in 2966 new series, so the warning about not applying a filter on the WQL is more true than ever...

### DIFF
--- a/collector/service.go
+++ b/collector/service.go
@@ -27,6 +27,7 @@ var (
 type serviceCollector struct {
 	State     *prometheus.Desc
 	StartMode *prometheus.Desc
+	Status    *prometheus.Desc
 
 	queryWhereClause string
 }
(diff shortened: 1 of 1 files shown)


VR6Pete:
Great news. What do you mean specifically about WQL filter?

Cheers

Pete.

carlpett:
@VR6Pete If you do not specify the `--collector.service.services-where` flag, we export the metrics for all services on the system. Since there's a lot of metrics (and even more after this PR), this puts quite a lot of pressure on your Prometheus storage. So it is something to be aware of.

VR6Pete:
Understood. 

At the moment I just monito all         services however as I have become to understand my estate I can probably filter this down. 

Does the filter support wildcards ? If so I then great. 

Any chance of cutting a new msi release?  Cheers

Pete.

carlpett:
Yes, wildcards are supported (although WQL uses `%` instead of `*`, see [this article](https://blogs.technet.microsoft.com/heyscriptingguy/2012/07/13/use-the-like-operator-to-simplify-your-wql-queries/) for some tips and tricks on WQL)
And yes, might be time for a new release. Look for v0.3.3 in a couple of minutes.

PR #200, opened 2018-05-15
Strip carriage-returns from textfile input

### LINKED ISSUE #199: textfile collector fails to read file
Hi, WMI exporter fails to export metrics from text file.  Error in event log:

```
Error parsing "C:\\Program Files\\wmi_exporter\\textfile_inputs\\test.prom": text format parsing error in line 1: expected float as value, got "8192\r"
```

file content:
```
node_directory_size_bytes{directory="."} 8192
```
I have tested the same file with the same content on node exporter on linux and it worked. Anyone tried to export metrics from file  ?

carlpett:
This PR removes all carriage return characters (`\r`) from input files in the textfile collector when they are read by the expfmt parser, since it does not support them.

Fixes #199

### DIFF
--- a/collector/textfile.go
+++ b/collector/textfile.go
@@ -17,6 +17,7 @@ package collector
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


martinlindhe:
Nice job! LGTM

PR #209, opened 2018-06-06
Refactor wmi query generator to use wildcard selector

carlpett:
This PR refactors the query generator to always use a wildcard query (`SELECT * FROM ...`) rather than the explicit listing of all properties (`SELECT Foo, Bar, Baz FROM ...`) that the `wmi` library does.
The advantage in this approach is that any properties that are for some reason missing on a given host will no longer cause the `Invalid query` error we currently get. The corresponding metric will be zero, but the rest will be fine.

### DIFF
--- a/collector/ad.go
+++ b/collector/ad.go
@@ -612,7 +612,7 @@ type Win32_PerfRawData_DirectoryServices_DirectoryServices struct {
 
 func (c *ADCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
 	var dst []Win32_PerfRawData_DirectoryServices_DirectoryServices
-	q := wmi.CreateQuery(&dst, "")
+	q := queryAll(&dst)
 	if err := wmi.Query(q, &dst); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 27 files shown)


collector/ad.go
```diff
@@ -612,7 +612,7 @@ type Win32_PerfRawData_DirectoryServices_DirectoryServices struct {
 
 func (c *ADCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
 	var dst []Win32_PerfRawData_DirectoryServices_DirectoryServices
-	q := wmi.CreateQuery(&dst, "")
+	q := queryAll(&dst)
```
martinlindhe: for the sake of consistency, could this line be changed in the same way so they look as in eg collector/cs.go ?

collector/cpu.go
```diff
@@ -117,7 +117,7 @@ type Win32_PerfRawData_Counters_ProcessorInformation struct {
 
 func (c *CPUCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
 	var dst []Win32_PerfRawData_PerfOS_Processor
-	q := wmi.CreateQuery(&dst, "")
+	q := queryAll(&dst)
```
martinlindhe: here too

collector/ad.go
carlpett: It actually looks like that we should, for consistency, change cs.go and the few others who follow that pattern (19 like this vs 4 inlined). But I agree, should be the same everywhere.

martinlindhe: I would do the 1-line version myself but its more edits. Your call :-)

martinlindhe:
LGTM feel free to merge whenever

PR #212, opened 2018-06-09
Fix IIS warning when IIS collector is NOT in use

xphyr:
Hi, whenever you start the collector on a machine without IIS installed you would get a warning like so:
C:\wmi_exporter>wmi_exporter.exe
WARN[0000] Couldn't open registry to determine IIS version:The system cannot find the file specified.  source="iis.go:34"

This error message was causing some confusion since the IIS collector was not enabled.  I moved the IIS version struct into the IISCollector and then put the code that checks for the version inside the "NewIISCollector" function.  This way the code checking for IIS only runs IF you enabled the IIS collector, otherwise it never runs, and you dont get an erroneous error.

I also moved the "defer" up to before the check to ensure that the defer is properly handled.


collector/iis.go
```diff
@@ -30,11 +29,11 @@ type simple_version struct {
 
 func getIISVersion() simple_version {
 	k, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\InetStp\`, registry.QUERY_VALUE)
+	defer k.Close()
```
carlpett: What is the reason for this change? In general, it might not be safe to call `Close()` if an error occurred, which is why you'd do it after the err-check. Was this causing some issue?

carlpett:
Hi @xphyr,
Thanks! This has actually been bugging me a bit too, just haven't found the time to prioritize it :)
One question added, otherwise it looks good!

xphyr:
I just put defer back to its proper place.  I have other code when I immediately put defer right after the open, but yeah what you are saying makes more sense.  I moved defer back to where you had it before.  Glad to help out and thanks for the feedback.

carlpett:
Great, thanks!

PR #226, opened 2018-07-19
fix iis collector panic

szook:
If the `dst_cache` slice is empty, the program panics when trying to access `dst_cache[0]`

added a `if len(dst_cache) > 0 {}` block around the use of that variable to prevent that panic.

closes https://github.com/martinlindhe/wmi_exporter/issues/225


carlpett:
Thanks!

PR #227, opened 2018-07-20
append IIS worker process name to corresponding w3wp processes

szook:
modify the `process` collector so that the IIS worker process name is
appended to the corresponding `w3wp` process.

before:
```
wmi_process_private_bytes{creating_process_id="2068",process="w3wp",process_id="12308"} 7.18204928e+08
```

after:
```
wmi_process_private_bytes{creating_process_id="2068",process="w3wp_our.website.com",process_id="12308"} 7.18204928e+08
```

reason:
We have some IIS servers hosting many .NET applications. When there is
resource contention on one of those servers, it's nice to know which IIS
application pool is the culprit.  Having only the process_id to
differentiate between w3wp processes requires additional work to figure
out which is which.  Also it does not allow for historial trending as
the process_id can change across restarts.

By appending the IIS ApplicationPool name to the process name, it makes it easier to identify the culprit and allows for consistent historical trending of resource usage of each app pool.

### DIFF
--- a/collector/process.go
+++ b/collector/process.go
@@ -177,12 +177,21 @@ type Win32_PerfRawData_PerfProc_Process struct {
 	WorkingSetPrivate       uint64
 }
 
+type WorkerProcess struct {
+	AppPoolName string
+	ProcessId   uint32
+}
+
 func (c *ProcessCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
 	var dst []Win32_PerfRawData_PerfProc_Process
 	q := queryAllWhere(&dst, c.queryWhereClause)
 	if err := wmi.Query(q, &dst); err != nil {
 		return nil, err
 	}
+
+	var dst_wp []WorkerProcess
+	q_wp := queryAll(&dst_wp)
+	wmi.QueryNamespace(q_wp, &dst_wp, "root\\WebAdministration")
 
 	for _, process := range dst {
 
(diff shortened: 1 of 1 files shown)


collector/process.go
```diff
@@ -177,13 +177,24 @@ type Win32_PerfRawData_PerfProc_Process struct {
...
 	var dst []Win32_PerfRawData_PerfProc_Process
 	q := queryAllWhere(&dst, c.queryWhereClause)
 	if err := wmi.Query(q, &dst); err != nil {
 		return nil, err
 	}
 
+	var dst_wp []WorkerProcess
+	q_wp := queryAll(&dst_wp)
+	if err := wmi.QueryNamespace(q_wp, &dst_wp, "root\\WebAdministration"); err != nil {
+		dst_wp = nil
```
carlpett: `dst_wp` is already nil, so you could skip this check entirely

carlpett:
This is a nice idea! We might be able to use this pattern for other cases down the line as well.
Just a minor nitpick, otherwise looks good!

collector/process.go
szook: good point.  Removed the check.

carlpett:
Great work, thanks! :tada:

VR6Pete:
This is a great addition, and as said could be used elsewhere.... perhaps to link processes with windows services and the likes. 

I appreciate that this is not a significant change to perhaps warrant a new release but would appreciate if it can be considered.

PR #230, opened 2018-07-25
Add mssql collector

szook:
add new collector `mssql` to collect select SQL Server metrics.

WMI Classes collected:
* Win32_PerfRawData_{instance}_SQLServerAvailabilityReplica
* Win32_PerfRawData_{instance}_SQLServerBufferManager
* Win32_PerfRawData_{instance}_SQLServerDatabaseReplica
* Win32_PerfRawData_{instance}_SQLServerDatabases
* Win32_PerfRawData_{instance}_SQLServerGeneralStatistics
* Win32_PerfRawData_{instance}_SQLServerLocks
* Win32_PerfRawData_{instance}_SQLServerMemoryManager
* Win32_PerfRawData_{instance}_SQLServerSQLStatistics

That list was chosen because that was what [bosun/scollector](https://github.com/bosun-monitor/bosun/blob/master/cmd/scollector/collectors/sql_windows.go) was collecting and we are in the process of migrating from Bosun to Pometheus and wanted to keep those metrics. 

The collector attempts to get a list of SQL instances from the registry and query metrics for each one.

collector is off by default.


collector/mssql.go
```diff
@@ -0,0 +1,2757 @@
...
+
+	params, err := k.ReadValueNames(0)
+	if err != nil {
+		log.Warn("Can't ReadSubKeyNames %#v", err)
+		return sqlInstances
+	}
+
+	for _, param := range params {
+		if val, _, err := k.GetStringValue(param); err == nil {
+			sqlInstances[param] = val
```
carlpett: What are keys and values here?

collector/mssql.go
```diff
@@ -0,0 +1,2757 @@
...
+		}
+	}
+
+	log.Debugf("Detected MSSql Instnaces: %#v\n", sqlInstances)
```
carlpett: Typo instances

collector/mssql.go
```diff
@@ -0,0 +1,2757 @@
...
+
+	"github.com/StackExchange/wmi"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+	"golang.org/x/sys/windows/registry"
+)
+
+type sqlInstancesType map[string]string
+
+var sqlInstances sqlInstancesType
```
carlpett: Any reason this is not a field on the `MSSQLCollector` struct?

carlpett:
Very nice! I left some minor comments. Apart from those, there's also some general things I'd like to see:
- It seems the WMI class reports in kB, but Prometheus exporters should use base units, so bytes in this case. Could you add some conversion on those fields (ie, multiply by 1024)?
- The `persec` data from WMI is most often not actually per second rates, but rather counters (they are normalized to rates in Perfmon and friends when displayed later). So remove the `_persec` suffix from them.

I just had time to skim the list of metrics (quite a few new ones!), will give that a look-over later when I have a bit more time.

Again, thanks, this will be a great addition?

collector/mssql.go
```diff
@@ -0,0 +1,2757 @@
...
+
+func (c *MSSQLCollector) collectAvailabilityReplica(ch chan<- prometheus.Metric, sqlInstance string) (*prometheus.Desc, error) {
+	var dst []win32PerfRawDataSQLServerAvailabilityReplica
+	class := fmt.Sprintf("Win32_PerfRawData_%s_SQLServerAvailabilityReplica", sqlInstance)
+	q := queryAllForClass(&dst, class)
+	if err := wmi.Query(q, &dst); err != nil {
+		return nil, err
+	}
+
+	if len(dst) > 0 {
```
szook: classes [AvailabilityReplica, DatabaseReplica, Databases, Locks] return multiple results.  
those results should be iterated over and each metric recorded and tagged.

szook:
@carlpett: made some more changes.  I'm sure there's a few more things I'm missing, but I feel like it's getting close.

collector/mssql.go
```diff
@@ -246,1021 +244,1026 @@ type MSSQLCollector struct {
...
-			"(Extensioninuseaspercentage)",
+			prometheus.BuildFQName(Namespace, subsystem, "bufman_extension_in_use_as_percentage"),
+			"(BufferManager.Extensioninuseaspercentage)",
 			[]string{"instance"},
 			nil,
 		),
 		ExtensionoutstandingIOcounter: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "extensionoutstanding_i_ocounter"),
-			"(ExtensionoutstandingIOcounter)",
+			prometheus.BuildFQName(Namespace, subsystem, "bufman_extension_outstanding_io_counter"),
```
carlpett: Drop the `_counter`

collector/mssql.go
```diff
@@ -246,1021 +244,1026 @@ type MSSQLCollector struct {
...
+			prometheus.BuildFQName(Namespace, subsystem, "databases_repl_pending_xacts"),
+			"(Databases.ReplPendingXacts)",
+			[]string{"instance", "database"},
 			nil,
 		),
 		ReplTransRate: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "repl_trans_rate"),
-			"(ReplTransRate)",
-			[]string{"instance"},
+			prometheus.BuildFQName(Namespace, subsystem, "databases_repl_trans_rate"),
```
carlpett: What are typical values for this metric? This might also be a confusingly named counter

collector/mssql.go
```diff
@@ -246,1021 +244,1026 @@ type MSSQLCollector struct {
...
-			"(SafeAutoParamsPersec)",
+			prometheus.BuildFQName(Namespace, subsystem, "sqlstats_safe_auto_params"),
+			"(SQLStatistics.SafeAutoParamsPersec)",
 			[]string{"instance"},
 			nil,
 		),
 		SQLAttentionrate: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "sql_attentionrate"),
-			"(SQLAttentionrate)",
+			prometheus.BuildFQName(Namespace, subsystem, "sqlstats_sql_attention_rate"),
```
carlpett: Check this for gauge/counter too

collector/mssql.go
```diff
@@ -246,1021 +244,1026 @@ type MSSQLCollector struct {
...
 
 // NewMSSQLCollector ...
 func NewMSSQLCollector() (Collector, error) {
-	sqlInstances = getMSSQLInstances()
 
 	const subsystem = "mssql"
 	return &MSSQLCollector{
+
 		// Win32_PerfRawData_{instance}_SQLServerAvailabilityReplica
 		BytesReceivedfromReplicaPersec: prometheus.NewDesc(
```
carlpett: Drop the `Persec` from the `Desc` field names as well

collector/mssql.go
```diff
@@ -246,1021 +244,1026 @@ type MSSQLCollector struct {
...
+			prometheus.BuildFQName(Namespace, subsystem, "databases_commit_table_entries"),
+			"(Databases.Committableentries)",
+			[]string{"instance", "database"},
 			nil,
 		),
 		DataFilesSizeKB: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "data_files_size_kb"),
-			"(DataFilesSizeKB)",
-			[]string{"instance"},
+			prometheus.BuildFQName(Namespace, subsystem, "databases_data_files_size"),
```
carlpett: Sorry for not being clearer about this... The convention is to keep the unit in the name, so `_bytes` on all of these

collector/mssql.go
```diff
@@ -246,1021 +244,1026 @@ type MSSQLCollector struct {
...
-			"(ExtensionpagereadsPersec)",
+			prometheus.BuildFQName(Namespace, subsystem, "bufman_extension_page_reads"),
+			"(BufferManager.ExtensionpagereadsPersec)",
 			[]string{"instance"},
 			nil,
 		),
 		Extensionpageunreferencedtime: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "extensionpageunreferencedtime"),
-			"(Extensionpageunreferencedtime)",
+			prometheus.BuildFQName(Namespace, subsystem, "bufman_extension_page_unreferenced_time"),
```
carlpett: Do you know what this measures? Is it absolute time or relative?

collector/mssql.go
szook: Looks like all the `_rate` metrics are analogous to the `persec` ones. 
Changing them to Counters and tweaking names to be less confusing.

collector/mssql.go
szook: done

collector/mssql.go
szook: done

collector/mssql.go
szook: According to [Microsoft](https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-buffer-manager-object?view=sql-server-2017):
`Average seconds a page will stay in the buffer pool extension without references to it.`

So I think a Gauge type is appropriate. 

Changed name to `bufman_extension_page_unreferenced_seconds`

carlpett:
Hey,
Sorry for the slow feedback here, see you've already gone ahead and implemented your idea. I would prefer if we do not create a bunch of collectors, actually. What you could do instead would be to spin of each of the classes as separate go routines in the main collect function. This way, we do not get the sprawl of collectors to enable/disable (I'm fairly sure a request for being able to toggle them all would come pretty quick) and keep the related code together. While you can still easily add duration metrics over the sub-collectors, as well as hypothetically have some collector-specific flag for toggling individual sub-collectors on/off.
What do you think?

carlpett:
For the other set of changes, looks good! This is coming together really nicely :)

collector/mssql.go
```diff
@@ -19,18 +19,53 @@
...
 	"fmt"
+	"regexp"
+	"sync"
+	"time"
 
 	"github.com/StackExchange/wmi"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 	"golang.org/x/sys/windows/registry"
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
```
carlpett: The package name is `kingpin` anyway, no need to alias.

collector/mssql.go
```diff
@@ -19,18 +19,53 @@
...
+		"collector.mssql.class-blacklist",
+		"Regexp of mssql WMI classes to blacklist. Name must both match whitelist and not match blacklist to be included.",
+	).Default("").String()
+
+	mssqlPrintCollectors = kingpin.Flag(
+		"collectors.mssql.class-print",
+		"If true, print available mssql WMI classes",
+	).Bool()
+
+	mssqlScrapeDurationDesc = prometheus.NewDesc(
```
carlpett: Not sure this really works? We'll end up with two identically named metrics, which I'm fairly sure results in an error?
Any reason not to throw it in a dedicated metric `wmi_mssql_subcollector_duration_seconds` (or similar) instead?

collector/mssql.go
```diff
@@ -19,18 +19,53 @@
...
 	"golang.org/x/sys/windows/registry"
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
 )
 
-type sqlInstancesType map[string]string
+var (
+	mssqlCollectorWhitelist = kingpin.Flag(
+		"collector.mssql.class-whitelist",
+		"Regexp of mssql WMI classes to whitelist. Name must both match whitelist and not match blacklist to be included.",
+	).Default(".+").String()
```
carlpett: Would you consider regexp better than comma-separated list here? Since the exact allowed names are known, regexp doesn't seem to add much? (You can even use the `Enums` kingpin type then to get validation early)

collector/mssql.go
```diff
@@ -246,13 +297,26 @@ type MSSQLCollector struct {
...
 
 // NewMSSQLCollector ...
 func NewMSSQLCollector() (Collector, error) {
 
+	mssqlCollectors := getMSSQLCollectors()
+	if *mssqlPrintCollectors {
+		fmt.Printf("Available SQLServer Classes:\n")
+		for name := range mssqlCollectors {
+			fmt.Printf(" - %s\n", name)
+		}
```
carlpett: Maybe it would be reasonable to exit here?

collector/mssql.go
```diff
@@ -1324,79 +1390,83 @@ type win32PerfRawDataSQLServerAvailabilityReplica struct {
 	SendstoTransportPersec         uint64
 }
 
-func (c *MSSQLCollector) collectAvailabilityReplica(ch chan<- prometheus.Metric, sqlInstance string) (*prometheus.Desc, error) {
+func mssqlCollectAvailabilityReplica(c *MSSQLCollector, ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
```
carlpett: Is there a reason not to keep these new collect functions as bound to the MSSQLCollector struct?

collector/mssql.go
szook: it actually does work.  the resulting metrics look like this: 
```
# HELP wmi_exporter_collector_duration_seconds wmi_exporter: Duration of a collection.
# TYPE wmi_exporter_collector_duration_seconds gauge
wmi_exporter_collector_duration_seconds{collector="cpu"} 1.5950872999999999
wmi_exporter_collector_duration_seconds{collector="cs"} 1.5630859
wmi_exporter_collector_duration_seconds{collector="logical_disk"} 0.9350555
wmi_exporter_collector_duration_seconds{collector="mssql"} 2.6281417
wmi_exporter_collector_duration_seconds{collector="mssql_availreplica"} 1.6430905
wmi_exporter_collector_duration_seconds{collector="mssql_bufman"} 1.6100894000000001
wmi_exporter_collector_duration_seconds{collector="mssql_databases"} 1.9571038
wmi_exporter_collector_duration_seconds{collector="mssql_dbreplica"} 2.6281417
wmi_exporter_collector_duration_seconds{collector="mssql_genstats"} 2.4711424
wmi_exporter_collector_duration_seconds{collector="mssql_locks"} 2.5011408
wmi_exporter_collector_duration_seconds{collector="mssql_memmgr"} 1.970111
wmi_exporter_collector_duration_seconds{collector="mssql_sqlstats"} 1.6220949
wmi_exporter_collector_duration_seconds{collector="net"} 1.5520828
wmi_exporter_collector_duration_seconds{collector="os"} 1.5080826
wmi_exporter_collector_duration_seconds{collector="process"} 2.6321453
wmi_exporter_collector_duration_seconds{collector="service"} 2.4591326000000002
wmi_exporter_collector_duration_seconds{collector="system"} 2.5171416
wmi_exporter_collector_duration_seconds{collector="tcp"} 1.5230799
```

collector/mssql.go
szook: That was VSCode trying to be helpful.  I'll remove that.

collector/mssql.go
szook: I couldn't figure out how to do that AND have the ability to pass the function reference as a parameter to the `mssqlExecute()` function.

If you can find an example of someone going that, I'd be happy to change this.

collector/mssql.go
szook: all the other white/blacklist options (iis, nic, logicaldisk) are using regex, so I'd figure I'd keep things consistent.   

But I see your point.  

I'll mimic the `collectors.enabled` option instead.

collector/mssql.go
carlpett: Interestnig. But I think that would be best to change. It might either be that the client library is clever and helpful, or it might even be some luck. Either way, I don't think it is a good idea to have multiple "different" metrics that end up with the same name.

collector/mssql.go
carlpett: Hm. Do you mean something like this? https://play.golang.org/p/HcgvRIBLCxe
Maybe I'm missing your point?

szook: thanks for that nudge.  I think I got it.

carlpett:
LGTM! :tada:
Thanks a lot for your efforts!

carlpett:
I'll make a release with this right away :)

VR6Pete:
Great effort everyone. Looking toward to deploying this update when I return from paternity leave. 

@szook - just wondering if you are ahead of the game here and have created some grafana dashboards and Prometheus alerts you could share wrapped around this collector?

Thank you. 

Pete.

szook:
@VR6Pete: not very exciting, but I have one dashboard work-in-progress that's based off the [15 SQL Server Performance Counters to Monitor](https://blogs.sentryone.com/allenwhite/sql-server-performance-counters-to-monitor/) article I found. 

https://gist.github.com/szook/4d56573ff7529cfeaf7c0b67f9b85902

PR #231, opened 2018-07-25
favicon.ico return 404 not found

szook:
during testing I found that the collectors were getting triggered twice
when hitting the exporter from chome or firefox.

turns out, they were trying to get `/favicon.ico` and the code was
redirecting that request to `/metrics`

added a explicit handler to return 404 NotFound when `/favicon.ico` is
requested.

This issue would not have been triggered with normal requests from prometheus,
only when testing with a browser.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -235,6 +235,7 @@ func main() {
 
 	http.Handle(*metricsPath, promhttp.Handler())
 	http.HandleFunc("/health", healthCheck)
+	http.HandleFunc("/favicon.ico", notFound)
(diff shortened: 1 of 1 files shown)


carlpett:
Yes, this is has bugged me for a while too. I'd actually prefer to make the path explicit instead, (that is, only responding on /metrics), but since that is a breaking change, I've been pushing it forward a bit.
This minor thing is actually also breaking in this regard (someone _could_ have set their prometheus to scrape favicon.ico...), so I'm inclined to not merge this until we decide to do a breaking version, and then do it "right" instead.
Thoughts?

szook:
understandable.  I'll close this for now.

PR #234, opened 2018-08-03
fix mssql metric name type

szook:
found that I had one metric with an extra `s` in `_genstatss_`, making it
inconsistent with it's siblings.


carlpett:
:+1:

PR #235, opened 2018-08-03
reorder functions

szook:
redordered functions so that the two functions that list available child collectors are next
to each other - should make it less likely to miss editing one of those
when adding additional wmi classes

also removed the 'filter' from the `mssqlFilterAvailableClassCollectors`
function as it's not filtering anyting.


carlpett:
Thanks!

PR #236, opened 2018-08-03
add mssql accessmethod class

szook:
in the process of trying to build a grafana dashboard with "useful"
mssql metrics, I was looking around for what metrics might be useful and
came across an article of [15 SQL Server Performace Counters to Monitor](https://blogs.sentryone.com/allenwhite/sql-server-performance-counters-to-monitor/)

two of the suggested metrics are provided by the AccessMethod
class, which I was not yet capturing.

So I added metrics for the
Win32_PerfRawData_MSSQLSERVER_SQLServerAccessMethods class.

PR #237, opened 2018-08-03
fix mssql for non-default sql instances

szook:
the only sql servers I have access to are using the default
`MSSQLSERVER` instance names.  I was contacted by someone using a
different instance name and it was failing.

Found 2 bugs in the code:

1) I was returning the default `MSSQLSERVER` as an instance even if it wasn't found in the registry

2) learned that the WMI class naming scheme for non-default instances
was not what I had coded. Changed to incorproate new knowledge.

### DIFF
--- a/collector/mssql.go
+++ b/collector/mssql.go
@@ -52,21 +52,22 @@ type mssqlInstancesType map[string]string
 func getMSSQLInstances() mssqlInstancesType {
 	sqlInstances := make(mssqlInstancesType)
 
-	// in case querying the registry fails, initialize list to the default instance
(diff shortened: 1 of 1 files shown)


szook:
work in progress until the `AccessMethods` change is merged and I can rebase.

carlpett:
#236 is merged!

carlpett:
LGTM, anything apart from rebase needed?

szook:
I'm also working offline with @sysadmind to see if this, indeed, works.

szook:
got confirmation that this works.  Removed WIP

carlpett:
Thanks! Cutting a new release in a few minutes

PR #238, opened 2018-08-06
fix typo in exporter.go

bbigras:


carlpett:
Ouch, that was bad. Thanks!

PR #241, opened 2018-08-07
protect against emtpy wmi query result sets

szook:
If a WMI query were to return an empty result set, trying to access an
element in the result set would result in a `panic: runtime error:
index out of range` error.

add logic to explicitly check if the result set size is 0 and return an
error rather than try to access an element of the result set.

Fixes https://github.com/martinlindhe/wmi_exporter/issues/240

### DIFF
--- a/collector/ad.go
+++ b/collector/ad.go
@@ -3,6 +3,8 @@
 package collector
 
 import (
+	"errors"
+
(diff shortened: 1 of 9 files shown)


carlpett:
Excellent, thank you!

PR #247, opened 2018-08-20
mssql skip subsequent invalid class

szook:
Created a map struct to keep track of any any instance/class name pairs that generate an "Exception occurred. (Invalid class)" error.

First time an instance/class pair gets that error, it will add to that map.

Before querying WMI data, it will check if the current instance/class is in that map and, if so, skip querying WMI data.

So what will happen is that on the first scrape, the `Exception occurred. (Invalid class)` error will get logged, but subsequent scrapes will not attempt to query that instance/class anymore.  (Until the wmi_exporter service is restarted.)

closes https://github.com/martinlindhe/wmi_exporter/issues/244

### DIFF
--- a/collector/mssql.go
+++ b/collector/mssql.go
@@ -94,6 +94,7 @@ func mssqlBuildWMIInstanceClass(suffix string, instance string) string {
 }
 
 type mssqlCollectorsMap map[string]mssqlCollectorFunc
+type mssqlInvalidClassMap map[string]bool
 
 func mssqlAvailableClassCollectors() string {
 	return "accessmethods,availreplica,bufman,databases,dbreplica,genstats,locks,memmgr,sqlstats"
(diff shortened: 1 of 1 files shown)


szook:
I agree that this is something that should be done via config management.  And I did not think through the startup race-condition ramifications - thanks for bringing that to my attention. 

but I'm also aware that not everyone has a robust config management solution, so what do you think about this compromise: Instead of skipping subsequent metric collection attempts on "invalid class" we just skip subsequent Error logging of that failure?

VR6Pete:
Is there an update on this ? :) still getting extended amounts of time where this collector is failing and timing out.

martinlindhe:
@VR6Pete you can subscribe to the issues you are interested in following, in order to get updates when they are available. Simply spamming tickets asking for updates is not adding anything to the discussion.

carlpett:
This hasn't had any movement in quite a while, and I think we'd want to avoid this approach anyway, as mentioned above.

Issue #252, opened 2018-08-27
Documentation for process wildcard filtering might be incorrect.


kafecho:
Hi, 

I am trying to configure the WMI exporter to collect data from processes which have a certain prefix.  I am doing this on a Windows Server 2016 machine from the command line (i.e. **cmd**, not **Powershell**).

The only way I got it to work is by escaping the % character with another % like so: 

`wmi_exporter.exe --collectors.enabled "cpu,cs,logical_disk,net,os,system,process" --collector.process.processes-where "Name LIKE 'chrome%%'"`

I suspect this is because % is a reserved character on the Windows command line.

If the % is not escaped, the WMI exporter runs a WMI query to only looks for processes with the name `chrome` and returns a single set of results.

carlpett:
Hi @kafecho,
Yes, you are right. If your shell treats `%` as a special symbol, it needs to be escaped.
Seeing as `cmd` might be a quite common choice there, this should probably be documented better. Thanks for noticing!

kafecho:
Sure no problem, I can do a pull request for this if you want.

carlpett:
That would be very appreciated :)

PR #253, opened 2018-08-28
Clarified that the % character has to be escaped

kafecho:
When using the wmi_exporter from the Windows command prompt (cmd) or from a .bat file, the '%' character in a wildcard query has to be escaped with another '%'.


carlpett:
LGTM, thanks a lot!

Fixes #252

Issue #261, opened 2018-09-26
Process collector does not check errors for IIS WorkerProcesses


carlpett:
This line should check for err
https://github.com/martinlindhe/wmi_exporter/blob/cb9da1ae222558c9f3039b782f00a44e17f2e4b3/collector/process.go#L194

We might want to consider adding the errcheck linter so this doesn't slip though again.

carlpett:
This was fixed in #263

Issue #110, opened 2017-08-10
Add sha256sums.txt to releases


martinlindhe:
Like https://github.com/prometheus/prometheus/releases/download/v2.0.0-beta.0/sha256sums.txt

Useful for http://scoop.sh/ or https://chocolatey.org/ (?)

carlpett:
Looking at Appveyor's docs, it is not clear to me that we could achieve that exact layout, with all the file hashes in the same file. As far as I understand, the artifacts from AppVeyor are pushed as they are ready, and there is no synchronization point where we have access to all the files from all the builds in the matrix and can produce that file. @FeodorFitsner, is that correctly understood?

Prometheus `promu` builder does this by building all the releases in the same run, so they do have all the binaries. We could ditch the matrix and just cross-compile the 386 version, at the cost of parallel builds. On the other hand, our builds take ~30s, so maybe that isn't such a big deal.

martinlindhe:
I suspected this was the case :-/ Yea, since our build is pretty small maybe we could scrap parallel builds.

martinlindhe:
Possibly we could make do using goreleaser.
Instead of letting appveyor build releases, then one would build locally.
goreleaser creates and uploads sha256sums.txt, cross-compiles for Windows (and more that is not useful for this project).

https://goreleaser.com/

carlpett:
@martinlindhe Is there a difference between that approach and using `promu` without the build matrix as we discussed previously? 
I'm not sure I understand "_Instead of letting appveyor produce releases, then one would build locally._", are you suggesting that release binaries should be built on someone's local machine and uploaded?

martinlindhe:
> @martinlindhe Is there a difference between that approach and using promu without the build matrix as we discussed previously?

I think promu would be a better approach then. Was just posting about a tool that resolves this for us.

> are you suggesting that release binaries should be built on someone's local machine and uploaded?

That is indeed one usage pattern for goreleaser. They also document CI integration on their homepage.

carlpett:
I don't have a strong opinion of promu vs goreleaser, just wondered if I was missing something :)
Re building locally, I do have a strong opinion, however. I do not think it is a good practice to build public releases on a personal machine. It is no longer (somewhat) simple for an end user to know what is in the binary, and the possibility of human errors (having some local change etc) is much higher.

martinlindhe:
@carlpett just to respond to your comments (i also don't have a strong opionon of this)

[...]

PR #263, opened 2018-10-02
WIP: Add linting

carlpett:
This PR adds gometalinter, which for example would catch #261.
Currently all linters are enabled, which obviously results in *a lot* of errors. We need to determine which ones we think are valuable, from looking at the results of the current output.

WIP until we have a good set of linters, and then fix the resulting errors.


martinlindhe:
Great idea Carl!

The underscore stuff needs to be ignored iirc, because serializing and naming conventions in wmi we cannot do much about

carlpett:
Yep, that is one clear ignore-rule. We also get a bunch of undocumented exported types, since all the WMI classes have a leading capital letter.
Ignoring those two, the `golint` package has two remaining classes of errors, bad capitalization (Cpu vs CPU) and malformed package comments. Those we might actually want to fix?

martinlindhe:
Yea, I think so. Eliminating the warnings now for less noise in future lint runs is a good idea

martinlindhe:
May be only me using Makefiles, but would appericate if this would be added to https://github.com/martinlindhe/wmi_exporter/blob/master/Makefile in this PR to easily trigger it manually ("make lint")

```
lint:
    gometalinter --vendor --config gometalinter.config ./...
```

martinlindhe:
Took a quick look at the lint output. The no-underscores in names could be silenced with this config:

```
{
  "Exclude": [
    "don't use underscores in Go names"
  ]
}
```

carlpett:
Did an initial reconfig now (and moved stuff to the Makefile, good point).
I'll try to get rid of the 74 errors we still have in follow-up commits per linter error.

collector/process.go
```diff
@@ -191,7 +191,10 @@ func (c *ProcessCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Des
 
 	var dst_wp []WorkerProcess
 	q_wp := queryAll(&dst_wp)
-	wmi.QueryNamespace(q_wp, &dst_wp, "root\\WebAdministration")
+	err := wmi.QueryNamespace(q_wp, &dst_wp, "root\\WebAdministration")
+	if err != nil {
+		log.Debugf("Could not query WebAdministration namespace for IIS worker processes: %v. Skipping", err)
```
carlpett: I don't like this, really. We should somehow be able to decide if this is good or bad and appropriately ignore or error on it. At the moment we'll always try this, even for machines without IIS, so we can't say that it is unexpected that it fails. On machines that do have IIS, however, this indicates a problem.

Should probably be fixed in another PR though.

carlpett:
Phew, done. The process even found and fixed a few bugs :)
Suggest reviewing per-commit to get better context.

Also, making this kind of changes makes me want integration tests... Something to think about.

carlpett:
Some of the errcheck fixes do feel a bit ceremony. On the other hand, it might be the most important linter to keep enabled, and "ignore-lint" comments are about as noisy, so...

collector/process.go
martinlindhe: I agree this should be handled more fully, but as you say lets leave it for another PR

collector/textfile.go
```diff
@@ -237,12 +237,16 @@ fileLoop:
 		}
 		var parser expfmt.TextParser
 		parsedFamilies, err := parser.TextToMetricFamilies(carriageReturnFilteringReader{r: file})
-		file.Close()
 		if err != nil {
 			log.Errorf("Error parsing %q: %v", path, err)
 			error = 1.0
 			continue
 		}
+		err = file.Close()
```
martinlindhe: Is this correct? Looks like if err != nil above at line 240, the file wont get closed

collector/ad.go
```diff
@@ -485,8 +485,8 @@ type Win32_PerfRawData_DirectoryServices_DirectoryServices struct {
 	DatabasemodifysPersec                                            uint32
 	DatabaserecyclesPersec                                           uint32
 	DigestBindsPersec                                                uint32
-	DRAHighestUSNCommittedHighpart                                   uint32
-	DRAHighestUSNCommittedLowpart                                    uint32
+	DRAHighestUSNCommittedHighpart                                   uint64
```
martinlindhe: Did gometalinter notice something off with the sizes here?

carlpett:
There are issues with some of the tooling around this for cross-compilation to 32-bit.
We've previously discussed whether the separate build/test run for 32-bit is valuable. It blocks #110, for example. Should we drop the separate run and just build both at the same time?

martinlindhe:
I think we should build all artifacts in one run (golang mutli arch support is great so its easy)

carlpett:
Trying to come back to this. Rebased it now, will try to unify the build as well.

carlpett:
Ok, that took a while. I'll submit that as a separate PR I think...

carlpett:
So, once that is merged, this should give green builds :)

carlpett:
Amazing, it finally clears CI :)

Makefile
```diff
@@ -1,14 +1,17 @@
 export GOOS=windows
 
-fmt:
```
carlpett: Just reordered this so that `build` will be the default action rather than `fmt`

martinlindhe:
LGTM! Please merge when convenient

PR #274, opened 2018-11-13
Detect duplicates across files in textfile collector

### LINKED ISSUE #272: duplicate metric in textfile cause scraping to fail completely
WMI-exporter expose absolutely no metrics when a textfile has duplicate metric. e.g the http://localhost:/metrics has single metric, only a message d'error is displayed (see below).

Error message on the `/metrics` page:

    An error has occurred during metrics gathering:
    collected metric wmi_zzz_test_duplicate_metric_name label:<name:"name" value:"foo" > untyped:<value:13 >  was collected before with the same name and label values

Expected result :

* The `.prom` file with invalid / duplicate entry should be ignored.
* The usual (WMI) metrics should be avaialble for scrape.
* The WMI exporter metric  `wmi_textfile_scrape_error` should equal 1
* The other `.prom` file in the textfile directory should be published

Step to reproduce:
* Create a .prom file with duplicate metrics in WMI-exporter's textfile directory.

Exemple invalid `.prom` file :

    wmi_zzz_test_duplicate_metric_name{name="foo"} 1
    wmi_zzz_test_duplicate_metric_name{name="foo"} 13

carlpett:
Fixes #272.
There are still a couple of cases where the textfile collector can cause the entire scrape to fail, but this should solve for a major contributor of the real-world cases, I hope.

### DIFF
--- a/collector/textfile.go
+++ b/collector/textfile.go
@@ -1,3 +1,5 @@
+// Below code originally copied from prometheus/node_exporter/collector/textfile.go:
+//
 // Copyright 2015 The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -11,11 +13,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build !notextfile
-
 package collector
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -29,7 +30,7 @@ import (
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
 	"github.com/prometheus/common/log"
-	kingpin "gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


collector/textfile.go
```diff
@@ -105,6 +106,14 @@ func convertMetricFamily(metricFamily *dto.MetricFamily, ch chan<- prometheus.Me
 			}
 		}
 
+		h := hash(metricFamily, metric)
+		if seenIn, ok := seen[h]; ok {
+			repr := friendlyString(*metricFamily.Name, names, values)
+			log.Warnf("Metric %s was read from %s, but has already been collected from file %s, skipping", repr, path, seenIn)
+			continue
```
brian-brazil: This is not a good idea, you're now randomly dropping metrics depending on the iteration order. We've had issues with this type of approach in the past.

carlpett: Yeah, it's not great, I agree. However, failing the scrape completely isn't that good either. Do you see some middle road or better approach?
Also, I just realized I've forgotten to set the error flag when this happens, so now it is not alertable. That needs to be fixed.

carlpett: Should we be failing the entire scrape, though? Skipping the textfile stuff completely seems okay (although I'll note that the metrics aren't _randomly_ dropped, iteration order is deterministic here, so which metrics are dropped is consistent across scrapes given consistent input), but any individual breakage from any collector resulting in `up=0` seems a bit harsh?

carlpett: Right, that would be my preference. It will require a bit of gymnastics though, since `promhttp` bails the entire scrape if we allow the data to get there.
I guess I'll do some sort of buffer to be able to detect failure before passing the metrics through the channel. Would this be something that should be done in node_exporter as well?

brian-brazil: That'd be a question for @SuperQ, I've heard no plans in that direction.

carlpett:
Won't be taking this approach, so closing this.

PR #277, opened 2018-11-15
collector/system.go: Add boot_time_seconds metric.

jpds:


carlpett:
Hi @jpds,
Thanks for the contribution! Am I understanding the idea right, that you are just truncating the same data that is exposed in `wmi_system_system_up_time`?
I'm not sure if this is something we can be certain to fix the instability in the underlying reported data? If it fluctuates slightly, it can probably also happen such that it will swap between different integers (being very close to a whole number)

Issue #170, opened 2018-03-14
Specify a custom query to filter process metrics as a MSI installer parameter


godiedelrio:
Hi everybody. As far as I know, the MSI installer does not have a parameter to specify a custom query to filter metrics, as can be done when launching wmi_exporter manually, i.e.

`wmi_exporter.exe -collectors.enabled cs,process -collector.process.processes-where "Name='wmi_exporter' OR Name='svchost'"`

As a workaround, the custom query parameter can be "embedded" in the MSI installer parameter ENABLED_CLLECTORS the following way:

`msiexec /i wmi_exporter-0.2.10-amd64.msi ENABLED_COLLECTORS="cs,process -collector.process.processes-where ""Name='wmi_exporter' OR Name='svchost'"""`

A dedicated MSI installer parameter, like ENABLED_COLLECTORS_FILTERS, should suffice.

kpettijohn:
I believe #291 should handle this request but unless you are building the exporter yourself you will need to wait until a new release is cut.

PR #291, opened 2018-11-27
MSI parameter - EXTRA_FLAGS

kpettijohn:
Added the `ENABLED_COLLECTOR_SERVICE_FILTERS` parameter to the MSI installer in order to allow specifying a custom service query (`collector.service.services-where`).

Example:
```
msiexec /i wmi_exporter-0.5.1-amd64.msi ENABLED_COLLECTORS="service" ENABLED_COLLECTOR_SERVICE_FILTERS="""Name LIKE 'sql%'"""
```

Overall this is a rough attempt (I'm not familiar with MSI/windows packaging) at addressing #170.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,11 +44,17 @@ Name | Description
 `LISTEN_PORT` | The port to bind to. Defaults to 9182.
 `METRICS_PATH` | The path at which to serve metrics. Defaults to `/metrics`
 `TEXTFILE_DIR` | As the `--collector.textfile.directory` flag, provide a directory to read text files with metrics from
+`EXTRA_FLAGS` | Allows passing full CLI flags. Defaults to an empty string.
 
-Parameters are sent to the installer via `msiexec`. Example invocation:
+Parameters are sent to the installer via `msiexec`. Example invocations:
 
 ```powershell
 msiexec /i <path-to-msi-file> ENABLED_COLLECTORS=os,iis LISTEN_PORT=5000
+```
+
+Example service collector with a custom query.
+```powershell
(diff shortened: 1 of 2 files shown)


carlpett:
Thanks @kpettijohn!
I was thinking about if it would make sense to just have a `EXTRA_FLAGS` parameter instead, so we don't need to do this for every flag to the exporter, what do you think?

kpettijohn:
@carlpett sounds good! I believe I have things working with the `EXTRA_FLAGS` parameter now after fighting with how quoting/escaping is being handled.

Example:
```
msiexec /log "C:\pkg_install.log" /i ./wmi_exporter-0.5.0-amd64.msi ENABLED_COLLECTORS="cpu,cs,net,os,service,system" --% EXTRA_FLAGS="--collector.service.services-where ""Name LIKE 'sql%'"""
```

Let me know if you need any changes!

carlpett:
Nice! The `--%` thing looks interesting, what does it mean? :smile: 
It'd be great with a short mention of this in the README file, too, but otherwise looks good!

kpettijohn:
>The stop-parsing symbol (--%) prevents PowerShell from interpreting arguments in program calls as PowerShell commands and expressions.

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_special_characters?view=powershell-6#stop-parsing----

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_parsing?view=powershell-6

Im not very familiar with PowerShell but it seemed required to use `--%` in order to prevent spaces/quotes/symbols in the `EXTRA_FLAGS` parameter from being interpreted incorrectly.

carlpett:
Thanks!

PR #294, opened 2018-11-30
Adding windows build constraints

hairyhenderson:
Obviously this isn't totally necessary, as it only makes sense to build this on Windows, but there's at least one collector (`textfille`) that has no Windows-specific dependencies, so can be built (and tested!) on other OSes.

I found this useful while testing #293, so I didn't have to fire up a Windows VM to test changes 😉

Signed-off-by: Dave Henderson <dhenderson@gmail.com>


hairyhenderson:
PTAL @carlpett @martinlindhe

carlpett:
Looks good :)

PR #305, opened 2019-01-11
Initial push of WMI Exporter Configuration Script.

pete-leese:
This PowerShell script will check for the presence of installed components and dynamically enable the relevant collectors.

Script currently support mssql, IIS and MSMQ however welcome any contributions to extend beyond my use case.

### DIFF
--- a/contrib/configuration_scripts/WMIExporterStateConfiguration.ps1
+++ b/contrib/configuration_scripts/WMIExporterStateConfiguration.ps1
@@ -1,0 +1,98 @@
+﻿cls
+$serviceName = "wmi_exporter"
+$MSMQCollector = "msmq"
+$IISCollector = "iis"
+$MSSQLCollector = "mssql"
+$MetricsPath = "C:\Metrics"
+$ImagePathlocation = "HKLM:\SYSTEM\CurrentControlSet\Services\wmi_exporter\"
+$RegString = "ImagePath"
+$ImagePathstr1 = '"C:\Program Files\wmi_exporter\wmi_exporter.exe" --log.format logger:eventlog?name=wmi_exporter --telemetry.addr :9182 --collector.textfile.directory C:\Metrics --collectors.enabled '
+$global:CoreCollectorsEnabled = "cpu,cs,logical_disk,net,os,process,service,textfile,system" -split ","
+$CollectorServiceQuery = @"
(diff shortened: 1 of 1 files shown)


contrib/configuration_scripts/WMIExporterStateConfiguration.ps1
```diff
@@ -0,0 +1,98 @@
+﻿cls
+$serviceName = "wmi_exporter"
+$MSMQCollector = "msmq"
+$IISCollector = "iis"
+$MSSQLCollector = "mssql"
+$MetricsPath = "C:\Metrics"
+$ImagePathlocation = "HKLM:\SYSTEM\CurrentControlSet\Services\wmi_exporter\"
+$RegString = "ImagePath"
+$ImagePathstr1 = '"C:\Program Files\wmi_exporter\wmi_exporter.exe" --log.format logger:eventlog?name=wmi_exporter --telemetry.addr :9182 --collector.textfile.directory C:\Metrics --collectors.enabled '
+$global:CoreCollectorsEnabled = "cpu,cs,logical_disk,net,os,process,service,textfile,system" -split ","
```
carlpett: You can use the `[defaults]` placeholder value here to not have to keep this list up to date as the defaults may change over time.

carlpett:
Thanks for sharing this! There are a few things that are probably quite specific to your deployment in there, eg looks like you are deploying on AWS, there is something called "Coalmine" which might be an internal application, etc? I think those might best be removed.

Getting this script fully general would of course be quite a bit of work, so let's see how it evolves over time.

VR6Pete:
Yes this really is just a starter for people to contribute and tweak away.

It's only specific in regards to enabled collectors and processes that I want to monitor, and I guess the same would be for others, but really just gives people visibility of what is possible I guess with a little scripting knowledge.

frittentheke:
Just out of curiosity - Is there really a downside of enabling collectors for components which are not installed?
I.e. having MS SQL enabled on machines which do not run MS SQL?

If so, would it make sense for the windows_exporter to be able to adapt automagically according to somewhat of the "list of tentatively enabled collectors" according to the command line option? So treating those as a list of things intended to be exported, but only if the service / component is actually available - otherwise print a startup message?

carlpett:
@frittentheke Yes, if you cannot know if a service is supposed to be enabled or not, it makes it much harder to do alerting. Using your example, if `windows_collector_success{collector="mssql"}` is zero, is that bad or not? Currently that'd be something you might potentially page the DBA team over, but if we'd always expose it, you'd likely end up having to write something that picks out database servers in the selector instead, like `windows_collector_success{collector="mssql",instance=~"db-.+"}` or similar.

fischerman:
Additionally, it fills up the event viewer in no time and rotates other application logs away.

PR #307, opened 2019-01-16
Added an example for textfile collector

giuliov:
Added a practical example for textfile collector using a Powershell snippet that generates a valid file.

### DIFF
--- a/docs/collector.textfile.md
+++ b/docs/collector.textfile.md
@@ -36,3 +36,21 @@ _This collector does not yet have any useful queries added, we would appreciate 
 
 ## Alerting examples
 _This collector does not yet have alerting examples, we would appreciate your help adding them!_
+
+# Example use
+This Powershell script, when run in the `collector.textfile.directory` (default `C:\Program Files\wmi_exporter\textfile_inputs`), generates a valid `.prom` file that should successfully ingested by wmi_exporter.
+
+```Powershell
+$alpha = 42
+$beta = @{ left=3.1415; right=2.718281828; }
+
+Set-Content -Path test1.prom -Encoding Ascii -NoNewline -Value ""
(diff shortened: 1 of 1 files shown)


carlpett:
Thanks, this is great! :tada:
Out of curiosity, what happens without the ascii encoding?

giuliov:
UTF8 encoding uses the BOM by default, so ASCII guarantees that you have no issue reading from Go. Clearly, this breaks if you use non-ASCII chars; in that case there are ways to control the encoding in Powershell but requires more code than fit in such an example. The ASCII is kind of a sign post "hic sunt leones" for PS people.

carlpett:
Suspected as much, but figured I'd ask. We do handle the BOM bits these days, see #293. Have you had issues with it since that was merged?

VR6Pete:
I took a slightly different approach with this but will upload my example script tomorrow.

giuliov:
I could test only on my machine, as my IT deployed 0.4.3. With 0.5 it simplifies to
```Powershell
$alpha = 42
$beta = @{ up=3.1415; down=2.718281828; }

Set-Content -Path test1.prom -Encoding UTF8 -Value ""
Add-Content -Path test1.prom -Encoding UTF8 -Value "# HELP test_alpha_total Some random metric."
Add-Content -Path test1.prom -Encoding UTF8 -Value "# TYPE test_alpha_total counter"
Add-Content -Path test1.prom -Encoding UTF8 -Value "test_alpha_total ${alpha}"
Add-Content -Path test1.prom -Encoding UTF8 -Value "# HELP test_beta_bytes Some other metric."
Add-Content -Path test1.prom -Encoding UTF8 -Value "# TYPE test_beta_bytes gauge"
foreach ($k in $beta.Keys) {
  Add-Content -Path test1.prom -Encoding UTF8 -Value "test_beta_bytes{spin=""${k}""} $( $beta[$k] )"
}
```

carlpett:
Thanks!

PR #308, opened 2019-01-18
Textfile collector docu

maglub:
It took me a while to understand that the text file directory should not have .txt files, but .prom files. This docu change will make this more obvious.


carlpett:
Hey @maglub,
Thanks for improving the docs! I think your PR is trying to merge into the wrong branch though (collector-docs rather than master), could you fix that? Right now I can't even find your change :)

maglub:
Wow, sorry, have no idea what happened here. You can drop this request!

PR #309, opened 2019-01-18
Clarifying .prom as file name suffix

maglub:


carlpett:
Looks good, thanks!

carlpett:
Hm, wait, that didn't look good at all... While the PR didn't include a lot of other files, it still merged to the wrong branch. Sorry that I didn't catch that! Could you try again, and make sure it uses `master` as merge base?

maglub:
I am a bit confused now too. As your last email instructed me to merge with the collector-docs branch.

Could it be that I for some reason read the docu in the v0.5.0 tag.

In any case, I am sorry for any confusion if I am the cause for it. I will wait for further instructions. Since the docu seems clear enough in the master branch, you can throw away my changes.

carlpett:
Sorry, I was probably a bit unclear in the previous PR, you had targeted the collector-docs branch there too, and in addition from an older revision of master.
And yes, what version of the source Google hits is quite unpredictable, I've come to old revisions of stuff quite often.
Anyway, if you think that the current version of the docs is clear enough, I think we don't need to do anything else. Otherwise, I'm very happy you wanted to help with improving the docs, and any other additions to make them clearer are much appreciated! :)

PR #311, opened 2019-01-20
Change AppVeyor build to cross-compile instead of having two runs

### LINKED ISSUE #110: Add sha256sums.txt to releases

carlpett:
This PR stops doing two AppVeyor builds, instead doing one pass and cross-compiling. Also updates to latest images and bringing the appveyor.yml file somewhat closer to best practices.

Added checksumming as well, so fixes #110. 

Initial plan was to use `promu`, but the cross-compile support does not yet work on Windows (it runs all builds inside Docker, and there isn't a Windows-based image).


carlpett:
Oops, accidentally brought in some of the linting stuff from #263.

PR #320, opened 2019-02-16
Container changes

sachinmsft:
PR for populating container metrics.


sachinmsft:
CI is giving this error **collector\container.go:9:2: cannot find package "github.com/Microsoft/hcsshim" in any of:**
can we load github.com/Microsoft/hcsshim pkg to machine?

collector/container.go
```diff
@@ -0,0 +1,258 @@
...
+	ContainerAvailable *prometheus.Desc
+	// memory
+	UsageCommitBytes            *prometheus.Desc
+	UsageCommitPeakBytes        *prometheus.Desc
+	UsagePrivateWorkingSetBytes *prometheus.Desc
+
+	// CPU
+	TotalRuntime100ns  *prometheus.Desc
+	RuntimeUser100ns   *prometheus.Desc
+	RuntimeKernel100ns *prometheus.Desc
```
carlpett: Since these three are scaled to seconds (right?), remove the 100ns suffix to avoid confusion

collector/container.go
```diff
@@ -0,0 +1,258 @@
...
+			nil,
+		),
+		UsagePrivateWorkingSetBytes: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "memory_usage_private_working_set_bytes"),
+			"Memory Usage Private Working Set Bytes",
+			[]string{"container_id"},
+			nil,
+		),
+		TotalRuntime100ns: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "cpu_usage_seconds_total"),
```
carlpett: Does usermode and kernelmode add upp to total, or is there some "other" time too? It might be more in line with how we report other cpu usage to have a single metric with a `mode` label, if possible.

collector/container.go
```diff
@@ -0,0 +1,258 @@
...
+
+	// Types Container is passed to get the containers compute systems only
+	containers, err := hcsshim.GetContainers(hcsshim.ComputeSystemQuery{Types: []string{"Container"}})
+	if err != nil {
+		log.Error("Err in Getting containers:", err)
+		return nil, err
+	}
+
+	if len(containers) == 0 {
+		log.Warn("no containers on node")
```
carlpett: I don't think we need to log a warning here. It might however be nice to add a metric for the number of containers on the host?

collector/container.go
```diff
@@ -0,0 +1,258 @@
...
+		)
+		ch <- prometheus.MustNewConstMetric(
+			c.RuntimeKernel100ns,
+			prometheus.CounterValue,
+			float64(cstats.Processor.RuntimeKernel100ns)*ticksToSecondsScaleFactor,
+			containerId,
+		)
+
+		if len(cstats.Network) == 0 {
+			log.Warn("No Network Stats for container: ", containerId)
```
carlpett: Does this indicate an invalid container configuration? Otherwise, like the warning for no containers above, I'd skip the warning and just possibly add a metric for the number of interfaces on the container.

collector/container.go
```diff
@@ -0,0 +1,258 @@
...
+			continue
+		}
+
+		networkStats := cstats.Network
+
+		for _, interafce := range networkStats {
```
carlpett: Typo, `interafce`->`interface`

collector/container.go
```diff
@@ -0,0 +1,258 @@
+// returns containers metrics from HCS
```
carlpett: This becomes a package-level comment if viewed in godoc, which is confusing. If we want to have this comment at all (I'm not sure it adds much information that isn't readily apparent from skimming the code), merge it with the struct comment for `ContainerMetricsCollector`

carlpett:
>CI is giving this error **collector\container.go:9:2: cannot find package "github.com/Microsoft/hcsshim" in any of:**
>can we load github.com/Microsoft/hcsshim pkg to machine?

We're using vendoring our dependencies with `dep`, so you need to add it by running `dep ensure -add github.com/microsoft/hcsshim`.

collector/container.go
sachinmsft: Yes. There should be network stats in normal scenario so having this log should be okay. I changed it to info instead of warning.

collector/container.go
sachinmsft: Done.

collector/container.go
sachinmsft: I have to confirm with the folks who wrote the hcsshim layer. Though as we have only 2 modes so having separate metrics for them should not be big problem.

collector/container.go
sachinmsft: corrected.

collector/container.go
sachinmsft: corrected.

collector/container.go
sachinmsft: Removed it.

sachinmsft:
> > CI is giving this error **collector\container.go:9:2: cannot find package "github.com/Microsoft/hcsshim" in any of:**
> > can we load github.com/Microsoft/hcsshim pkg to machine?
> 
> We're using vendoring our dependencies with `dep`, so you need to add it by running `dep ensure -add github.com/microsoft/hcsshim`.

I added the dependency in MakeFile, but still CI is not passing. Am I making any mistake. ?

carlpett:
Ah, sorry about being unclear regarding vendoring. You seem to have fixed it since though (not adding dep ensure in Makefile, but checking in the vendor folder), good!
Currently the build seems to fail on some CGO compilation, but only while compiling the test code. This is odd, since we are pruning out test code, so vendoring shouldn't affect test...

Could you verify your version of  `dep`? It should be at least 0.4.1 (but there's no reason not to use 0.5.0).

sachinmsft:
I am using the correct version of dep. Actually Microsoft/hcsshim has dependency on runtime/cgo.
I think during make test it try to run the test and runtime/cgo serach for C compiler. I tried to install min-gw but somehow it is not working properly. 
I was able to run locally by installing TDM-GCC compiler(http://tdm-gcc.tdragon.net/download), but his pkg is not available through choco.
is it possible to configure the appveyor machine manually and install this application?

sachinmsft:
@carlpett, Do you have any suggestions on this. BTW, Is it possible to not build the vendor code on make test.?

carlpett:
Sorry for being slow to reply, been busy with dayjob stuff lately.
As I mentioned, `dep` _should_ be pruning out all the vendor testing code, so I don't see why this doesn't seem to happen here. You could try `dep prune` to run another pass over it, I guess, although it is supposed to be obsolete.

We can do some extra configuration on the machine if necessary, although VS2017 is already installed on the machine, which I would expect to be enough?

sachinmsft:
the compiler was already installed on machine. just path was missing. It worked finally.

sachinmsft:
@carlpett due to https://github.com/martinlindhe/wmi_exporter/issues/321 I assume you are waiting till the end of the week to merge this.
I have made a change to consume these container metrics in the kubernetes-mixin project https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/164 . They do not want to merge that PR until there is a wmi_exporter release that contains these container changes. 
Is there anything I can do to help to speed up the time for a release once this is merged?

collector/container.go
```diff
@@ -0,0 +1,272 @@
...
+	// Types Container is passed to get the containers compute systems only
+	containers, err := hcsshim.GetContainers(hcsshim.ComputeSystemQuery{Types: []string{"Container"}})
+	if err != nil {
+		log.Error("Err in Getting containers:", err)
+		return nil, err
+	}
+
+	count := len(containers)
+
+	if count == 0 {
```
carlpett: We should still be reporting the `ContainersCount` metric here, I think, so move this down below that metric being sent.

collector/container.go
```diff
@@ -0,0 +1,272 @@
...
+
+		container, err := hcsshim.OpenContainer(containerId)
+		if err != nil {
+			log.Error("err in opening container: ", containerId)
+			continue
+		}
+
+		cstats, err := container.Statistics()
+		if err != nil {
+			log.Error("err in fetching container Statistics: ", containerId)
```
carlpett: I think we want the error logged as well

carlpett:
No worries, I'll make a release as soon as this is merged :)

collector/container.go
sachinmsft: added err.

collector/container.go
sachinmsft: done

collector/container.go
```diff
@@ -139,6 +139,14 @@ func (c *ContainerMetricsCollector) Collect(ch chan<- prometheus.Metric) error {
 	return nil
 }
 
+// containerClose closes the container resource
+func containerClose(c hcsshim.Container) {
+	err := c.Close()
+	if err != nil {
+		log.Fatal(err)
```
carlpett: We do not want to quit the exporter if this happens, so I'd say a `log.Error` is sufficient.

sachinmsft: Okay. changed to log.Error.

sachinmsft:
go-winio is added as hcsshim has dependecy on this pkg. I am not sure if we can directly delete test tar files and push the changes to vendor. I can open a issue for this and can take it later if this is not blocker.

sachinmsft:
I deleted testdata tar balls.

carlpett:
Great work! Really happy we'll have this capability in the exporter! :)

Just realized that there is one small thing missing though - docs. So sorry I didn't think about that until now! There is a template [here](https://github.com/martinlindhe/wmi_exporter/blob/master/docs/collector-template.md). The `Classes` bit in the table doesn't make much sense here, so feel free to remove that row.

Apart from that, I think this is ready to merge :tada:

sachinmsft:
@carlpett I have added the doc, though build is failing while installing make
choco install gitversion.portable make -y

sachinmsft:
I have fixed the appveyor.yml and build is green now.

appveyor.yml
```diff
@@ -14,7 +14,8 @@ install:
   - set PATH=%PATH%;C:\mingw-w64\x86_64-7.2.0-posix-seh-rt_v5-rev1\mingw64\bin
   - go get -u github.com/prometheus/promu
   - go get -u github.com/alecthomas/gometalinter && gometalinter --install
-  - choco install gitversion.portable make -y
+  - choco install gitversion.portable --pre -y
```
carlpett: Why is this necessary? I don't think we want prereleases?

sachinmsft:
Build was failing if --pre was not used.
Official page also mentions to use --pre.

https://chocolatey.org/packages/GitVersion.Portable

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

carlpett:
Only to use that particular beta version. The latest stable release _should_ work, I would think: https://chocolatey.org/packages/GitVersion.Portable/4.0.0.
I tested re-running an older version, and there it installed fine. Perhaps you were a bit unlucky and the package was unavailable for a short period of time? Choco isn't always completely reliable, sadly.

carlpett:
I reverted that last commit, and it works again, so let's get this merged! :smile:

carlpett:
Thanks for this contribution! And for staying with me during the rather long review process, sorry it took so long this time.
I'll make a release as well, should be up in a few minutes!

sachinmsft:
Thanks @carlpett for all the help. I see new release with my changes though it does not have .msi and .exe files to download, it just has source code zipped file. does it take some time to populate executable files.?

carlpett:
Interestingly, the build failed on _another_ chocolatey package. Retriggered the build and all went well, binaries are now uploaded!

sachinmsft:
Yes. Thanks a lot.

szediktam:
@sachinmsft  container metrics can have the namespace/pod/container label?

sachinmsft:
> @sachinmsft container metrics can have the namespace/pod/container label?

container metrics are retrieved directly from hcs. which is not aware of pod/container labels. it just have container Id.

PR #327, opened 2019-03-14
msmq - incorrect condition to log a warning

marek-vrana:
A warning is logged to event viewer when a where clause is provided - that should, in fact, lower down the results, so the condition had to be inverted.


carlpett:
Thanks, good catch!

PR #329, opened 2019-03-26
Logical disk new counters

anconina:
Hello,

I have implemented new logical disk counters that represents disk operations latency according the following article : [https://blogs.technet.microsoft.com/askcore/2012/02/07/measuring-disk-latency-with-windows-performance-monitor-perfmon/](url)

> What counters in Windows Performance Monitor show the physical disk latency? 
> “Physical disk performance object -> Avg. Disk sec/Read counter” –  Shows the average read latency. 
> “Physical disk performance object -> Avg. Disk sec/Write counter” –  Shows the average write latency. 
> “Physical disk performance object -> Avg. Disk sec/Transfer counter” – Shows the combined averages for both read and writes. 

Thanks
Moshe

### DIFF
--- a/collector/logical_disk.go
+++ b/collector/logical_disk.go
@@ -29,17 +29,20 @@ var (
 
 // A LogicalDiskCollector is a Prometheus collector for WMI Win32_PerfRawData_PerfDisk_LogicalDisk metrics
 type LogicalDiskCollector struct {
-	RequestsQueued  *prometheus.Desc
-	ReadBytesTotal  *prometheus.Desc
-	ReadsTotal      *prometheus.Desc
-	WriteBytesTotal *prometheus.Desc
-	WritesTotal     *prometheus.Desc
-	ReadTime        *prometheus.Desc
-	WriteTime       *prometheus.Desc
-	TotalSpace      *prometheus.Desc
-	FreeSpace       *prometheus.Desc
-	IdleTime        *prometheus.Desc
-	SplitIOs        *prometheus.Desc
+	RequestsQueued   *prometheus.Desc
+	ReadBytesTotal   *prometheus.Desc
+	ReadsTotal       *prometheus.Desc
+	WriteBytesTotal  *prometheus.Desc
+	WritesTotal      *prometheus.Desc
+	ReadTime         *prometheus.Desc
+	WriteTime        *prometheus.Desc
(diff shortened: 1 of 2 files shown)


collector/logical_disk.go
```diff
@@ -250,6 +277,27 @@ func (c *LogicalDiskCollector) collect(ch chan<- prometheus.Metric) (*prometheus
 			float64(volume.SplitIOPerSec),
 			volume.Name,
 		)
+
+		ch <- prometheus.MustNewConstMetric(
+			c.ReadLatency,
+			prometheus.CounterValue,
+			float64(volume.AvgDiskSecPerRead),
```
carlpett: These values are probably actually the total number of seconds spent on reads. For example, on my system, I have `AvgDisksecPerRead : 1561802799`

anconina: Hi @carlpett ,

Thank you for looking the code.
this value (AvgDisksecPerRead : 1561802799) represent the raw data sample... 
you should calculate the interval between samples, then calculate the difference to get an "average over time period".

Thanks
Moshe

AnSmol:
Latency metrics are vital in terms of disk performance measurement.
I also would love to have these metrics.


@anconina
Thanks for the PR.  To be honest, for me it looks weird too. 
Why the metric called “Average seconds per read(operation)”  gives us huge constantly growing value? There is no such value in perfmon.exe or powershell wmi interface.

collector/logical_disk.go
carlpett: Yes. I think what I meant (and I wasn't very clear, sorry about that) is that we should probably rename the metric to reflect what the value actually is. So `read_latency_seconds_total`, and so on, to match the upstream best practice naming style.

manohrn: can you please get this MSI which has new logical disk perf metrics is added

carlpett: Please do not nag like this. I have updated your PR with a comment on what you need to do to get it merged.

breed808:
Disk latency metrics would be very desirable.

@carlpett I'm happy to copy this PR and correct the metric names, is that OK?

manohrn:
Yes. Pls do that.

carlpett:
@breed808 Certainly, please do!

carlpett:
Replaced by #387

PR #334, opened 2019-05-03
Cpu based on os version

wizardmatas:
Hello, some patch for CPU metrics.

Current exporter doesn't work well in multi socket system, and not provide all cpu metrics.
I have implemented some additional logic to provide metrics based on OS release.

This also solves bug : https://github.com/martinlindhe/wmi_exporter/issues/141

example metrics:

```
# TYPE wmi_cpu_cstate_seconds_total counter
wmi_cpu_cstate_seconds_total{core="0,0",state="c1"} 4.5225769106444e+06
wmi_cpu_cstate_seconds_total{core="0,0",state="c2"} 0
wmi_cpu_cstate_seconds_total{core="0,0",state="c3"} 0
wmi_cpu_cstate_seconds_total{core="0,1",state="c1"} 3.9345788321461e+06
wmi_cpu_cstate_seconds_total{core="0,1",state="c2"} 0
wmi_cpu_cstate_seconds_total{core="0,1",state="c3"} 0
wmi_cpu_cstate_seconds_total{core="0,2",state="c1"} 1.37964725880006e+07
wmi_cpu_cstate_seconds_total{core="0,2",state="c2"} 0
wmi_cpu_cstate_seconds_total{core="0,2",state="c3"} 0
wmi_cpu_cstate_seconds_total{core="0,3",state="c1"} 1.37038457974563e+07
wmi_cpu_cstate_seconds_total{core="0,3",state="c2"} 0
# HELP wmi_cpu_core_frequence_MHz Core frequency in megahertz
# TYPE wmi_cpu_core_frequence_MHz gauge
wmi_cpu_core_frequence_MHz{core="0,0"} 2195
wmi_cpu_core_frequence_MHz{core="0,1"} 2195
wmi_cpu_core_frequence_MHz{core="0,2"} 2195
wmi_cpu_core_frequence_MHz{core="0,3"} 2195
```

Please review it.

Thanks!


collector/cpu.go
```diff
@@ -85,7 +122,7 @@ type Win32_PerfRawData_PerfOS_Processor struct {
 	PercentUserTime       uint64
 }
 
-/* NOTE: This is an alternative class, but it is not as widely available. Decide which to use
+// NOTE: This is an alternative class, but it is not as widely available. Decide which to use
```
martinlindhe: I think the "Decide which to use" part of comment should be dropped since this patch makes use of it.

martinlindhe:
Generally looks good to me, just looking into why appveyor is failing

martinlindhe:
@carlpett Im at loss as to why appveyor is failing here. It's to do with https://github.com/martinlindhe/wmi_exporter/blob/master/appveyor.yml#L27 but I don't understand how this is causing issues for this PR, maybe you have an idea?

collector/cpu.go
wizardmatas: I will remove it

carlpett:
Looks like the compile fails, but I don't understand why we don't get any error message. It might be that `promu` has broken on Windows? I'll take a look.

collector/cpu.go
```diff
@@ -116,99 +152,174 @@ type Win32_PerfRawData_Counters_ProcessorInformation struct {
...
 
 func (c *CPUCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
-	var dst []Win32_PerfRawData_PerfOS_Processor
-	q := queryAll(&dst)
-	if err := wmi.Query(q, &dst); err != nil {
-		return nil, err
-	}
+	version := getWindowsVersion()
+	// Windows version by number https://docs.microsoft.com/en-us/windows/desktop/sysinfo/operating-system-version
+	if version > 6.05 {
```
carlpett: Add a comment here explaining why 6.05 is the magic version :)

collector/cpu.go
```diff
@@ -51,6 +82,12 @@ func NewCPUCollector() (Collector, error) {
 			[]string{"core"},
 			nil,
 		),
+		ProcessorFrequency: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "core_frequence_MHz"),
```
carlpett: Lower-case the `MHz`

collector/cpu.go
```diff
@@ -3,23 +3,54 @@
...
 	Factories["cpu"] = NewCPUCollector
 }
 
+//A function to get windows version from registry
+
+func getWindowsVersion() float64 {
+	k, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\Windows NT\CurrentVersion`, registry.QUERY_VALUE)
+	if err != nil {
+		log.Warn("Couldn't open registry", err)
+		return currentv_flt
```
carlpett: I don't understand this, looks like this variable isn't defined yet here? How does this work?
Could we just return 0 on any error? Or maybe NaN. The only usage is comparing with the magic constant 6.05, right?

carlpett:
Another question, this PR changes the `core` label from a single digit to `<socket>,<core>`. Does this matter/do we care about this? Is there a reasonable way someone could depend on it being just numbers?

collector/cpu.go
wizardmatas: Hi, since `Get-WmiObject Win32_PerfRawData_PerfOS_Processor` work on Windows before 2008 and earlier ( Windows version 6.0 or less) , and `Get-WmiObject Win32_PerfRawData_Counters_ProcessorInformation` on 2008 R2 +(Windows 6.1 or higher) , the split number is 6.05. I will add comment about it.

carlpett: Yeah, I guessed so too, but I think it makes sense to spell it out :)

collector/cpu.go
carlpett: I tried building the PR locally, and get errors about the undefined variable, so I'm guessing this is also why CI fails @martinlindhe. Really weird that we don't get any build errors in the output, though...

wizardmatas: Yes, you are correct, I will return 0 on error. That should failback to legacy wmi query for cpu metrics. Thanks for catch!

wizardmatas:
> Another question, this PR changes the `core` label from a single digit to `<socket>,<core>`. Does this matter/do we care about this? Is there a reasonable way someone could depend on it being just numbers?

Hm, not sure, but for prometheus it shouldn't be a problem.
Example how it looks after exporter upgrade:
image

carlpett:
Yes, I was thinking about if anyone could be doing something that somehow breaks due to the labels changing like that, but I can't think of anything.
I think it'll be sufficient to just add a sentence about it in the release notes, though.

wizardmatas:
Hello, @carlpett when we can expect new release including this fix ? Thanks.

carlpett:
Hey @wizardmatas,
I'm finishing up #335 which I'd like to get in before release. Hopefully this week.

PR #335, opened 2019-05-06
Use perflib for cpu collector

carlpett:
This is a draft implementation of how we could start switching collectors over to perflib. Main structural change is introducing a "scrape context" which for now just carries the perflib data, but could later be used for cancellations etc. 
This scrape context is populated with a "Global" perflib reading at the start of the scrape, and collectors can then pull from this reading. 
I have implemented a simple tag-based unmarshaller of perflib data, so the collector merely builds a struct with tags matching the counters, calls unmarshal and can then get down to pushing metrics back the channel.

### DIFF
--- a/collector/perflib.go
+++ b/collector/perflib.go
@@ -63,10 +63,14 @@ func unmarshalObject(obj *perflib.PerfObject, vs interface{}) error {
 
 			ctr, found := counters[tag]
 			if !found {
-				log.Debugf("missing counter %q, has %v", tag, counters)
+				log.Debugf("missing counter %q, have %v", tag, counterMapKeys(counters))
+				continue
 			}
 			if !target.Field(i).CanSet() {
-				return fmt.Errorf("tagged field %v cannot be written to", f)
+				return fmt.Errorf("tagged field %v cannot be written to", f.Name)
+			}
+			if fieldType := target.Field(i).Type(); fieldType != reflect.TypeOf((*float64)(nil)).Elem() {
+				return fmt.Errorf("tagged field %v has wrong type %v, must be float64", f.Name, fieldType)
 			}
 
 			switch ctr.Def.CounterType {
@@ -85,4 +89,12 @@ func unmarshalObject(obj *perflib.PerfObject, vs interface{}) error {
 	}
 
 	return nil
+}
+
+func counterMapKeys(m map[string]*perflib.PerfCounter) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	return keys
 }
--- a/collector/perflib.go
+++ b/collector/perflib.go
@@ -1,0 +1,89 @@
+package collector
+
+import (
+	"fmt"
+	"reflect"
+
+	perflibCollector "github.com/leoluk/perflib_exporter/collector"
+	"github.com/leoluk/perflib_exporter/perflib"
+	"github.com/prometheus/common/log"
+)
+
+func getPerflibSnapshot() (map[string]*perflib.PerfObject, error) {
+	objects, err := perflib.QueryPerformanceData("Global")
+	if err != nil {
+		return nil, err
+	}
+
+	indexed := make(map[string]*perflib.PerfObject)
+	for _, obj := range objects {
+		indexed[obj.Name] = obj
+	}
+	return indexed, nil
+}
+
+func unmarshalObject(obj *perflib.PerfObject, vs interface{}) error {
+	if obj == nil {
+		return fmt.Errorf("counter not found")
+	}
+	rv := reflect.ValueOf(vs)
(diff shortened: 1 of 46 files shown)


collector/perflib.go
```diff
@@ -0,0 +1,131 @@
...
+
+	indexed := make(map[string]*perflib.PerfObject)
+	for _, obj := range objects {
+		indexed[obj.Name] = obj
+	}
+	return indexed, nil
+}
+
+const (
+	perfCounterRawcountHex         uint32 = 0
```
martinlindhe: I would write these numbers in hex as there's less symmetry in their decimal representation

martinlindhe:
Looks good, anything left to work out for the cpu collection over perflib?

Also looks like CI chokes on perflib not vendored

carlpett:
Yeah, haven't vendored yet because I'm waiting for a PR to be merged, figured I could get some feedback first :)
If you agree about the design I don't think there is much more work needed.

carlpett:
... Except docs, I just realized as I was about to merge. This adds a bunch new metrics.

collector/perflib.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	indexed := make(map[string]*perflib.PerfObject)
+	for _, obj := range objects {
+		indexed[obj.Name] = obj
+	}
+	return indexed, nil
+}
+
+const (
```
leoluk: Most of these constants are exported by perflib_exporter: https://github.com/leoluk/perflib_exporter/blob/master/collector/mapper.go

(as of https://github.com/leoluk/perflib_exporter/pull/14)

carlpett: Ah, I was only looking at the perflib package.

theimdal:
Hi @carlpett.

Will these perflib metrics be completely new or will they replace the old WMI metrics? I ask this question because I think there will be some work to transition from the old wmi metrics to these new ones, dashboards, alerts and so on. This is a transition we must do, because of loosing a lot of metrics during peek loads.

carlpett:
@theimdal They will be identical to the current metrics, so you should not need to make any changes at all!

poblahblahblah:
This looks great. We've got a fairly large windows install base and have been hitting a lot of WMI issues lately, so for us this is definitely a good direction to take this project.

nbellowe:
I'm excited about this concept! 
I've seen other collectors besides the cpu collector experiencing wmi timeout issues when the system is loaded. Are there plans/desire to read raw perfcounters in other collectors as well?

carlpett:
@nbellowe Yes, I would expect the majority of collectors to be converted once we feel confident in the results. There are a few collectors which don't have any direct raw counter, eg `cs`, so we'll see how we deal with those.

For now, we'd really appreciate it if you have time to test out the prerelease ([here](https://github.com/martinlindhe/wmi_exporter/releases/tag/v0.7.999-preview.2)) on a few systems and see if it makes the situation better for you! Especially the new `--scrape.max-duration` flag to allow the exporter to return data even when WMI has stalled seems relevant to your problems.

petedishman:
@carlpett 
Hi, I've just tried out this build but am actually seeing a big performance drop compared with 0.6.0?
Requests with 0.6.0 are taking around 500ms, whereas 0.7.999-preview2 is taking around 1750-2000ms

I'm running both instances with these arguments:
```
 --collector.textfile.directory="C:\Application\Monitoring\WmiExporter" --telemetry.addr="127.0.0.1:9182" --telemetry.path="/wmi/metrics" --collectors.enabled="cpu,iis,logical_disk,cs,memory,net,os,process,system" --collector.process.processes-where="Name LIKE 'php-cgi%' OR Name LIKE 'w3wp%' OR Name Like 'mysqld'"
```
I wondered if the process filtering could be an issue, but didn't really see any difference after removing that (and the process collector)

Is there any extra information I can provide?

carlpett:
@petedishman Thanks for trying it out! I've opened a new issue (#345) where we can investigate further.

poblahblahblah:
How is this looking overall? Is this a path wmi_exporter will want to continue down?

carlpett:
@poblahblahblah Yes, I believe so. It fixes a lot of issues we've been seeing over the years, so I definitely want to move ahead with it. I've been somewhat swamped the last few weeks/months though, and haven't had time to put the finishing touches in on this. Hope to have a free weekend soon.

carlpett:
Sigh, thought I was ready to merge, but managed to find a data race.

carlpett:
This now passes quite a long period of the Go race detector at least, and is probably easier to reason about than the previous nest of goroutines and channels.

carlpett:
@martinlindhe Please have a look when you have time

leoluk:
🎉 Glad to see perflib being used!

carlpett:
:tada:

Issue #270, opened 2018-11-06
WMI Exporter - up - not reliable


VR6Pete:
Morning guys,

I am using WMI Exporter and leveraging Up to determine that the instance is up, however this does not seem to be reliable under high CPU conditions and WMI Exporter stops responding.

This is becoming quite frustrating for my on-call engineers who are getting woken up when an OpsGenie Alert stating an instance is down, when infact it isn't and its a case that CPU is running hot, but not critical.

Is there a way of making this more robust or being able to mitigate against this?

carlpett:
Hi,
What is your alerting rule? I see two "simple" mitigations:
1. Longer time before firing the alert, say `for: 10m`
2. Combine with other indicators such as cpu not being high when the machine last was up

I'd generally suggest alternative 1, but of course this will prolong time until you react when things are _actually_ down

VR6Pete:
Hi Carl,

We already have set for: 10 and reluctant to set it any higher.

What I have noticed is that if I remove all of the collectors apart from textfile it does report as being up consistently and metrics endpoint loads instantly.

I assume something is timing out whilst collecting metrics. 

Below are the collectors I have enabled;

cpu,cs,logical_disk,net,os,process,service,system,mssql,textfile

This is taking well over 30 seconds to load, but seems to be considerably quicker when "system" and "mssql" collectors are removed.

VR6Pete:
mssql class collector availreplica failed after 164.965436s: Exception occurred. (Invalid class )

I wonder if this has anything to do with it timing out ?

VR6Pete:
It doesn't return any metrics on /metrics when SQL collector is enabled, I just see the error in event viewer.

It has worked historically and has been for some months, It might be that it is still timing out, but not enough to prevent metrics from loading under lower cpu conditions.

kedare:
We are having the same issue here, down set to 10 minutes for alerting on Windows and it is randomly failing on high load.

I think the up check should not even rely on wmi, it should always return that it's up (Because if the host is down anyway the exporter will not respond)

VR6Pete:
I think it is a fundamental flaw with this exporter and it does need a heartbeat of some description to validate the system is up.

I have seen it fail under different conditions, but essentially server was still responding as was able to log on via RDP.

I have considered a couple of workarounds;

1) deploy a new instance of WMI exporter runs on a separate port that only uses the textfile collector and create a suitable metrics file, create an absent based alert to determine if host / up down.
[...]

PR #340, opened 2019-05-15
Abort scrapes after configurable timeout

carlpett:
Depends on #335.
This implements a rough abort after a timeout, so the scrape can return even if WMI has hung for some collectors (eg #89, #270).

I'm not super happy about the implementation, it feels like a mess of interleaved goroutines that are not easy to understand. Tried a few variations of using the normal `context`, but it also had a lot of them. 
Happy for suggestions. Maybe the main `Collect` function should be rewritten.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -410,7 +410,7 @@ func (mh *metricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		var err error
 		timeoutSeconds, err = strconv.ParseFloat(v, 64)
 		if err != nil {
-			log.Warnf("Couldn't parse X-Prometheus-Scrape-Timeout-Seconds: %q. Defaulting timeout to %d", v, defaultTimeout)
+			log.Warnf("Couldn't parse X-Prometheus-Scrape-Timeout-Seconds: %q. Defaulting timeout to %f", v, defaultTimeout)
 		}
 	}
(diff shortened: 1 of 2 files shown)


exporter.go
```diff
@@ -309,3 +396,36 @@ loop:
...
+
+func (mh *metricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	const defaultTimeout = 10.0
+
+	var timeoutSeconds float64
+	if v := r.Header.Get("X-Prometheus-Scrape-Timeout-Seconds"); v != "" {
+		var err error
+		timeoutSeconds, err = strconv.ParseFloat(v, 64)
+		if err != nil {
+			log.Warnf("Couldn't parse X-Prometheus-Scrape-Timeout-Seconds: %q. Defaulting timeout to %d", v, defaultTimeout)
```
martinlindhe: CI says

.\exporter.go:413:4: Warnf format %d has arg defaultTimeout of wrong type float64
FAIL	github.com/martinlindhe/wmi_exporter [build failed]

carlpett: Oops. Wonder why I didn't get that locally... Fixing.

carlpett:
Merging this (but note that it merges into the perflib-cpu branch, not master)

PR #348, opened 2019-06-11
a warning fixed: wrong type

youyunhua:


carlpett:
Hi @youyunhua,
Thanks for having a look! However, the warning you fixed is from a third party library we use and then include in our repo (what is called vendoring). I would suggest you open a PR there instead: https://github.com/prometheus/common

youyunhua:
Hi @carlpett, thanks for your comment.

PR #351, opened 2019-06-17
Changed conversion from microseconds to seconds to match the descript…

sqlkabouter:
Resolves martinlindhe/wmi_exporter#350


carlpett:
Nice catch!

PR #352, opened 2019-06-17
Added metrics to MSSQL collector for base counters

sqlkabouter:
Resolves martinlindhe/wmi_exporter#250

Added metrics for the following base counters:
- wmi_mssql_accessmethods_worktables_from_cache_ratio_base
- wmi_mssql_bufman_buffer_cache_hit_ratio_base
- wmi_mssql_databases_log_cache_hit_ratio_base
- wmi_mssql_locks_average_wait_seconds_base

### DIFF
--- a/collector/mssql.go
+++ b/collector/mssql.go
@@ -135,49 +135,50 @@ type MSSQLCollector struct {
 	mssqlScrapeSuccessDesc  *prometheus.Desc
 
 	// Win32_PerfRawData_{instance}_SQLServerAccessMethods
-	AccessMethodsAUcleanupbatches             *prometheus.Desc
-	AccessMethodsAUcleanups                   *prometheus.Desc
-	AccessMethodsByreferenceLobCreateCount    *prometheus.Desc
-	AccessMethodsByreferenceLobUseCount       *prometheus.Desc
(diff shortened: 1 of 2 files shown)


sqlkabouter:
> Thanks for taking this on!
> Looks good, but I think the naming could probably be improved. _Usually_ the `_Base` metric is the "total number" (total cache lookups, total time spent, etc) and the non-`_Base` is the "positive outcome", (cache hits, and so on). Could we change the names to reflect this?

What kind of names would you suggest? Something like:

- _wmi_mssql_bufman_buffer_cache_hits_
- _wmi_mssql_bufman_buffer_cache_lookups_ or _wmi_mssql_bufman_buffer_cache_total_lookups_

These names would differ from the Microsoft docs so it won't be as easy to find them in the documentation.

carlpett:
Yes, those sound like good suggestions! The point about the differing names is valid, but having misleading/incorrect names isn't great either. We usually stick the name of the originating counter in the help text to make it easier to find if you know that name.

secustor:
@sqlkabouter @carlpett 
We would need this base metrics for a correct calculation of the cache hit ratios. 

I would push this over the finish line if nobody minds.

carlpett:
This PR is superseded by #463. Closing.

Issue #73, opened 2017-05-02
Feature Request: CPU Temperature


jeremyje:
It'd be nice to get metrics from https://msdn.microsoft.com/en-us/library/aa394493(v=vs.85).aspx

carlpett:
Hi!
That would be interesting for sure. However, I don't seem to get any values in those WMI classes. Do your systems report anything there?

sandersaares:
https://github.com/openhardwaremonitor/openhardwaremonitor reports useful hardware stats and appears to do some of it through WMI queries. Perhaps this would be a useful source of inspiration.

carlpett:
@sandersaares I had a quick browse through the code, and as far as I can tell, they don't use WMI for actually getting temperatures etc, but only for probing what hardware is installed? They do register a custom WMI provider as well, which I couldn't see what they actually used for, though. 
Did you see any concrete usage of WMI to get the temperatures that I missed?

sandersaares:
No, I just quickly rifled through the code and saw some WMI but I also did not see too direct of a connection. Guess it is no good then.

CasparChou:
hi @carlpett , actually, we really need this index. I have several windows machine for digital signage. They are installed inside the bus stop smartpole, digital advertising box. They are often overheating and make windows system slowly. I need these cpu temperature index to verify is it overheating or not.

StewartThomson:
The documentation above states that 

> Most of the information that the Win32_TemperatureProbe WMI class provides comes from SMBIOS. Real-time readings for the CurrentReading property cannot be extracted from SMBIOS tables. For this reason, current implementations of WMI do not populate the CurrentReading property. The CurrentReading property's presence is reserved for future use.

I played around [this script](https://gist.github.com/jeffa00/9577816) and found that the temperature returned is unchanging likely because of the issue above.

Unless there's features elsewhere, it looks like WMI doesn't support temperature readings at this time, unfortunately

StewartThomson:
[This is promising](http://stackoverflow.com/questions/21884264/retrieve-intel-pch-temperature-with-powershell/36915900#36915900). Ran some bench marks and the value seems to be updating in real time

carlpett:
With the merge of #359, we can close this. Thanks Stewart!

PR #359, opened 2019-07-12
Added thermal zone information and documentation

StewartThomson:
For issue #73 This PR adds a go file for ThermalZoneInformation that was autogenerated by New-Collector.ps1

### DIFF
--- a/collector/thermalzone.go
+++ b/collector/thermalzone.go
@@ -1,0 +1,105 @@
+package collector
+
+import (
+	"github.com/StackExchange/wmi"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+)
+
+func init() {
+	Factories["thermalzone"] = NewThermalZoneCollector
+}
+
+// A ThermalZoneCollector is a Prometheus collector for WMI Win32_PerfRawData_Counters_ThermalZoneInformation metrics
+type ThermalZoneCollector struct {
+	HighPrecisionTemperature *prometheus.Desc
+	PercentPassiveLimit      *prometheus.Desc
+	Temperature              *prometheus.Desc
+	ThrottleReasons          *prometheus.Desc
+}
+
+// NewThermalZoneCollector ...
+func NewThermalZoneCollector() (Collector, error) {
+	const subsystem = "thermalzone"
+	return &ThermalZoneCollector{
+		HighPrecisionTemperature: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "high_precision_temperature"),
+			"(HighPrecisionTemperature)",
+			nil,
+			nil,
+		),
+		PercentPassiveLimit: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "percent_passive_limit"),
+			"(PercentPassiveLimit)",
+			nil,
+			nil,
+		),
+		Temperature: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "temperature"),
+			"(Temperature)",
+			nil,
+			nil,
+		),
+		ThrottleReasons: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "throttle_reasons"),
+			"(ThrottleReasons)",
+			nil,
+			nil,
+		),
+	}, nil
+}
+
+// Collect sends the metric values for each metric
+// to the provided prometheus Metric channel.
+func (c *ThermalZoneCollector) Collect(ch chan<- prometheus.Metric) error {
+	if desc, err := c.collect(ch); err != nil {
+		log.Error("failed collecting thermalzone metrics:", desc, err)
+		return err
+	}
+	return nil
+}
+
+// Win32_PerfRawData_Counters_ThermalZoneInformation docs:
+// - <add link to documentation here>
+type Win32_PerfRawData_Counters_ThermalZoneInformation struct {
+	Name string
+
+	HighPrecisionTemperature uint32
+	PercentPassiveLimit      uint32
+	Temperature              uint32
+	ThrottleReasons          uint32
+}
+
+func (c *ThermalZoneCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
+	var dst []Win32_PerfRawData_Counters_ThermalZoneInformation
+	q := queryAll(&dst)
+	if err := wmi.Query(q, &dst); err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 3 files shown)


collector/thermalzone.go
```diff
@@ -0,0 +1,105 @@
...
+	Temperature              *prometheus.Desc
+	ThrottleReasons          *prometheus.Desc
+}
+
+// NewThermalZoneCollector ...
+func NewThermalZoneCollector() (Collector, error) {
+	const subsystem = "thermalzone"
+	return &ThermalZoneCollector{
+		HighPrecisionTemperature: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "high_precision_temperature"),
```
carlpett: We should suffix this with the unit, so `_kelvin` in this case (unless we convert to Celsius?).
On closer inspection, the value looks like it is actually "decikelvin"(?), I have a value of 3010

collector/thermalzone.go
```diff
@@ -0,0 +1,105 @@
...
+			"(HighPrecisionTemperature)",
+			nil,
+			nil,
+		),
+		PercentPassiveLimit: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "percent_passive_limit"),
```
carlpett: Do you know what this one means? It is alway 100 on my machine.

collector/thermalzone.go
```diff
@@ -0,0 +1,105 @@
...
+			"(PercentPassiveLimit)",
+			nil,
+			nil,
+		),
+		Temperature: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "temperature"),
```
carlpett: As with the high-precision one, suffix with unit. Do we really need both?

collector/thermalzone.go
```diff
@@ -0,0 +1,105 @@
...
+			"(Temperature)",
+			nil,
+			nil,
+		),
+		ThrottleReasons: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "throttle_reasons"),
```
carlpett: Have you found any documentation on what values this can have?

collector/thermalzone.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func (c *ThermalZoneCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
+	var dst []Win32_PerfRawData_Counters_ThermalZoneInformation
+	q := queryAll(&dst)
+	if err := wmi.Query(q, &dst); err != nil {
+		return nil, err
+	}
+
+	ch <- prometheus.MustNewConstMetric(
```
carlpett: I get two zones back when testing in Powershell, so I think we should loop over the returned ones here

docs/collector.thermalzone.md
```diff
@@ -0,0 +1,31 @@
+# tcp collector
+
+The thermalzone collector exposes metrics about system temps. Note that temperature is given in Kelvin
+
+|||
+-|-
+Metric name prefix  | `thermalzone`
+Classes             | [`Win32_PerfRawData_Counters_ThermalZoneInformation`](https://msdn.microsoft.com/en-us/library/aa394341(v=vs.85).aspx)
```
carlpett: The link seems off? I get TCP/IP docs.

carlpett:
Question though - does it seem to give proper data on your machine? I get seemingly constant temperatures of 301 Kelvin, even after running stress tests for a good while.

collector/thermalzone.go
StewartThomson: I'd be more than happy to convert to Celsius, fits our use case better anyway. Wasn't sure if y'all modified the generated code, or if you just wanted a 1 to 1 mapping to the powershell commands.

collector/thermalzone.go
StewartThomson: That's the % limit that the thermal zone is placing on the devices it controls. So 100% would be unconstrained

collector/thermalzone.go
StewartThomson: I think it would be good to keep high precision, but call it Temperature. Not sure why one would want less precision in any case. From what I can tell it's only 1 decimal

collector/thermalzone.go
StewartThomson: per [this documentation](https://wutils.com/wmi/root/cimv2/win32_perfrawdata_counters_thermalzoneinformation/#temperature_properties), it holds a uint32. [This page](https://docs.microsoft.com/en-us/windows-hardware/design/device-experiences/examples--requirements-and-diagnostics) shows the codes to their corresponding meaning. I'll add some documentation for that.

collector/thermalzone.go
StewartThomson: Interesting, I only get one. Will add a loop.

collector/thermalzone.go
carlpett: The generated code is just to get a starting point and not have to write so much boilerplate by hand all the time :)

collector/thermalzone.go
carlpett: Ah, ok. Putting that in the help text would be great :+1:

docs/collector.thermalzone.md
StewartThomson: Woops, I took the tcp collector .md as a template for this. Clearly missed some stuff lol. fixed.

collector/thermalzone.go
carlpett: Agree. I suspect it is the usual backwards compatibility stuff - at some point they could get better precision, but then there's a bunch of code out there assuming it it an integer value in whole degrees, etc.

StewartThomson:
I do get appropriate data. It matches up with what is displayed in Novabench, and temps go up appropriately when benchmarking. It may have something to do with the fact that the code isn't looping over all of the zones. Also, I wasn't sure if you guys modified the generated go files or not, but I'll make the changes described above on Monday. Have a good weekend!

carlpett:
Excellent, thanks!

StewartThomson:
Resolved the above comments

docs/collector.thermalzone.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Flags
+
+None
+
+## Metrics
+
+Name | Description | Type | Labels
+-----|-------------|------|-------
+`wmi_thermalzone_percent_passive_limit` | % Passive Limit is the current limit this thermal zone is placing on the devices it controls. A limit of 100% indicates the devices are unconstrained. A limit of 0% indicates the devices are fully constrained. | gauge | None
```
carlpett: Extra credit for great documentation :+1:

PR #360, opened 2019-07-13
adding windows service example alerts/queries

chrisbloemker:
Just contributing some useful examples that I have made using for monitoring Microsoft SQL Server instances.

### DIFF
--- a/docs/collector.service.md
+++ b/docs/collector.service.md
@@ -86,22 +86,22 @@ groups:
 
   # Sends an alert when the 'sqlserveragent' service is not in the running state for 3 minutes. 
   - alert: SQL Server Agent DOWN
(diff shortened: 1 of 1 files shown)


carlpett:
Thanks a lot @chrisbloemker!
Just out of curiousity, you seem to have a non-standard labeling scheme with regards to setting the `name` label everywhere?

chrisbloemker:
Ahh, yes.. the way we have Prometheus setup for our ec2 scraping is the instance id is relabeled as `instance`, leaving the tag "Name" to be relabeled as `name` I can change the PR to reflect the examples to be more standardized if you'd like.

carlpett:
Yes, I think that would be good, probably less confusing that way. Thanks!

chrisbloemker:
Done :)

carlpett:
Thanks!

Issue #357, opened 2019-07-11
Container names or labels


usego:
I have number of Windows servers running different type of containers. Any chance to get container names and/or labels to simplify identifying of containers? At this moment only container_id are in metrics, which is hard to consume in prometheus / graphana.

carlpett:
Hi @usego,
It looks like this should be possible, it appears we have all these fields available: [ContainerProperties spec](https://godoc.org/github.com/microsoft/hcsshim/internal/schema1#ContainerProperties).
I currently don't have any system with Windows containers running, so I don't have a good way to test this, though. If you'd like to contribute a PR, we'd gladly review it!

Drewster727:
Unfortunately, I'm not seeing anything useful in the ContainerProperties spec:
![image](https://user-images.githubusercontent.com/4528753/61586020-ea0c5480-ab2f-11e9-8944-98d67e99a76f.png)

Drewster727:
@carlpett welp, per:
https://github.com/microsoft/hcsshim/issues/652

Looks like hcsshim won't provide what is needed. Is there another common library we can utilize to integrate?
-Drew

carlpett:
Yep, we need to go about it differently for sure. I'm somewhat concerned about how to support this well, given that Windows aims to [support multiple "container platforms"](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/containerd) (that is, not just Docker). CRI would probably be pretty similar to Docker support, but the more Hyper-V-like containers might not have the same concepts to map to.
Perhaps it would be easiest to get right if we do not put the labels on every metric, but rather have a `container_<type>_info` metric which can be used for lookups? Something like `wmi_container_docker_info{container_id="...", name="...", ...}`. This way we wouldn't run into issues if the different types don't have the same set of information. The queries will be a bit more complicated, of course, you'd need to do something like `wmi_container_cpu_usage_seconds_total * on(container_id) wmi_container_docker_info{name="my-container"}` etc.
What do you think?

wyaopeng:
Can I use github.com/docker/docker/client to get container name then save to class instead microsoft/hcsshim Container_name?

carlpett:
It is a fairly heavy package to import from a dependency point of view, so I'd rather not. How about generating the info metric I suggested above using a shell script and the `textfile` collector?

Smuggla:
[...]

PR #364, opened 2019-07-20
Add container collector doc link to readme

Drewster727:


carlpett:
Hey @Drewster727,
Thanks, this looks good to me!

carlpett:
@Drewster727 I just realized that we have the labels listed in the docs - could you please open a new PR to add `container_name` there? Thanks!

Drewster727:
@carlpett absolutely I can do that. However, I'm hesitant to call that "good to go" as the container name is just the container ID, which isn't very helpful IMO. It's not really providing the name from a docker perspective, and I can't determine if there's a way to provide the name of the image the container is running on. Any thoughts?

carlpett:
Huh, I assumed from the docs that this would do that... Premature merge then, perhaps.
I opened [hcsshim/#652](https://github.com/microsoft/hcsshim/issues/652) to see if this is intentional.

simmohall:
Hi guys,
I ran into the same issue of having the containerId being useless and the hcsshim Name property not being the useful container name.
I ended up having to use the library github.com/docker/docker/client to get the actual container name. (didn't create a PR as i'm not sure it's the best way to do it and I've never used GO :))
btw - awesome exporter guys, really making good use of it

carlpett:
@Drewster727 I think we'll need to revert this until we've closed the discussion in #357 :cry:

PR #370, opened 2019-08-01
Add support for SQL Errors counter

callvirtual:


martinlindhe:
SQLErrorsErrors sounds rather redundant. What about naming it simply SQLErrors?

callvirtual:
> SQLErrorsErrors sounds rather redundant. What about naming it simply SQLErrors?

I agree, naming looks redundant. I've tried to keep things in consist with original naming pattern: `[CategoryName]+[Counter name]`. But if it's required I can remove "Errors" suffix. Just to clarify, should I rename counter name too? So it will became `{prefix}_sqlerrors` instead of `{prefix}_sqlerrors_errors`

carlpett:
How about `{prefix}_sql_errors_total`? That'd be in line with the general Prometheus recommendations.

carlpett:
Thanks @callvirtual! Just did a quick check in the docs you linked (https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-sql-errors-object), and there it seems there are multiple errors tracked. Are these all summed up in this collector?

callvirtual:
> How about `{prefix}_sql_errors_total`? That'd be in line with the general Prometheus recommendations.

Agreed, a good suggestion. I have already uploaded the fix. Tell me if something else needs to be fixed.

callvirtual:
> Thanks @callvirtual! Just did a quick check in the docs you linked (https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-sql-errors-object), and there it seems there are multiple errors tracked. Are these all summed up in this collector?

There's how it looks (don't have real db near by it works fine, already tested)
https://i.imgur.com/kITXnNX.png

So the errors count can be split up by `resource` label in different buckets according to docs.

carlpett:
Ah, great. Then this looks good to me!

carlpett:
Thanks!

callvirtual:
I also plan to add support for [Transactions counters](https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-transactions-object) soon as well. Will send PR in a couple days

PR #372, opened 2019-08-04
Add support for MSSQL Transactions counters

callvirtual:
Pull request changes:
+ Add support for [Transactions counters](https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-transactions-object)
+ Fix wrong counter name for errors count (#370). I forgot to push changes, sorry :(

Also I'm not pretty happy with some counter names:
```
transactions_nonsnapshot_version_active_total
transactions_snapshot_active_total
transactions_active_total
transactions_update_snapshot_active_total
```
If follow the original microsoft naming pattern, they should be called
```
transactions_nonsnapshot_version_transactions
transactions_snapshot_transactions
transactions_transactions
transactions_update_snapshot_transactions
```
It's again a bit redundant, so I replaced the `transactions` suffix with `active_total` which is more appropriate in terms of Prometheus naming convention. I decided to add `active` because microsoft docs says about only active transactions but not transactions at all. This detail may not be obvious if we use just `total` prefix. But as I mentioned I'm not really sure if this decision is trully correct. Therefore if there are any suggestions on how names can be enchanced, feel free to share them :)

### DIFF
--- a/collector/mssql.go
+++ b/collector/mssql.go
@@ -1672,84 +1672,72 @@ func NewMSSQLCollector() (Collector, error) {
 			[]string{"instance"},
 			nil,
 		),
-
 		TransactionsLongestTransactionRunningTime: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "transactions_longest_transaction_running_seconds"),
 			"(Transactions.LongestTransactionRunningTime)",
 			[]string{"instance"},
 			nil,
 		),
-
 		TransactionsNonSnapshotVersionTotalActive: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "transactions_nonsnapshot_version_active_total"),
-			"(Transactions.NonSnapshotVersionTotalActive)",
+			"(Transactions.NonSnapshotVersionTransactions)",
 			[]string{"instance"},
 			nil,
 		),
-
 		TransactionsSnapshotTotalActive: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "transactions_snapshot_active_total"),
-			"(Transactions.SnapshotTotalActive)",
+			"(Transactions.SnapshotTransactions)",
 			[]string{"instance"},
 			nil,
 		),
-
 		TransactionsTotalActive: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "transactions_active_total"),
-			"(Transactions.TotalActive)",
+			"(Transactions.Transactions)",
 			[]string{"instance"},
 			nil,
 		),
-
 		TransactionsUpdateConflictRatio: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "transactions_update_conflict_ratio"),
 			"(Transactions.UpdateConflictRatio)",
 			[]string{"instance"},
 			nil,
 		),
-
 		TransactionsUpdateSnapshotTotalActive: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "transactions_update_snapshot_active_total"),
-			"(Transactions.UpdateSnapshotTotalActive)",
+			"(Transactions.UpdateSnapshotTransactions)",
 			[]string{"instance"},
 			nil,
 		),
-
 		TransactionsVersionCleanupRateKBs: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "transactions_version_cleanup_rate_bytes"),
 			"(Transactions.VersionCleanupRateKBs)",
 			[]string{"instance"},
 			nil,
 		),
-
 		TransactionsVersionGenerationRateKBs: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "transactions_version_generation_rate_bytes"),
 			"(Transactions.VersionGenerationRateKBs)",
 			[]string{"instance"},
 			nil,
 		),
-
 		TransactionsVersionStoreSizeKB: prometheus.NewDesc(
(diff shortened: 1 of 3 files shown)


collector/mssql.go
```diff
@@ -3594,10 +3690,138 @@ func (c *MSSQLCollector) collectSQLErrors(ch chan<- prometheus.Metric, sqlInstan
...
+		c.TransactionsUpdateSnapshotTotalActive,
+		prometheus.CounterValue,
+		float64(v.UpdateSnapshotTransactions),
+		sqlInstance,
+	)
+
+	ch <- prometheus.MustNewConstMetric(
+		c.TransactionsVersionCleanupRateKBs,
+		prometheus.GaugeValue,
+		float64(v.VersionCleanuprateKBPers),
```
carlpett: Given the name of the data, this needs to be normalized to bytes, I suppose?

collector/mssql.go
```diff
@@ -3594,10 +3690,138 @@ func (c *MSSQLCollector) collectSQLErrors(ch chan<- prometheus.Metric, sqlInstan
...
+		sqlInstance,
+	)
+
+	ch <- prometheus.MustNewConstMetric(
+		c.TransactionsVersionGenerationRateKBs,
```
carlpett: Same kilobytes->bytes here.

collector/mssql.go
```diff
@@ -3594,10 +3690,138 @@ func (c *MSSQLCollector) collectSQLErrors(ch chan<- prometheus.Metric, sqlInstan
...
+		prometheus.GaugeValue,
+		float64(v.VersionGenerationrateKBPers),
+		sqlInstance,
+	)
+
+	ch <- prometheus.MustNewConstMetric(
+		c.TransactionsVersionStoreSizeKB,
```
carlpett: More kilobytes

collector/mssql.go
```diff
@@ -360,7 +361,22 @@ type MSSQLCollector struct {
 	SQLStatsUnsafeAutoParams        *prometheus.Desc
 
 	// Win32_PerfRawData_{instance}_SQLServerSQLErrors
-	SQLErrorsTotal                  *prometheus.Desc
+	SQLErrorsTotal *prometheus.Desc
+
+	// Win32_PerfRawData_{instance}_SQLServerTransactions
+	TransactionsFreeSpaceInTempDbKB           *prometheus.Desc
```
carlpett: Make sure these matches the name of the actual metric we expose, if it is not the same as the WMI name or we improved the naming.

carlpett:
>It's again a bit redundant, so I replaced the transactions suffix with active_total which is more appropriate in terms of Prometheus naming convention. I decided to add active because microsoft docs says about only active transactions but not transactions at all. This detail may not be obvious if we use just total prefix. But as I mentioned I'm not really sure if this decision is trully correct. 

In general, I find it more important for us to have good metric names that indicate what they mean, vs staying close to Microsoft naming. So having `active_total` for example is a good thing :)

callvirtual:
@carlpett Thanks for sensible comments! They're fixed now

carlpett:
Nice :tada:
Just noticed one metric on the second read-through that I missed - `transactions_update_conflict_ratio`. What are some typical values you can read from this one? I would suspect this is actually not a ratio that we get from WMI, but rather the total count. Could you see if that is the case?
Apart from that, looks good to me!

callvirtual:
> What are some typical values you can read from this one? I would suspect this is actually not a ratio that we get from WMI, but rather the total count. Could you see if that is the case?

If we can rely on [Microsoft docs](https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-transactions-object?view=sql-server-2017) it says about percentage ratio not just count. 

> The percentage of those transactions using the snapshot isolation level that have encountered update conflicts within the last second. [...]

Also I found mention of counter in [SQL Server book](https://books.google.com/books?id=y5hCAwAAQBAJ&pg=PT1069&lpg=PT1069&dq=Update+conflict+ratio&source=bl&ots=OYLpYs5nID&sig=ACfU3U1RUYgy7V9iKcxay4tpyrxdxNKp2w&hl=ru&sa=X&ved=2ahUKEwjB9YqXn-njAhWmwqYKHZRpANQQ6AEwAnoECAUQAQ#v=onepage&q=Update%20conflict%20ratio&f=false) that defines it as follows

> This counter monitors the ratio of update snapshot transactions that have update conflicts. It's the ratio of the number of conflicts compared to the total number of update snapshot transactions.

But true this or not I can only check tomorrow on staging db. On my laptop value of counter is just pointless 0 :(

callvirtual:
Actually I found a way to look counter values on staging db and `Update conflict ratio` equals to `1476` and also there is `Update conflict ratio base` which is equal to `7175995`. So you're right it's not looks like some sort of ratio and more seems to be total_count or something else.

carlpett:
Ah, as I suspected then.

>If we can rely on Microsoft docs it says about percentage ratio not just count.

Thing is, we can't, a lot of the time. Since we use the PerfRaw values, but the docs are about the "cooked" values, ie after they have some calculation performed on them. In this case, dividing by the base value (which tends to be total number of seconds elapsed since counting started, or similar)

Usually what we do is then just expose the nominator as a counter, so you'd rename it as `update_conflicts_total`. Prometheus can then help us calculate rates over different time periods as needed.

callvirtual:
Got the idea, thanks! Honestly, these unobvious things make me frustrated. The documentation talks about one thing when counter values are completely different and have nothing to do with docs 😬. Will fix naming soon

carlpett:
Yes, I know. We should write something up about how to understand what they _really_ mean. But time is a bit limited, so we still haven't managed to do so. We're sorry that it leads to frustrations for you and other contributors!

carlpett:
This looks good now, thanks a lot for your contribution!

PR #378, opened 2019-08-09
Logical disk latency metrics

manohrn:


carlpett:
Hi @manohrn,
Thanks for the contribution! This overlaps with #329, and the same comments that I mentioned there are relevant to your patch as well.

manohrn:
Can you please help me to get the wmi exporter.msi  Which has BELOW METRICS.
AvgDiskSecPerRead
AvgDiskSecPerWrite

carlpett:
There is no MSI with those metrics, since the mentioned PR is not merged. This is because it has some issues, which also exist in your PR. If you could make those fixes to your PR, we can merge it.

The changes needed are:
- The metric names should show that it is total time spent in each operation, so `read_latency_seconds_total`, `write_latency_seconds_total` and `read_write_latency_seconds_total`.
-  Your PR additionally needs to be rebased in order to be mergeable.

carlpett:
Replaced by #387

PR #393, opened 2019-08-28
Set current_bandwidth to gauge type

breed808:
Bandwidth estimate for an interface may decrease or increase


carlpett:
Good catch, thanks!

PR #396, opened 2019-09-04
Added a required detail for the .prom file to work properly

charlesmorin:
After adding the `role.prom` file on Windows on around 15 virtual machines, we discovered that if we omit to insert an empty line feed at the end of the file, Prometheus won't get metrics from the virtual machine. Adding a new line fixes the issue and immediately start gathering metrics for the virtual machine.


carlpett:
Thanks!

PR #399, opened 2019-09-17
Add adfs collector

breed808:
Perflib is used to collect base AD FS performance counters.
A subset of the total performance counters has been added, but more will
likely be added in the future.

Documentation for the AD FS counters is poor. As such, some counters
have been omitted until their nature can be interpreted.


carlpett:
Looks really good. Thanks a lot!

PR #400, opened 2019-09-22
Use Perflib for logical_disk exporter

breed808:
PR modifies the logical_disk exporter to use Perflib in place of WMI. The PR also fixes value format for several latency metrics (ticks -> seconds).

However, perflib is reporting `% Free Space` as the number of free megabytes (identical to `Free Megabytes`), breaking the `TotalSpace` metric.
I've opened an issue in the perflib_exporter repository regarding this: leoluk/perflib_exporter/issues/23


collector/logical_disk.go
```diff
@@ -204,98 +202,98 @@ func (c *LogicalDiskCollector) collect(ch chan<- prometheus.Metric) (*prometheus
...
-			float64(volume.PercentFreeSpace)*1024*1024,
+			volume.PercentFreeSpace_Base*1024*1024,
 			volume.Name,
 		)
 
 		ch <- prometheus.MustNewConstMetric(
 			c.TotalSpace,
 			prometheus.GaugeValue,
-			float64(volume.PercentFreeSpace_Base)*1024*1024,
+			volume.PercentFreeSpace*1024*1024,
```
carlpett: You're swapping base/non-base here, is this intentional?

collector/logical_disk.go
```diff
@@ -171,26 +170,25 @@ func (c *LogicalDiskCollector) Collect(ctx *ScrapeContext, ch chan<- prometheus.
 // - https://msdn.microsoft.com/en-us/library/ms803973.aspx - LogicalDisk object reference
 type Win32_PerfRawData_PerfDisk_LogicalDisk struct {
```
carlpett: Bit of a nitpick, but now that we don't use WMI here, we don't need to have the type name match the WMI class name. We also don't need to export the type, so perhaps change to just `logicalDisk`?

breed808: Yes, better to clean it up now then to leave others wondering at the type name in the future.
Are you happy to leave the documentation link for the type?

carlpett: I think the doc link still looked relevant, but I didn't look super closely, is there something off with it?

collector/logical_disk.go
breed808: Yes, this is deliberate. The change is confusing, but the new perflib `PercentFreeSpace_Base` matches with the WMI `PercentFreeSpace` and vice versa.
Same for the perflib `PercentFreeSpace` and WMI `PercentFreeSpace_Base`.

I had to re-review this section a number of times :confounded: but the output metric value is the same between the `0.8.3` WMI and the proposed perflib.

carlpett: Ok! :+1:

collector/logical_disk.go
breed808: Not at all. My preference would be to leave it in, as it is still largely relevant for the perflib metrics. Just checking.

breed808: Type has been renamed to `logicalDisk`. I've rebased the commits and force pushed them to my branch.

carlpett:
LGTM, thanks!

PR #401, opened 2019-09-23
ADFS: explicitly use first perflib result

breed808:
Perflib ADFS only returns a single data result, so looping over data is
unnecessary


carlpett:
:+1:

PR #402, opened 2019-09-23
Use perflib for memory collector

breed808:
Self-explanatory. Perflib exposes metrics under the same name, so this was very simple to implement and test.


collector/memory.go
```diff
@@ -257,247 +256,246 @@ func NewMemoryCollector() (Collector, error) {
...
 func (c *MemoryCollector) Collect(ctx *ScrapeContext, ch chan<- prometheus.Metric) error {
-	if desc, err := c.collect(ch); err != nil {
+	if desc, err := c.collect(ctx, ch); err != nil {
 		log.Error("failed collecting memory metrics:", desc, err)
 		return err
 	}
 	return nil
 }
 
 type Win32_PerfRawData_PerfOS_Memory struct {
```
carlpett: As with the logical disk collector, would be nice to rename this type

breed808: Type has been renamed

carlpett:
Thank you so much for these refactorings!

PR #403, opened 2019-09-24
Use perflib for net collector

breed808:
As with the memory collector, this is a very straightforward conversion.

### DIFF
--- a/collector/net.go
+++ b/collector/net.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"regexp"
 
-	"github.com/StackExchange/wmi"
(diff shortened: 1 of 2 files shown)


collector/net.go
```diff
@@ -149,26 +148,25 @@ func mangleNetworkName(name string) string {
 // Win32_PerfRawData_Tcpip_NetworkInterface docs:
 // - https://technet.microsoft.com/en-us/security/aa394340(v=vs.80)
 type Win32_PerfRawData_Tcpip_NetworkInterface struct {
```
carlpett: Type name change

breed808: Fixed!

PR #411, opened 2019-09-27
Set tcp_connections_established to gauge type

### LINKED ISSUE #407: error wmi_tcp_connections_established metric type
why wmi_tcp_connections_established is counter type, but actually gauge is

breed808:
Resolves #407

While the ConnectionsEstablished property in the
Win32_PerfRawData_Tcpip_TCP class is listed as a counter, real-world
metric values have been shown to increase *and* decrease.

Documentation for the property states "Number of TCP connections for
which the **current** state is either ESTABLISHED or CLOSE-WAIT" which
would imply the metric is a gauge.

No other class properties have shown this behaviour.


carlpett:
Good find!

PR #412, opened 2019-10-01
Update collector.hyperv.md

Schlump:
Typo in metrics label.
Metrics exposed for Hyper-V collector are actually named "wmi_hyperv" and not "wmi_hyper".

### DIFF
--- a/docs/collector.hyperv.md
+++ b/docs/collector.hyperv.md
@@ -1,4 +1,4 @@
-# hyperv collector
+ # hyperv collector
 
 The hyperv collector exposes metrics about the Hyper-V hypervisor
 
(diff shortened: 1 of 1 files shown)


breed808:
This would cause the documentation metric names to no longer match those in `collector/hyperv.go`.
Renaming all of the `hyperv` collector metrics would be a significant change, maybe a semver version bump would be necessary?

Schlump:
I'm not really into go, but what i meant was that the documentation is missing the "v" from "hyperv"

The metrics are generated correctly: 

`func NewHyperVCollector() (Collector, error) {
	buildSubsystemName := func(component string) string { return "hyperv_" + component }`

Which is why they're exposed like:

`wmi_hyperv_host_cpu_guest_run_time{core="15"} 1.012039581281e+12`

But in the documentation for the metric its written:
wmi_hyper_host_cpu_guest_run_time

PR #413, opened 2019-10-01
Add product name & version to os collector

der-eismann:
I'm pretty sure I've seen an issue or PR with that before, but it would be really cool to have this info as a metric, similar to `wmi_exporter_build_info` since I really don't want another PowerShell script running, just this shiny exporter.
I could also split these in two metrics, but in my opinion they belong together and would just waste space.

### DIFF
--- a/collector/os.go
+++ b/collector/os.go
@@ -17,6 +17,7 @@ func init() {
 
 // A OSCollector is a Prometheus collector for WMI metrics
 type OSCollector struct {
+	OSInformation           *prometheus.Desc
 	PhysicalMemoryFreeBytes *prometheus.Desc
 	PagingFreeBytes         *prometheus.Desc
 	VirtualMemoryFreeBytes  *prometheus.Desc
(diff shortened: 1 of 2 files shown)


carlpett:
Seems reasonable :) Just one question really, is the caption localized?

carlpett:
@der-eismann Any reason to keep this in draft? My only concern above doesn't really warrant it, I think.

der-eismann:
The only reason is that I still haven't setup my Go toolchain on Windows to actually build this, sorry for the long wait. I'm also not sure if it's localized or not :/

carlpett:
No worries! Do you have access to a German (guessing by your Github location field) Windows machine, perhaps?

der-eismann:
Alright, I finally managed to build this 😄 Yes, my machine is German, however the product names are pretty much equal to the english ones.
I got the following output on my machine:
```
# HELP wmi_os_info OperatingSystem.Caption, OperatingSystem.Version
# TYPE wmi_os_info gauge
wmi_os_info{product="Microsoft Windows 10 Education",version="10.0.18363"} 1
```

SriLanka0:
How long will the approval phase take?

carlpett:
Thanks @der-eismann, looks like we don't need to worry about that, then :)

PR #420, opened 2019-10-19
Switch to go modules

carlpett:
This PR changes to use Go modules instead of `dep`. Still vendoring for now, mainly to keep down changes in CI - I don't think it adds much as of Go 1.13 with the caching proxy solution that would protect us from disappearing dependencies.


carlpett:
Meh. Okay, the AppVeyor stuff will need tweaking anyway. Reworking this.

carlpett:
This seems to work now, albeit with some sad workarounds in the AppVeyor setup. 
The versions chosen for dependencies are not the exact same either, but after quite a bit of testing I believe everything works the same from external observations.

@martinlindhe Good to go?

martinlindhe:
Looks great!
Generally, since it's a large change I would suggest to separate the package upgrades (gometalinter to golangci-lint and prometheus/client_golang) from the go mod switch.

But if you feel confident, lets go!

carlpett:
That was the ambition, but sadly it doesn't work :( `gometalinter` doesn't work with modules, and `client_golang` has to be upgraded to the lowest version any of our depedencies use (I have a branch to bump it "for real" later, to v1.0, but saving that)
So, for lack of better options, I think this is going to have to be it. And it seems to work:tm: (my kingdom for some integration tests...)

carlpett:
:tada:

PR #455, opened 2020-01-17
Add Cache collector

wbagdon:
Add cache collector (https://docs.microsoft.com/en-us/previous-versions/aa394267(v=vs.85))

This has been tested on Windows Server 2016

(Also fixed a stray `)` in the new collector powershell script)


wbagdon:
Is there anything I need to do to get this reviewed? I don't see any contribution guidelines

marcusteixeira:
@carlpett can you look at this PR?

carlpett:
So sorry! I must have accidentally marked this as read in my todo somehow. I'll have a look right away

collector/cache.go
```diff
@@ -0,0 +1,453 @@
+package collector
+
+import (
+	"github.com/StackExchange/wmi"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+)
+
+func init() {
+	Factories["cache"] = NewCacheCollector
```
carlpett, line 10: This is my fault for missing to review this for so long, but we've made some refactorings which has changed how the registration is done. Please have a look at the new `registerCollector` function we use elsewhere

collector/cache.go
```diff
@@ -0,0 +1,453 @@
...
+	SyncMDLReadsPersec           *prometheus.Desc
+	SyncPinReadsPersec           *prometheus.Desc
+}
+
+// NewCacheCollector ...
+func NewCacheCollector() (Collector, error) {
+	const subsystem = "cache"
+	return &CacheCollector{
+		AsyncCopyReadsPersec: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "async_copy_reads_persec"),
```
carlpett, line 51: There's a lot of these `_persec` metrics here, which from a look at one of my systems seem to be named differently from their actual values, because their names are from a calculated value that is displayed in Performance Monitor. The values we read are actually counters, with the total number of operations.

collector/cache.go
```diff
@@ -0,0 +1,453 @@
...
+	if desc, err := c.collect(ch); err != nil {
+		log.Error("failed collecting cache metrics:", desc, err)
+		return err
+	}
+	return nil
+}
+
+// Win32_PerfFormattedData_PerfOS_Cache  docs:
+// - https://docs.microsoft.com/en-us/previous-versions/aa394267(v=vs.85)
+type Win32_PerfFormattedData_PerfOS_Cache struct {
```
carlpett, line 239: If we use WMI, we actually want to use the `PerfRawData` classes rather than the formatted ones. 
But it seems we could actually use the new Perflib interface instead of WMI, the counter name is `Cache`. Would you be willing to refactor to use that instead? It is many, many times faster than WMI, and more reliable. You can look at eg the `logical_disk` collector for an example. Unfortunately, there isn't a generator script yet.

wbagdon: I'll take a look at refactoring this to Perflib

breed808:
Closing as this has been implemented in #723

Issue #374, opened 2019-08-06
Error on Exchange Server


pasarn:
Hi,

i have update the exporter to 0.8.
Now we become this error:
```
An error has occurred during metrics gathering:

error collecting metric Desc{fqName: "wmi_exporter_collector_success", help: "wmi_exporter: Whether the collector was successful.", constLabels: {}, variableLabels: [collector]}: failed to prepare scrape: ReqQueryValueEx failed: Die angeforderte Ressource wird bereits verwendet. errno 170
```

Enabled collectors: net, os, service, system, textfile, cpu, cs, logical_disk

carlpett:
Hi @pasarn,
Thanks for reporting! This seems to be an issue with the new Perflib integration. Could you share your Windows version, if you've made any customizations in the install etc?

carlpett:
Also, do you have other servers where you do not see this problem?

pasarn:
This server does't work:
![grafik](https://user-images.githubusercontent.com/2298311/62526527-b58bde80-b839-11e9-94cc-53501e8ea793.png)

An other server with the same OS Build have no problem. A deference is the cpu cores. The working server have 4 and the other 8.

carlpett:
It seems this problem would be caused by something else locking the performance counters registry key. Does it happen on every scrape or just sometimes? I haven't seen it locked during a long period, but something else could have a bug causing it not to be released.
Some troubleshooting steps:
1. Does running the Windows Performance Monitor (perfmon) work? How about `typeperf`, for example running `typeperf -qx Processor` in a command line prompt?
2. Check what is holding the lock on HKEY_PERFORMANCE_DATA with [Process Explorer](https://docs.microsoft.com/en-gb/sysinternals/downloads/process-explorer).

pasarn:
It is on every scrape.

Output form typeperf:
```
C:\Windows\system32>typeperf -qx Prozessor
\Prozessor(0)\Prozessorzeit (%)
\Prozessor(1)\Prozessorzeit (%)
\Prozessor(2)\Prozessorzeit (%)
\Prozessor(3)\Prozessorzeit (%)
\Prozessor(4)\Prozessorzeit (%)
\Prozessor(5)\Prozessorzeit (%)
\Prozessor(6)\Prozessorzeit (%)
\Prozessor(7)\Prozessorzeit (%)
\Prozessor(_Total)\Prozessorzeit (%)
\Prozessor(0)\Benutzerzeit (%)
[...]

PR #461, opened 2020-02-09
Only query the perflib objects we need

carlpett:
The code is pretty horrible, just wanted to get something to test with for #374.
We want to have this even if it doesn't solve that issue, though - it seems to cut the perflib snapshot time by ~90%.

### DIFF
--- a/collector/ad.go
+++ b/collector/ad.go
@@ -11,7 +11,7 @@ import (
 )
 
 func init() {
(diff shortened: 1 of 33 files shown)


eikaas:
I can confirm that this seems to resolve the Exchange-issue we are experiencing. Response-times are back at ~500ms

PR #463, opened 2020-02-14
WIP: Implement mssql base counters

### LINKED ISSUE #250: sql collector : strange values for bufman_buffer_cache_hit_ratio
Hi,

While trying to set up a dashboard for a MS SQL, I have come upon  wmi_mssql_bufman_buffer_cache_hit_ratio which sometimes has values over 100.

I'm not very familiar with MSSQL, so I was wondering wether this is expected behaviour or if something is wrong ?

I'm using wmi_exporter 0.4.3 on windows 2016 and sql server express 13.0.4001.0.

secustor:
Resolves #250

Continuation of https://github.com/martinlindhe/wmi_exporter/pull/352 as this contribution seems to be abandoned.

### DIFF
--- a/collector/mssql.go
+++ b/collector/mssql.go
@@ -137,49 +137,50 @@ type MSSQLCollector struct {
 	mssqlScrapeSuccessDesc  *prometheus.Desc
 
 	// Win32_PerfRawData_{instance}_SQLServerAccessMethods
-	AccessMethodsAUcleanupbatches             *prometheus.Desc
-	AccessMethodsAUcleanups                   *prometheus.Desc
-	AccessMethodsByreferenceLobCreateCount    *prometheus.Desc
-	AccessMethodsByreferenceLobUseCount       *prometheus.Desc
-	AccessMethodsCountLobReadahead            *prometheus.Desc
-	AccessMethodsCountPullInRow               *prometheus.Desc
-	AccessMethodsCountPushOffRow              *prometheus.Desc
-	AccessMethodsDeferreddroppedAUs           *prometheus.Desc
-	AccessMethodsDeferredDroppedrowsets       *prometheus.Desc
-	AccessMethodsDroppedrowsetcleanups        *prometheus.Desc
-	AccessMethodsDroppedrowsetsskipped        *prometheus.Desc
-	AccessMethodsExtentDeallocations          *prometheus.Desc
-	AccessMethodsExtentsAllocated             *prometheus.Desc
-	AccessMethodsFailedAUcleanupbatches       *prometheus.Desc
-	AccessMethodsFailedleafpagecookie         *prometheus.Desc
-	AccessMethodsFailedtreepagecookie         *prometheus.Desc
-	AccessMethodsForwardedRecords             *prometheus.Desc
-	AccessMethodsFreeSpacePageFetches         *prometheus.Desc
-	AccessMethodsFreeSpaceScans               *prometheus.Desc
-	AccessMethodsFullScans                    *prometheus.Desc
-	AccessMethodsIndexSearches                *prometheus.Desc
-	AccessMethodsInSysXactwaits               *prometheus.Desc
-	AccessMethodsLobHandleCreateCount         *prometheus.Desc
-	AccessMethodsLobHandleDestroyCount        *prometheus.Desc
-	AccessMethodsLobSSProviderCreateCount     *prometheus.Desc
-	AccessMethodsLobSSProviderDestroyCount    *prometheus.Desc
-	AccessMethodsLobSSProviderTruncationCount *prometheus.Desc
-	AccessMethodsMixedpageallocations         *prometheus.Desc
-	AccessMethodsPagecompressionattempts      *prometheus.Desc
-	AccessMethodsPageDeallocations            *prometheus.Desc
-	AccessMethodsPagesAllocated               *prometheus.Desc
-	AccessMethodsPagescompressed              *prometheus.Desc
-	AccessMethodsPageSplits                   *prometheus.Desc
-	AccessMethodsProbeScans                   *prometheus.Desc
-	AccessMethodsRangeScans                   *prometheus.Desc
-	AccessMethodsScanPointRevalidations       *prometheus.Desc
-	AccessMethodsSkippedGhostedRecords        *prometheus.Desc
-	AccessMethodsTableLockEscalations         *prometheus.Desc
-	AccessMethodsUsedleafpagecookie           *prometheus.Desc
-	AccessMethodsUsedtreepagecookie           *prometheus.Desc
-	AccessMethodsWorkfilesCreated             *prometheus.Desc
-	AccessMethodsWorktablesCreated            *prometheus.Desc
-	AccessMethodsWorktablesFromCacheRatio     *prometheus.Desc
+	AccessMethodsAUcleanupbatches              *prometheus.Desc
+	AccessMethodsAUcleanups                    *prometheus.Desc
+	AccessMethodsByreferenceLobCreateCount     *prometheus.Desc
+	AccessMethodsByreferenceLobUseCount        *prometheus.Desc
+	AccessMethodsCountLobReadahead             *prometheus.Desc
+	AccessMethodsCountPullInRow                *prometheus.Desc
+	AccessMethodsCountPushOffRow               *prometheus.Desc
+	AccessMethodsDeferreddroppedAUs            *prometheus.Desc
+	AccessMethodsDeferredDroppedrowsets        *prometheus.Desc
+	AccessMethodsDroppedrowsetcleanups         *prometheus.Desc
+	AccessMethodsDroppedrowsetsskipped         *prometheus.Desc
+	AccessMethodsExtentDeallocations           *prometheus.Desc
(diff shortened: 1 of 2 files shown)


collector/mssql.go
```diff
@@ -137,49 +137,50 @@ type MSSQLCollector struct {
...
+	AccessMethodsRangeScans                    *prometheus.Desc
+	AccessMethodsScanPointRevalidations        *prometheus.Desc
+	AccessMethodsSkippedGhostedRecords         *prometheus.Desc
+	AccessMethodsTableLockEscalations          *prometheus.Desc
+	AccessMethodsUsedleafpagecookie            *prometheus.Desc
+	AccessMethodsUsedtreepagecookie            *prometheus.Desc
+	AccessMethodsWorkfilesCreated              *prometheus.Desc
+	AccessMethodsWorktablesCreated             *prometheus.Desc
+	AccessMethodsWorktablesFromCacheRatio      *prometheus.Desc
+	AccessMethodsWorktablesFromCacheRatio_Base *prometheus.Desc
```
carlpett: It would be nice to not leak the counter-internal names into our "clean" structs, but rather have the names here reflect what they mean. So something like `AccessMethodsWorktablesFromCacheLookups`?
(This goes for all the new fields added below as well)

collector/mssql.go
```diff
@@ -662,6 +666,12 @@ func NewMSSQLCollector() (Collector, error) {
 			[]string{"instance"},
 			nil,
 		),
+		AccessMethodsWorktablesFromCacheRatio_Base: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "accessmethods_worktables_from_cache_ratio_lookups"),
```
carlpett: Skip the `_ratio` here, since it is not lookups on the ratio

carlpett: Also, the metric `accessmethods_worktables_from_cache_ratio` should probably be `accessmethods_worktables_from_cache_hits`, right?

carlpett: _(The above seems to repeat for all the new changes)_

collector/mssql.go
```diff
@@ -1430,6 +1452,12 @@ func NewMSSQLCollector() (Collector, error) {
 			[]string{"instance", "resource"},
 			nil,
 		),
+		LocksAverageWaitTimems_Base: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "locks_average_wait_seconds_lookups"),
```
carlpett: This isn't clear to me - what does this mean?

docs/collector.mssql.md
```diff
@@ -249,7 +253,16 @@ Name | Description | Type | Labels
 _This collector does not yet have explained examples, we would appreciate your help adding them!_
 
 ## Useful queries
-_This collector does not yet have any useful queries added, we would appreciate your help adding them!_
+
+### Buffer Cache Hit Ratio
+
+When you read the counter in perfmon you will get the the percentage pages found in the buffer cache. This percentage is calculated internally based on the total number of cache hits divided by the total number of cache lookups over the last few thousand page accesses.
+This collector retrieves the two internal values separately. In order to give any meaning to the value you will have to calculate the Buffer Cache Hit Ratio in PromQL.
```
carlpett: This is a bit strongly formulated - the values *do* have meaning separately. Just not the hit ratio, which the current naming implies. So with improved naming in the PR, I think this will be less confusing.
The example query is still good to have here, though!

collector/mssql.go
secustor: The bases have no meaning on itself if you go by the Microsoft documentation.    
Better said MS is not sharing  what they are in their [documentation](https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-buffer-manager-object?view=sql-server-ver15).    
Because of this I would prefer the naming style of @sqlkabouter .

We can't be sure here as we have to go with the documentation of Microsoft.

collector/mssql.go
secustor: As there is no description in the MS [documentation](https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-locks-object?view=sqlallproducts-allversions) I can only guess that this is are the number of lock events.

collector/mssql.go
carlpett: I'm sorry to be a bit contrarian here, but while I appreciate that the documentation is a bit vague and the current naming wrong, I'm not sure I'm keen on merging something which introduces _more_ unclarity. The name here is pretty bad in that regard, in my opinion.
Looking at the docs you linked:
>The ratio is the total number of cache hits divided by the total number of cache lookups over the last few thousand page accesses.

By that definition, it seems pretty clear to me that we get the total lookups in the base, and total hits in the "non-base" counter? Are there other reasonable interpretations I'm missing?

collector/mssql.go
carlpett: Ok, so then we'd have `lock_wait_seconds_total` and `lock_waited_total` or something similar in naming?

collector/mssql.go
secustor: It seems to be the case after testing with MS monitoring tools.    
Our DB team will hopefully confirm with this Snapshot that the results using Prometheus+WMI_Exporter and MS tools report the same stats.

secustor:
@carlpett Can give me feedback what is not working in the pipeline?    
It seems like I have no access to the logs.

carlpett:
@secustor It seems it didn't even try to build... And Github's usability team could do better here I think - I need to hover over the failing check to get the text "AppVeyor was unable to build non-mergeable pull request". So if you resolve the conflict it will build.

carlpett:
I hope you can see the build logs this time? If not, the failing part is linting, specifically that the files are not `gofmt`:ed.

collector/mssql.go
```diff
@@ -1054,9 +1070,15 @@ func NewMSSQLCollector() (Collector, error) {
...
-			prometheus.BuildFQName(Namespace, subsystem, "databases_log_cache_hit_ratio"),
-			"(Databases.LogCacheHitRatio)",
+		DatabasesLogCacheHits: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "databases_log_cache_hits"),
+			"(Databases.LogCacheHits)",
+			[]string{"instance", "database"},
+			nil,
+		),
+		DatabasesLogCacheLookups: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "databases_log_cache_hit_lookups"),
```
carlpett: Typo?
```suggestion
			prometheus.BuildFQName(Namespace, subsystem, "databases_log_cache_lookups"),
```

collector/mssql.go
```diff
@@ -1862,7 +1890,8 @@ type win32PerfRawDataSQLServerAccessMethods struct {
 	Usedtreepagecookie            uint64
 	WorkfilesCreatedPersec        uint64
 	WorktablesCreatedPersec       uint64
-	WorktablesFromCacheRatio      uint64
+	WorktablesFromCacheHits       uint64
+	WorktablesFromCacheLookups    uint64
```
carlpett: Is this a find/replace mistake? The underlying counters are `WorktablesFromCacheRatio` and `WorktablesFromCacheRatio_Base`, right?

collector/mssql.go
```diff
@@ -2281,7 +2317,8 @@ func (c *MSSQLCollector) collectAvailabilityReplica(ch chan<- prometheus.Metric,
 //   https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-buffer-manager-object
 type win32PerfRawDataSQLServerBufferManager struct {
 	BackgroundwriterpagesPersec   uint64
-	Buffercachehitratio           uint64
+	Buffercachehits               uint64
+	Buffercachelookups            uint64
```
carlpett: Same here, the field names need to match the WMI counter names

collector/mssql.go
```diff
@@ -2703,7 +2747,8 @@ type win32PerfRawDataSQLServerDatabases struct {
 	DBCCLogicalScanBytesPersec       uint64
 	GroupCommitTimePersec            uint64
 	LogBytesFlushedPersec            uint64
-	LogCacheHitRatio                 uint64
+	LogCacheHits                     uint64
+	LogCacheLookups                  uint64
```
carlpett: One more

collector/mssql.go
```diff
@@ -3298,7 +3350,8 @@ func (c *MSSQLCollector) collectGeneralStatistics(ch chan<- prometheus.Metric, s
 // - https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-locks-object
 type win32PerfRawDataSQLServerLocks struct {
 	Name                       string
-	AverageWaitTimems          uint64
+	LockWaitTime               uint64
+	LockCount                  uint64
```
carlpett: And the last one

collector/mssql.go
secustor: You are correct

collector/mssql.go
secustor: Yes this is a typo

carlpett:
Thank you so very much @secustor! :clap:

Issue #454, opened 2020-01-16
How to get Hostname and operating system version from wmi exporter ? 


khadar-sk:
Hi Team, 
How to get hostname and operating system version from wmi exporter ?  By default wmi exporter provides those details or need to enable some collectors. 

Enabled OS, CS collectors but am not able to get those information. 

Can someone suggest me,  if am missing something to configure or need to add  some thing at exporter level etc.,

carlpett:
Duplicate of/overlaps #160, #409 and #430 
As of #413 we will expose the Windows version. For hostname, see discussions in linked isses.

khadar-sk:
Thank You @carlpett  for you response. 
I have seen you code in repo there is no code pointing to this  "Version, CSName , OSArchitecture" in os.go file in  master branch. 

While referring to the issue 413 you pointed,  have seen a commit (0865061) it is available there. 

My Question is when will be these changes available ? When you are planning to release next version ?

I used this code to build executable file.

`// +build windows

package collector

import (
	"errors"
	"time"

	"github.com/StackExchange/wmi"
	"github.com/prometheus/client_golang/prometheus"
	"github.com/prometheus/common/log"
)

func init() {
	Factories["os"] = NewOSCollector
}

// A OSCollector is a Prometheus collector for WMI metrics
type OSCollector struct {
	OSInformation           *prometheus.Desc
	PhysicalMemoryFreeBytes *prometheus.Desc
	PagingFreeBytes         *prometheus.Desc
	VirtualMemoryFreeBytes  *prometheus.Desc
	ProcessesLimit          *prometheus.Desc
	ProcessMemoryLimitBytes *prometheus.Desc
	Processes               *prometheus.Desc
	Users                   *prometheus.Desc
	PagingLimitBytes        *prometheus.Desc
	VirtualMemoryBytes      *prometheus.Desc
	VisibleMemoryBytes      *prometheus.Desc
	Time                    *prometheus.Desc
	Timezone                *prometheus.Desc
}

// NewOSCollector ...
func NewOSCollector() (Collector, error) {
	const subsystem = "os"

	return &OSCollector{
		OSInformation: prometheus.NewDesc(
			prometheus.BuildFQName(Namespace, subsystem, "info"),
			"OperatingSystem.Caption, OperatingSystem.Version, OperatingSystem.CSName, OperatingSystem.OSArchitecture",
			[]string{"product", "version", "hostname", "architecture"},
			nil,
		),
[...]

Issue #160, opened 2018-02-16
hostname and ip is missing


deniskril:
how get hostname and ip from wmi exporter?

carlpett:
This is something you typically let Prometheus target discovery handle. You'll get the `instance` label set automatically.

PR #470, opened 2020-02-25
Add collector for hostname information

### LINKED ISSUE #409: How can I  get an hostname ?
like node_exporter's Metrics name node_uname_info.
for example:
node_uname_info{domainname="(none)",instance="172.18.0.79:9100",job="node-exporter",machine="x86_64",nodename="OPS-Zabbix-Server"}

sll552:
This can be useful for building grafana dashboards with dropdowns for multiple hosts.
Or for managed instances of Prometheus where the user is not able to add labels via config.

It is built in a similar way as node_exporter already does it with its uname metric.

Fixes #409 #160 #454

### DIFF
--- a/collector/cs.go
+++ b/collector/cs.go
@@ -18,6 +18,7 @@ func init() {
 type CSCollector struct {
 	PhysicalMemoryBytes *prometheus.Desc
 	LogicalProcessors   *prometheus.Desc
+	Hostname            *prometheus.Desc
 }
 
 // NewCSCollector ...
@@ -35,6 +36,15 @@ func NewCSCollector() (Collector, error) {
 			prometheus.BuildFQName(Namespace, subsystem, "physical_memory_bytes"),
 			"ComputerSystem.TotalPhysicalMemory",
 			nil,
+			nil,
+		),
+		Hostname: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "hostname"),
+			"Labeled system hostname information as provided by ComputerSystem.DNSHostName and ComputerSystem.Domain",
+			[]string{
+				"hostname",
+				"domain",
+				"fqdn"},
 			nil,
 		),
 	}, nil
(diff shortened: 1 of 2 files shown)


collector/cs.go
```diff
@@ -18,6 +18,9 @@ func init() {
 type CSCollector struct {
 	PhysicalMemoryBytes *prometheus.Desc
 	LogicalProcessors   *prometheus.Desc
+	Hostname            *prometheus.Desc
+	Domain              *prometheus.Desc
+	Fqdn                *prometheus.Desc
```
carlpett: These don't seem to be used?

collector/cs.go
```diff
@@ -79,5 +93,21 @@ func (c *CSCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Desc, er
 		float64(dst[0].TotalPhysicalMemory),
 	)
 
+	var fqdn string
+	if len(dst[0].Domain) > 0 {
+		fqdn = dst[0].DNSHostname + "." + dst[0].Domain
+	} else {
+		fqdn = dst[0].DNSHostname
+	}
```
carlpett: When is the domain empty? From some quick testing, `Domain` is the workgroup name on non-domain joined machines, are there more cases here I'm missing?

collector/cs.go
sll552: Removed, those were left overs of an earlier version with metrics for each value

collector/cs.go
sll552: I have only tried it on domain joined machines, I just assumed that `Domain` would be empty as there is another `Workgroup` field in that class.

I don't know if there are cases in which this could be empty. You are right that this is the Workgroup name on non-domain joined machines. 

Also I don't know if it would be expected behavior that the fqdn on such machines would then be `Hostname.WORKGROUP` in the metric. Maybe I should change this check to be more like this
```go
if dst[0].Domain != dst[0].Workgroup {
        fqdn = dst[0].DNSHostname + "." + dst[0].Domain
} else {
	fqdn = dst[0].DNSHostname
}
```
What do you think?

carlpett: That looks reasonable! Or we could skip the fqdn altogether? It is quite simple to add via relabeling if it is needed, and I'm not sure how many users would have multiple AD domains scraped by the same Prometheus instance?

sll552: I adapted the check.
I wanted to include the fqdn for dashboarding reasons. It is much easier to make a Grafana Dashboard with dropdowns for the hosts (e.g. https://grafana.com/grafana/dashboards/1860) with the availability of such a metric. So i would like to keep it even if it might seem redundant.

carlpett:
LGTM, thanks a lot!

sll552:
Thanks for review and merge, is there a timeframe for the next release?

carlpett:
I'm hoping to land a few of the open PRs, but end of this week at latest

PR #474, opened 2020-03-02
Add option to limit concurrent requests

carlpett:
This PR adds a flag to limit concurrent scrapes, `--max-requests <n>`. Any requests coming in while there are already `n` scrapes in progress get a 503.
I picked a default of 5 somewhat arbitrarily, but most systems start to have problems servicing even two or three concurrent requests with heavier WMI queries. Given our 10 second default timeout, this default would allow for one scrape every two seconds, which is quite a bit more frequent than what is normally recommended.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -265,6 +265,10 @@ func main() {
 			"telemetry.path",
 			"URL path for surfacing collected metrics.",
 		).Default("/metrics").String()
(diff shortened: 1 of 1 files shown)


carlpett:
@martinlindhe Do you have time to take a look at this and #473?

PR #489, opened 2020-03-24
BREAKING: Convert the process collector to use perflib

carlpett:
This PR converts the process collector to use perflib rather than WMI. Without filters, it speeds up collection on my machine by 10-50x.
However, since the current implementation has filtering based on WQL, which we cannot support for perflib, this is a breaking change. I changed the flags to match what we have in eg `logical_disk`, with regexp white/blacklists.

### DIFF
--- a/collector/process.go
+++ b/collector/process.go
@@ -3,6 +3,8 @@
 package collector
 
 import (
+	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


YangLu1031:
Thanks @carlpett! Just curious when do we expect to be released?

carlpett:
@YangLu1031 I think it would be nice to get a couple of the larger changes in at the same time so we don't need so many releases with breaking changes, but: soon.

YangLu1031:
@carlpett I see, thanks! And how to create our own .msi file including this change?

PR #493, opened 2020-03-31
Add dhcp collector

breed808:
Unfortunately perflib does not expose scope statistics
(DhcpServerv4ScopeStatistics), and WMI queries to this class don't
appear to be possible with the stackexchange/wmi module.

### DIFF
--- a/collector/dhcp.go
+++ b/collector/dhcp.go
@@ -1,0 +1,387 @@
+// +build windows
+
+package collector
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	registerCollector("dhcp", NewDhcpCollector, "DHCP Server")
+}
+
+// A DhcpCollector is a Prometheus collector perflib DHCP metrics
+type DhcpCollector struct {
+	PacketsReceivedTotal                             *prometheus.Desc
+	DuplicatesDroppedTotal                           *prometheus.Desc
+	PacketsExpiredTotal                              *prometheus.Desc
+	ActiveQueueLength                                *prometheus.Desc
+	ConflictCheckQueueLength                         *prometheus.Desc
+	DiscoversTotal                                   *prometheus.Desc
+	OffersTotal                                      *prometheus.Desc
+	RequestsTotal                                    *prometheus.Desc
+	InformsTotal                                     *prometheus.Desc
+	AcksTotal                                        *prometheus.Desc
+	NacksTotal                                       *prometheus.Desc
+	DeclinesTotal                                    *prometheus.Desc
+	ReleasesTotal                                    *prometheus.Desc
+	OfferQueueLength                                 *prometheus.Desc
+	DeniedDueToMatch                                 *prometheus.Desc
+	DeniedDueToNonMatch                              *prometheus.Desc
+	FailoverBndupdSentTotal                          *prometheus.Desc
+	FailoverBndupdReceivedTotal                      *prometheus.Desc
+	FailoverBndackSentTotal                          *prometheus.Desc
+	FailoverBndackReceivedTotal                      *prometheus.Desc
+	FailoverBndupdPendingOutboundQueue               *prometheus.Desc
+	FailoverTransitionsCommunicationinterruptedState *prometheus.Desc
+	FailoverTransitionsPartnerdownState              *prometheus.Desc
+	FailoverTransitionsRecoverState                  *prometheus.Desc
+	FailoverBndupdDropped                            *prometheus.Desc
+}
+
+func NewDhcpCollector() (Collector, error) {
+	const subsystem = "dhcp"
+
+	return &DhcpCollector{
+		PacketsReceivedTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "packets_received_total"),
+			"Total number of packets received by the DHCP server (PacketsReceivedTotal)",
+			nil,
+			nil,
+		),
+		DuplicatesDroppedTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "duplicates_dropped_total"),
+			"Total number of duplicate packets received by the DHCP server (DuplicatesDroppedTotal)",
+			nil,
+			nil,
+		),
+		PacketsExpiredTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "packets_expired_total"),
+			"Total number of packets expired in the DHCP server message queue (PacketsExpiredTotal)",
+			nil,
+			nil,
+		),
+		ActiveQueueLength: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "active_queue_length"),
+			"Number of packets in the processing queue of the DHCP server (ActiveQueueLength)",
+			nil,
+			nil,
+		),
+		ConflictCheckQueueLength: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "conflict_check_queue_length"),
+			"Number of packets in the DHCP server queue waiting on conflict detection (ping). (ConflictCheckQueueLength)",
+			nil,
+			nil,
+		),
+		DiscoversTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "discovers_total"),
+			"Total DHCP Discovers received by the DHCP server (DiscoversTotal)",
+			nil,
+			nil,
+		),
+		OffersTotal: prometheus.NewDesc(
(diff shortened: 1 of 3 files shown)


collector/dhcp.go
```diff
@@ -0,0 +1,412 @@
...
+	DeniedDueToNonmatch                              float64 `perflib:"Denied due to nonmatch."`
+	FailoverBndupdSentTotal                          float64 `perflib:"Failover: BndUpd sent/sec."`
+	FailoverBndupdReceivedTotal                      float64 `perflib:"Failover: BndUpd received/sec."`
+	FailoverBndackSentTotal                          float64 `perflib:"Failover: BndAck sent/sec."`
+	FailoverBndackReceivedTotal                      float64 `perflib:"Failover: BndAck received/sec."`
+	FailoverBndupdPendingOutboundQueue               float64 `perflib:"Failover: BndUpd pending in outbound queue."`
+	FailoverTransitionsCommunicationinterruptedState float64 `perflib:"Failover: Transitions to COMMUNICATION-INTERRUPTED state."`
+	FailoverTransitionsPartnerdownState              float64 `perflib:"Failover: Transitions to PARTNER-DOWN state."`
+	FailoverTransitionsRecoverState                  float64 `perflib:"Failover: Transitions to RECOVER state."`
+	FailoverBndupdDropped                            float64 `perflib:"Failover: BndUpd Dropped."`
```
carlpett, line 246: These also sound counter-y, or how should a gauge value be interpreted?

collector/dhcp.go
```diff
@@ -0,0 +1,412 @@
...
+	OffersTotal                                      float64 `perflib:"Offers/sec"`
+	RequestsTotal                                    float64 `perflib:"Requests/sec"`
+	InformsTotal                                     float64 `perflib:"Informs/sec"`
+	AcksTotal                                        float64 `perflib:"Acks/sec"`
+	NacksTotal                                       float64 `perflib:"Nacks/sec"`
+	DeclinesTotal                                    float64 `perflib:"Declines/sec"`
+	ReleasesTotal                                    float64 `perflib:"Releases/sec"`
+	OfferQueueLength                                 float64 `perflib:"Offer Queue Length"`
+	DeniedDueToMatch                                 float64 `perflib:"Denied due to match."`
+	DeniedDueToNonmatch                              float64 `perflib:"Denied due to nonmatch."`
```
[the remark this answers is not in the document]
breed808: These are almost certainly totals, I'll run this on our DHCP server to confirm the value is increasing over time.

collector/dhcp.go
breed808: I think `FailoverBndupdDropped` is likely a counter, but the other metrics may be indicating the DHCP server failover state. E.G `FailoverTransitionsPartnerdownState 1` may indicate the DHCP server's partner is down, or that 1 partner in the failover cluster is down.

I'll see if I can find more documentation on this, as forcing a partner down in our prod environment may not win me any friends :sweat_smile:

collector/dhcp.go
```diff
@@ -0,0 +1,401 @@
...
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func init() {
+	registerCollector("dhcp", NewDhcpCollector, "DHCP Server")
+}
+
+// A DhcpCollector is a Prometheus collector for both WMI and perflib DHCP metrics
```
carlpett, line 13: Nit: As far as I find in a seconds readthrough now, it doesn't actually have any WMI readouts?

collector/dhcp.go
```diff
@@ -0,0 +1,401 @@
...
+			nil,
+		),
+		PacketsExpiredTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "packets_expired_total"),
+			"Total number of packets expired in the DHCP server message queue (PacketsExpiredTotal)",
+			nil,
+			nil,
+		),
+		MillisecondsPerPacketAvg: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "milliseconds_per_packet_avg"),
```
carlpett, line 66: Is this an actual average, or a counter pair of total time spent and total packets processed? 
Either way, we should expose seconds instead of milliseconds

collector/dhcp.go
carlpett: "Ooops" :smile:

carlpett:
>[...] and WMI queries to this class don't appear to be possible with the stackexchange/wmi module.

Is it a namespaced class, or something more exotic? We do namespaced queries in the process/iis collector: 
https://github.com/martinlindhe/wmi_exporter/blob/91a64fecb850fae0a90219bfcd045c3bc1838cc7/collector/process.go#L189

collector/dhcp.go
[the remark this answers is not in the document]
breed808: I've been running the collector on our DHCP servers for about a week now and haven't seen this metric increment, or change in any way.
Perhaps someone else could run the collector and monitor the metrics for changes?

collector/dhcp.go
breed808: After running the collector for a week, the metric has **not** changed once with a value of '13' and '16' for the primary and second DHCP servers, respectively.
I've omitted the metric from the collector.

collector/dhcp.go
breed808: After monitoring, I've seen the `FailoverTransitionsCommunicationinterruptedState` metric increment from 51 to 55 on the primary, and 11 to 15 on the secondary. I think it's safe to assume this is a counter.
I would also assume the `FailoverTransitionsPartnerdownState` and `FailoverTransitionsRecoverState` are also counters. I'm less certain of `FailoverBndupdDropped` but am comfortable leaving it as a counter.

breed808:
It's been a while since I've looked at the WMI class for the scope statistics. Would it be preferable to investigate the class as a separate collector, to keep the Perflib & WMI components separate?

collector/dhcp.go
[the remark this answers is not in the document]
carlpett: I don't have a Windows DHCP server to test on, unfortunately. I guess we have two options, either leave them out for a first iteration and add them back once someone has a way to validate it, or just go with what seems reasonable and set them to counters and consider it a bug if that turns out to be false.
I'd be fine with the latter, since it seems very likely to be true. What do you think?

docs/collector.dhcp.md
```diff
@@ -0,0 +1,54 @@
...
+None
+
+## Metrics
+
+Name | Description | Type | Labels
+-----|-------------|------|-------
+`packets_received_total` | Total number of packets received by the DHCP server | counter | None
+`duplicates_dropped_total` | Total number of duplicate packets received by the DHCP server | counter | None
+`packets_expired_total` | Total number of packets expired in the DHCP server message queue | counter | None
+`milliseconds_per_packet_avg` | Average time per packet taken by the DHCP server to send a response | gauge | None
```
carlpett, line 23: This doesn't actually seem to be exposed?

carlpett:
>It's been a while since I've looked at the WMI class for the scope statistics. Would it be preferable to investigate the class as a separate collector, to keep the Perflib & WMI components separate?

I think we'd rather have it in a single collector, but it might make sense to split it in different collect-functions as we've done in some places.

collector/dhcp.go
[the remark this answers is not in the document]
breed808: I agree with the latter, let's treat them as Counters and amend if necessary. I've updated the type in a recent push.

docs/collector.dhcp.md
breed808: These are exposed as far as I can tell (metrics are present).

Milliseconds per packet avg metric was dropped in a previous push.

breed808:
I think most of the issues have been ironed out by now, I just need to investigate the WMI class. Will update when this has been done.

breed808:
I've looked into the WMI class by testing the `wmi` library directly. However, this has not returned any data. Running `Get-WmiObject -namespace "root\Microsoft\Windows\DHCP" -query 'SELECT * FROM DhcpServerv4ScopeStatistics'` also returns no data.

I can see [someone else has had this issue](https://social.technet.microsoft.com/Forums/en-US/e1262c31-23fe-43d3-805d-b654b702edf1/wmi-select-from-dhcpserverv4scope-returns-empty?forum=winserveripamdhcpdns) but the general advice appears to be _"Use the `Get-DhcpServerv4ScopeStatistics` PowerShell command instead of querying WMI directly"_.

I'm not sure how this can be queried with WMI :disappointed:

docs/collector.dhcp.md
carlpett: Yes, but it still seems to be in the docs in the latest version?

breed808: Apologies for missing that Calle! I've removed it from the docs.

PR #498, opened 2020-04-03
Fix README for process whitelist and expand docs

Mario-Hofstaetter:
Since PR #489 `--collector.process.processes-where` no longer works, changing example in `README` to use `--collector.process.whitelist` with regexp.

Also expand docs for process filter regexp a little bit.


carlpett:
Thanks @Mario-Hofstaetter!

PR #501, opened 2020-04-10
copy/paste towards [4352] SMB Server Shares

zeph:
- violent copy/paste of metrics
- long but readable names
- copy pasting in the for loop
- 1st round of copy/paste errors
- we shall have them all
- Counter & Gauge assigned correctly
- naming
- focus on the OpenFiles
- our own white/black Lists
- just a word
- we go nowhere with the wrong class mapping
- that was not a counter

### DIFF
--- a/collector/smb_server_shares.go
+++ b/collector/smb_server_shares.go
@@ -1,0 +1,396 @@
+// +build windows
+
+package collector
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+func init() {
+	registerCollector("smb_server_shares", NewSmbServerSharesCollector, "SMB Server Shares")
+}
+
+var (
+	shareWhitelist = kingpin.Flag(
+		"collector.smb_server_shares.share-whitelist",
+		"Regexp of shares to whitelist. Share name must both match whitelist and not match blacklist to be included.",
+	).Default(".+").String()
+	shareBlacklist = kingpin.Flag(
+		"collector.smb_server_shares.share-blacklist",
+		"Regexp of shares to blacklist. Share name must both match whitelist and not match blacklist to be included.",
+	).Default("").String()
+)
+
+// A SmbServerSharesCollector is a Prometheus collector for perflib SmbServerShares metrics
+type SmbServerSharesCollector struct {
+	CurrentOpenFileCount        *prometheus.Desc
+	TotalFileOpenCount          *prometheus.Desc
+	FilesOpenedPerSec           *prometheus.Desc
+	CurrentDurableOpenFileCount *prometheus.Desc
+
+	shareWhitelistPattern *regexp.Regexp
+	shareBlacklistPattern *regexp.Regexp
+}
+
+// NewSmbServerSharesCollector ...
+func NewSmbServerSharesCollector() (Collector, error) {
+	const subsystem = "smb_server_shares"
+
+	return &SmbServerSharesCollector{
+		CurrentOpenFileCount: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "current_open_file_count"),
+			"The number of file handles that are currently open in this share (SmbServerShares.CurrentOpenFileCount)",
+			[]string{"share"},
+			nil,
+		),
+		TotalFileOpenCount: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "file_open_count"),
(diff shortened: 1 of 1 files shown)


collector/smb_server_shares.go
```diff
@@ -0,0 +1,396 @@
...
+	"fmt"
+	"regexp"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+func init() {
+	registerCollector("smb_server_shares", NewSmbServerSharesCollector, "SMB Server Shares")
```
carlpett, line 15: I think maybe we'd like to keep other SMB-related metrics in the same collector, so generalise this to just being the `smb` collector
(Unless there's something I'm missing which makes that a bad idea)

collector/smb_server_shares.go
```diff
@@ -0,0 +1,396 @@
...
+
+	return &SmbServerSharesCollector{
+		CurrentOpenFileCount: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "current_open_file_count"),
+			"The number of file handles that are currently open in this share (SmbServerShares.CurrentOpenFileCount)",
+			[]string{"share"},
+			nil,
+		),
+		TotalFileOpenCount: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "file_open_count"),
```
carlpett, line 52: This should have a `_total` suffix

collector/smb_server_shares.go
```diff
@@ -0,0 +1,396 @@
...
+			"The number of files that have been opened by the SMB File Server on behalf of its clients on this share since the server started. (SmbServerShares.TotalFileOpenCount)",
+			[]string{"share"},
+			nil,
+		),
+		FilesOpenedPerSec: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "files_opened_total"),
```
carlpett, line 58: Isn't this the same as the above metric?

collector/smb_server_shares.go
```diff
@@ -0,0 +1,396 @@
...
+		log.Error("failed collecting smb_server_shares metrics:", desc, err)
+		return err
+	}
+	return nil
+}
+
+// :9432/dump?query=4352#4352
+type SmbServerShares struct {
+	Name string
+	/*
```
carlpett, line 88: Do you intend to add the commented out metrics later, or are they documentation for what else exists?

collector/smb_server_shares.go
```diff
@@ -0,0 +1,396 @@
...
+// to the provided prometheus Metric channel.
+func (c *SmbServerSharesCollector) Collect(ctx *ScrapeContext, ch chan<- prometheus.Metric) error {
+	if desc, err := c.collect(ctx, ch); err != nil {
+		log.Error("failed collecting smb_server_shares metrics:", desc, err)
+		return err
+	}
+	return nil
+}
+
+// :9432/dump?query=4352#4352
```
carlpett, line 85: What does this mean? :)

collector/smb_server_shares.go
```diff
@@ -0,0 +1,396 @@
...
+			ch <- prometheus.MustNewConstMetric(
+				c.ReadBytesPerSec,
+				prometheus.CounterValue,
+				share.ReadBytesPerSec,
+				share.Name,
+			) */
+
+		ch <- prometheus.MustNewConstMetric(
+			c.TotalFileOpenCount,
+			prometheus.GaugeValue,
```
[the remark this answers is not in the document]
zeph: not according to the dump http endpoint from the perflib tool

collector/smb_server_shares.go
zeph: sure, I just reflected the naming structure I saw so far
...I'll change it to smb

collector/smb_server_shares.go
zeph: ack

collector/smb_server_shares.go
zeph: hah, no idea... I did grab what I could out of perflib

collector/smb_server_shares.go
zeph: i wanted to implement ALL i could see from perflib, but then I was running out of time
...this code was developed over the weekend during an outage at a customer as a tool to triage

collector/smb_server_shares.go
zeph: it's the URL query to perflib (I had not found the microsoft documentation reference as u had in there from the copy&paste source file I used... the logic_drivers one)... I believe it is the class id in the array/list in perflib memory

PR #508, opened 2020-04-23
Adds MS Exchange collector

rmyhren:
Adds a collector for MS Exchange. A review would be most welcome :)


collector/exchange.go
```diff
@@ -0,0 +1,679 @@
...
+		if err := collFunc(ch); err != nil {
+			log.Errorf("Error in %s: %s", className(collFunc), err)
+		}
+	}
+	return nil
+}
+
+func (c *exchangeCollector) collectLDAP(ch chan<- prometheus.Metric) error {
+	data := []win32_PerfRawData_MSExchangeADAccess_MSExchangeADAccessProcesses{}
+	if err := wmi.Query("SELECT * FROM "+className(data), &data); err != nil {
```
carlpett, line 347: This pattern is a bit different from other WMI collectors, where we use `queryAll(&data)` (which basically does what you do here). Is there a reason not to?

collector/exchange.go
```diff
@@ -0,0 +1,679 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/StackExchange/wmi"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+const subsystem string = "exchange"
```
carlpett, line 15: This will "pollute" the entire package, which might lead to unexpected usages elsewhere. Could we move it into the func where it is used?

collector/exchange.go
```diff
@@ -0,0 +1,679 @@
+// +build windows
+
+package collector
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/StackExchange/wmi"
```
carlpett, line 9: Did you check if these classes are available as Performance Counters that we could read with perflib instead? That would be much preferred from a performance point of view.

collector/exchange.go
```diff
@@ -0,0 +1,679 @@
...
+func toLabelName(name string) string {
+	return strings.ReplaceAll(strings.Join(strings.Fields(strings.ToLower(name)), "_"), ".", "_")
+}
+
+// newExchangeCollector returns a new Collector
+func newExchangeCollector() (Collector, error) {
+	// https://docs.microsoft.com/en-us/exchange/exchange-2013-performance-counters-exchange-2013-help
+	c := exchangeCollector{
+		// MS Exchange RPC Client Access
+		RPCAveragedLatency:  desc("rpc_avg_latency", "The latency (ms), averaged for the past 1024 packets"),
```
carlpett, line 224: Two points here:
1. Is the reported value actually an average? There's frequently confusion around names which do not reflect the raw data, and this could be one of them.
2. Make to have the unit in the name, and also to use base units. Ie seconds in this case, so you'll need to divide the value read by 1000.

collector/exchange.go
```diff
@@ -0,0 +1,679 @@
...
+		RPCRequests:         desc("rpc_requests", "Number of client requests currently being processed by  the RPC Client Access service"),
+		ActiveUserCount:     desc("rpc_active_user_count", "Number of unique users that have shown some kind of activity in the last 2 minutes"),
+		ConnectionCount:     desc("rpc_connection_count", "Total number of client connections maintained"),
+		RPCOperationsPerSec: desc("rpc_ops_per_sec", "The rate (ops/s) at wich RPC operations occur"),
```
carlpett, line 228: This also looks like something that is probably a counter.

collector/exchange.go
```diff
@@ -0,0 +1,679 @@
...
+			if contains(disabledCollectors, name) {
+				state = "[disabled] "
+			}
+
+			if contains(enabledCollectors, name) {
+				state = "[enabled] "
+			}
+
+			fmt.Printf("%-15s %-32s %-32s\n", state, name, exchangeCollectorFuncDesc[name])
+		}
```
carlpett, line 329: Shouldn't we `os.Exit(0)` after this?

collector/exchange.go
eikaas: Hello and thanks for the review. 

I havent really looked into perlib much at all, but I'm willing to do so. Any way to quickly establish if there are perflib equivalents?

collector/exchange.go
eikaas: Yes, definitely

collector/exchange.go
eikaas: 1. [According to Microsoft](https://docs.microsoft.com/en-us/exchange/exchange-2013-performance-counters-exchange-2013-help) it's the latency averaged for the past 1024 packets. I haven't done any further investigations

2. I'll update the unit names/values where applicable

collector/exchange.go
eikaas: Maybe it served some purpose before, but it doesn't look like it anymore. 😄I'll update it

collector/exchange.go
eikaas: I agree, we probably should

rmyhren:
Hi! Robin has now resolved most of the code issues, and we have checked the various metrics and assigned them what we believe is the correct type based on what we see from running the exporter in a large environment for some time.

collector/exchange.go
carlpett: Hey,
The easiest way is probably "can you find it in perfmon?". The name of the counter object is the "heading" in perfmon, so "Processor Information" for example here:
image

I whipped up a small tool to look at a counter object here: https://github.com/carlpett/perfcounter-inspector
You can clone that and do `go run . <name of counter>` to get a dump of what fields there are, and a sample of the values. Might be useful? 
(I will rewrite the collector generator with this approach)

collector/exchange.go
carlpett: Yes, but Microsoft only documents how it is intended to be visualised, not the actual raw data. 
There's some docs [here](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2003/cc785636(v=ws.10)) on the data types. In previous cases of similarly-named fields, we've seen this as a sum-and-count pair, so you get the total number of milliseconds in this case, and how many operations those are measured over.

collector/exchange.go
rmyhren: Ok, so Robin has changed from WMI to perflib, and we have walked through every metric and to the best of our knowledge they now have the correct type set. Can you have a look?

collector/exchange.go
```diff
@@ -0,0 +1,618 @@
...
+		nil,
+	)
+}
+
+// toLabelName converts strings to lowercase and replaces all whitespace and dots with underscores
+func toLabelName(name string) string {
+	s := strings.ReplaceAll(strings.Join(strings.Fields(strings.ToLower(name)), "_"), ".", "_")
+	s = strings.ReplaceAll(s, "__", "_")
+	return s
+}
```
carlpett, line 608: Since these have pretty generic names and the `collector` package namespace is shared with all other collectors, could they either be renamed to something more specific, or made methods on the collector object? (Eg `func (c *exchangeCollector) desc(...)` rather than free functions)

eikaas: I'll fix the functions. The toLabelName is probably not the first function that does something like this and would probably be better implemented in a general way with tests and such.

The metric names/types are confusing, particularly when Microsoft has been consistently inconsistent when naming things. I'll go over them again

Can I trust the perflib type reporting? Is something called a _per_sec_ but a _counter_-type always a _total_

eikaas: Here is a better view of the metrics if and their types. 
Let me know if you see any errors.

https://gist.github.com/eikaas/8bb26d397b70e049c4339bd299e6c568

carlpett: Yeah, it is frustrating, I know... I don't think we've run into any cases where the types lie, though. So if perflib reports it as a counter, it should be correct, even/especially if the name is something like per second.
For the vast majority of cases, things called per sec has been a counter, and things called something like average latency or such has been a counter of number of occurences/events total, and a base counter with the total time. This isn't amazingly documented in our perflib wrapper, but when there's a base/value pair in a counter you can get the base by throwing a suffix `_Base` on it, so eg `perflib:"RPC Averaged Latency"` likely also has a `perflib:"RPC Averaged Latency_Base"`, where the first is the total time, and the second the total number of rpc:s.

carlpett:
@eikaas Do you still have work planned, or can we remove the draft status and do a final review/merge? :)

carlpett:
Github doesn't send out a notification for "marked ready for review", apparently...

But I'm happy to merge this! Great effort @eikaas and @rmyhren! :tada:

PR #523, opened 2020-05-20
Replace Get-WMIObject with Get-CimInstance for PS7 support

basroovers:
This is somewhat related to https://github.com/martinlindhe/wmi_exporter/issues/522

Where I ran into issues with the generator. I tried to run it in Powershell 7.0.1 but found out that Get-WMIObject is no longer supported as Powershell 7 is based on .Net Core and .Net Core does not support WMI. We will have to use Get-CimInstance instead.


carlpett:
Nice catch, thanks!

PR #527, opened 2020-05-23
feat (service): added support for service StartName

paologallinaharbur:
Adding support for the `startName` (the name of the service account) creating the service.
 
 - In case it is not available the lablel is left empty

Quoting from [the documentation](https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-service?redirectedfrom=MSDN) the explanation of the added label
>StartName:
>Account name under which a service runs. Depending on the service type, the account name may be in the form of "DomainName\Username" or UPN format ("*Username@DomainName*"). 

>The service process is logged by using one of these two forms when it runs. If the account belongs to the built-in domain, then ".\Username" can be specified. For kernel or system-level drivers, StartName contains the driver object name (that is, "\FileSystem\Rdr" or "\Driver\Xns") which the I/O system uses to load the device driver.

>Additionally, if NULL is specified, the driver runs with a default object name created by the I/O system based on the service name.```

### DIFF
--- a/collector/service.go
+++ b/collector/service.go
@@ -45,7 +45,7 @@ func NewserviceCollector() (Collector, error) {
 		Information: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "info"),
 			"A metric with a constant '1' value labeled with service information",
-			[]string{"name", "display_name", "process_id"},
+			[]string{"name", "display_name", "process_id", "start_name"},
 			nil,
 		),
 		State: prometheus.NewDesc(
@@ -89,6 +89,7 @@ type Win32_Service struct {
 	State       string
 	Status      string
 	StartMode   string
+	StartName   *string
 }
 
 var (
@@ -131,9 +132,13 @@ func (c *serviceCollector) collect(ch chan<- prometheus.Metric) (*prometheus.Des
 	if err := wmi.Query(q, &dst); err != nil {
 		return nil, err
 	}
-
 	for _, service := range dst {
 		pid := strconv.FormatUint(uint64(service.ProcessId), 10)
+
+		startName := ""
+		if service.StartName != nil {
+			startName = *service.StartName
+		}
 		ch <- prometheus.MustNewConstMetric(
 			c.Information,
 			prometheus.GaugeValue,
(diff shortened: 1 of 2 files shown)


carlpett:
Hi @paologallinaharbur,
Thanks, functionally, I think everything looks good! Two things I'd like to discuss:
1. Do we want to have empty string for cases where the value is empty, or should there be a placeholder?
2. `start_name` isn't very descriptive of what it means. Do we keep it because it is what Microsoft has named it internally, or change it to something clearer? Microsoft's UIs seem to all refer to it as `Log On As` or `Run As`, both of which I think could be better.

paologallinaharbur:
Hello @carlpett thanks for the quick reply!

Honestly, I had the very same doubts: 
 - regarding the empty values, I thought it was the standard way to deal with these kind of scenarios in prometheus, please correct me if I am wrong! 
For example, in the documentation, in a VBScript they follow a different approach having a default string:
``` if IsNull(startName) then startName = "<None>"```
I could do the same.

 - regarding the `StartName` I decided to stick with the windows naming convention, in this way it would be easy for people to check the official documentation to understand the values it assumes.

In both cases I am open to suggestions, let me know if I should implement it differently and update the PR.
Have a nice day! :)

carlpett:
Sure, if `"<none>"` would be the alternative, then `""` is okay. I wondered if there was something implicit that we could assume (eg it'd run as LOCAL SYSTEM or similar - I haven't done any digging into what it actually means).
Regarding `StartName`, I did a quick poll on some users and none of them knew what `StartName` meant in this context. So I think naming it `run_as` (what they expected it to be named) and then making sure that the help text mentions where it comes from is a good approach. Sounds okay?

paologallinaharbur:
I changed the label name to `run_as`. Now is more clear for sure and I added a snippet in the documentation explaining it.

The issue as I tried to clarify in the documentation is that in case of a null value the service would run as the `LocalSystem` account and, for kernel or system-level drive, it runs with a default object name created by the I/O system based on the service name, for example, `DWDOM\Admin`. [1]

Let me know if you believe further modification should be applied, have a nice day!

[1] https://docs.microsoft.com/en-us/sql/relational-databases/wmi-provider-configuration-classes/sqlservice-class/setserviceaccount-method-sqlservice-class?view=sql-server-ver15

paologallinaharbur:
Rebased and forced pushed in order to resolve conflicts due to the exporter renaming.

carlpett:
:tada:
Thanks!

Issue #519, opened 2020-05-09
Rename to windows_exporter, move to prometheus-community


carlpett:
As decided in #499, we will rename the exporter to windows_exporter. This issue tracks the required changes.
After these steps are completed, we will be ready to [move the repo](https://github.com/prometheus-community/community/issues/13) to the prometheus-community organization :tada:

- [x] Change metric namespace ([collector.go](https://github.com/martinlindhe/wmi_exporter/blob/54d94c261b37836cfc7d5e599f7f14a865b3e7e0/collector/collector.go#L16)).
- [x] Change textfile collector defaults.
- [x] Change Go module name.
- [x] Update build automation.
- [x] Change the installer.
- [x] Update contrib/console_templates.
- [x] Update docs.
- [x] Rename repo.

Mario-Hofstaetter:
This there an ETA on this to prepare for?
Since renaming the metrics is quite the breaking change and we are just rolling out the exporter and dashboards.

carlpett:
Hey @Mario-Hofstaetter,
Very good question! There's a bunch of unreleased changes on master (three new collectors!), so I will make sure that is released as 0.12. Then hopefully do the move within the next week, workload permitting.
I have prepared most of the changes, but ran out of time to validate them. Especially the installer requires a bit of care.

If you want to prepare your environment, you could actually set up a metric relabeling in the Prometheus server to switch the metric namespace already with the current versions. Something like this:

```yaml
  - job_name: windows
    static_configs:
    - targets: ['localhost:9182']
    metric_relabel_configs:
    - source_labels: ['__name__']
      target_label: '__name__'
      regex: 'wmi_(.*)'
      replacement: 'windows_$1'
```

andrewmostello:
On the Chocolatey front, I've created a new package with the new name: [prometheus-windows-exporter.install](https://chocolatey.org/packages/prometheus-windows-exporter.install). That's usable now with v0.12.0 to smooth the transition.

0.12 will be the last release pushed to [prometheus-wmi-exporter.install](https://chocolatey.org/packages/prometheus-wmi-exporter.install). After the name change, I'll deprecate that package according to the Chocolatey guidelines here: [Deprecate a Package](https://chocolatey.org/docs/how-to-deprecate-a-chocolatey-package). All updates will then go to the new package.

If any issues arise, please let me know and I'll address them.

[...]

PR #530, opened 2020-05-24
Rename to windows_exporter

carlpett:
This PR addresses the points in #519, preparing the repo for renaming and moving to prometheus-community.

I've tested upgrading from an installed wmi_exporter 0.11.1, and it seamlessly moves over to being called windows_exporter and uninstalling the old version. However, it does *not* remove the old installation directory completely, leaving the `textfile_inputs` inside. This is apparently a limitation of the Windows Installer framework (unless writing custom code), and didn't remove the directory previously either if you did an uninstall, so I think it'll have to be okay?

Something that **isn't** tested, and quite likely to break, is AppVeyor releases. There we need to generate new Github tokens first, and this can only happen after the repo is moved.

Any other scenarios that needs to be tested?


paologallinaharbur:
Great news and great job! 😄 
Do we have an expected timeline for this change to happen? 
Es. one week, one month, more?

collector/mssql.go
```diff
@@ -396,13 +396,13 @@ func NewMSSQLCollector() (Collector, error) {
 		// meta
 		mssqlScrapeDurationDesc: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "collector_duration_seconds"),
-			"wmi_exporter: Duration of an mssql child collection.",
+			"windows_exporter: Duration of an mssql child collection.",
```
martinlindhe, line 399: Do we need the prefix here at all? It seems other collectors doesnt have a "wmi_exporter:" prefix.

carlpett: I _think_ the idea is to distinguish that this is a metric about the exporter itself, rather than the monitored system? :shrug:

carlpett:
@martinlindhe Thanks! :bow: 

@paologallinaharbur With the review done, I believe we just need to do the human synchronization of moving the repo :) We'll try to get this coordinated asap.

Issue #526, opened 2020-05-21
Add argument to prevent service registration on non interactive sessions 


gsanchezgavier:
Hi @carlpett , 
We are executing the exporter binary from another process without installing it as a service, and I found that the [exporter detects ](https://github.com/martinlindhe/wmi_exporter/blob/master/exporter.go#L313)that is a non interactive session and tries to run as a service. This is causing us to get the Error message 

`Failed to start service: The service process could not connect to the service
controller.`

I would like to propose to add an extra argument flag to skip that registration in case is selected. 
Is this something that you would merge from a PR?
Thanks

carlpett:
Hey @gsanchezgavier,
Interesting! What is your use case for doing that? In general, it isn't a great idea to import a main package, and it'd probably be better to just include the collector package. Would that work?

gsanchezgavier:
Thanks for your answer @carlpett, our use case is that we want to access the metrics through the exporter http interface but we want to have the control of the exporter process. If we just include the collectors packages we will loose the web server interface. 
I was thinking to add a flag like "skip_service_registration" that could keep the current functionality unless is set to true to not register the service in a non interactive session. That will keep the backward compatibility.

carlpett:
I understand your proposed solution, but could you elaborate on what "control of the exporter process" is lacking with running it as-is? Is it something very specific to your environment, or is there a deeper use-case here which could be shared by other users?

gsanchezgavier:
yes sure it means that we execute, terminate and health check the exporter binary from our application, we don't install it form the .msi (we don't want to install it as separate windows service). For this case when we execute the exporter from our application, the exporter detects that is a non interactive session and tries to register with the Service Controller, and we want to avoid that. 
I think the flag could be used by any user in the very same situation but I don't know if is a common case.

gsanchezgavier:
Hi @carlpett  I have added the PR #543  regarding this issue , let me know if there is anything that needs to be clarify. Thanks

morecar:
Hej @carlpett, any thoughts on this? It's been a while.

carlpett:
Hey,
Sorry about the lack of feedback. After quite a lot of deliberation, I don't think this is something that we want to support. While the PR is uncomplicated enough, there's a couple of reasons:
[...]

PR #543, opened 2020-06-12
Add service.skip-run-as-service flag

gsanchezgavier:
Add service.skip-run-as-service flag to skip running the exporter as a service even if is executer from a non interactive session.  #526 

Using this flag can prevent the error `Failed to start service: The service process could not connect to the service controller.` if the exporter binary is being executed from other process.

This was tested on Win Server 16


carlpett:
As decided in #526, this isn't something we want to take on. Sorry for the slow response!

PR #555, opened 2020-06-23
Add endpoint for getting the version of the binary

### LINKED ISSUE #553: Add /version endpoint
We are working on some automatic scripts for installing and upgrading our servers. For that we need some way of checking the version of the running service, so we know if we need to upgrade it or not. 

Prometheus itself has a /version endpoint that just returns a json object, with amongst other, the version of Prometheus. 

If this has any interest, i'll be happy to PR it myself :)

zlepper:
Fixes #553

Just makes it available as plaintext without any really advanced features. 

I also cannot run the project locally due to needing GCC (And honestly i have no idea how to actually set that up)

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	_ "net/http/pprof"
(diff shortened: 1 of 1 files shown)


carlpett:
Thanks @zlepper!
I think it might be nice to follow Prometheus's lead and dump an identical json blob:
https://github.com/prometheus/prometheus/blob/bfcd9282a846c8ebdf66376fba08f673d396891e/web/web.go#L948-L953

As for building, you can disable CGO by setting `CGO_ENABLED=0` as an environment variable. You won't get the `container` collector, but everything else will work.

zlepper:
@carlpett There you go 😃

carlpett:
Thanks!

PR #563, opened 2020-07-04
Set Windows versioninfo on build

### LINKED ISSUE #561: File version missing on the EXE
Hi,

The file version, and product name are missing from the EXE. It makes reporting, deployment and patching livecycle with corporate tools like SCCM sub optimal. 
Does anyone know an easy way to add that on the build process?

carlpett:
This PR adds versioninfo generation for the .exe to the build steps, meaning checking the file properties will now look like this:
image

I also had a moment of inspiration when I saw the field for the icon, so I also threw together the below and if people think it looks okay, it will be the icon for the exe and in Add/Remove programs:
logo-small
(@brian-brazil - I couldn't find any separate license/usage guidelines for the Prometheus logo and my understanding is that this type of derivative work is fine, correct?)

Fixes #561

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ VERSION
 output/
 .vscode
 .idea
+*.syso
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,6 +22,7 @@ install:
   - ps: |
       $env:GO111MODULE="off"
       go get -u github.com/prometheus/promu
+      go get -u github.com/josephspurrier/goversioninfo/cmd/goversioninfo
       $env:GO111MODULE="on"
 
 test_script:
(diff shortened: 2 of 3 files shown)


brian-brazil:
https://www.linuxfoundation.org/trademark-usage/ is the relevant policy, and reading through it "A logo should not be displayed with color variations, or with other elements superimposed on top of the logo.".

JDA88:
Thanks for adding the version info!
For the icon, it's nice, but a bit pointless on a Windows Service has it will unfortunately not be displayed anywhere. Task Manager and Service console don’t display icon on services, it will only show on the explorer.

majerus1223:
It also shows in appwiz.cpl so might not be a bad idea to add an icon to make it look complete.

carlpett:
I reached out to the CNCF trademark team back in July but still haven't heard back, so ditching the logo thing.

JDA88:
Ok, great! If at some point they reach back to you we can still add it back.

Issue #57, opened 2017-03-04
TLS authentication?


dimitertodorov:
Hi,

Have you thought about implementing TLS for the endpoint. I really would prefer to secure the metric endpoints at least with a client cert check.

I know it breaks the normal Prometheus pattern of doing authentication with a proxy.
But on Windows, we really don't have many options for lightweight proxy setup.

Shouldn't be too hard to implement, I could even take a look at writing a patch maybe.

D

martinlindhe:
It certainly won't be hard to implement, the Prometheus stance is the following

> *Why don't the Prometheus server components support TLS or authentication? Can I add those?*
>
> While TLS and authentication are frequently requested features, we have intentionally not implemented > them in any of Prometheus's server-side components. There are so many different options and
> parameters for both (10+ options for TLS alone) that we have decided to focus on building the best
> monitoring system possible rather than supporting fully generic TLS and authentication solutions in
> every server component.
>
> If you need TLS or authentication, we recommend putting a reverse proxy in front of Prometheus.
> See for example [Adding Basic Auth to Prometheus with Nginx](https://www.robustperception.io/adding-basic-auth-to-prometheus-with-nginx/).
>
> Note that this applies only to inbound connections. Prometheus does support [scraping TLS- and
> auth-enabled targets](https://prometheus.io/docs/operating/configuration/#%3Cscrape_config%3E), and other Prometheus components that create outbound connections have similar
> support.

Source: https://prometheus.io/docs/introduction/faq/#why-don%27t-the-prometheus-server-components-support-tls-or-authentication?-can-i-add-those?

martinlindhe:
@brian-brazil Could you weight in on this please? Is it a reasonable idea to add tls support to this exporter given it is more hassle to set up a reverse-proxy in Windows land than in Linux? I've no strong opinion either way.

brian-brazil:
We have no opinion on this for 3rd party integrations from the Prometheus side, we've just chosen not to do this ourselves and also none of us are Windows experts.

martinlindhe:
@brian-brazil Okay thanks!

@dimitertodorov A patch is welcome if you want to have a go at this :-)

carlpett:
@martinlindhe I'd be a bit wary of taking this on, for much the same reasons as in the docs. Maybe a very simple reverse proxy could be written side-by-side instead?

martinlindhe:
[...]

PR #581, opened 2020-07-31
Adding a Kubernetes/Docker config

chzbrgr71:
This PR includes the resources needed to deploy to a container and run in Kubernetes. 

- Dockerfile
- YAML for DaemonSet, Service, ServiceMonitor
- Instructions
- Updated main README to link in

### DIFF
--- a/kubernetes/Dockerfile
+++ b/kubernetes/Dockerfile
@@ -1,0 +1,7 @@
+FROM mcr.microsoft.com/dotnet/framework/aspnet:4.8-windowsservercore-ltsc2019
+
+COPY windows_exporter-0.13.0-amd64.exe C:
+
+ENTRYPOINT [ "c:\\windows_exporter-0.13.0-amd64.exe" ]
+
+EXPOSE 9100
--- a/kubernetes/kubernetes.md
+++ b/kubernetes/kubernetes.md
@@ -1,0 +1,81 @@
+# windows_exporter on Kubernetes
+
+With Kubernetes now supporting Windows nodes (as of [v1.14](https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/)), it is useful to run this windows_exporter as a container on Windows to export metrics for your Prometheus implementation.
+
+Please note: 
+* This implementation uses port 9100. Adjust the Dockerfile as needed if port changes are desired. 
+* Validate the base image against your Windows implementation. When these images do not match, the container may not run as expected.
+* The windows_exporter flags and settings can be customized in the DaemonSet yaml file. 
+
+## Container Image
+
+Follow these steps to create a Docker image with your selected release of the windows_exporter. 
+
+> Note. Ideally there would be a central public repository with pre-built standard images for this. 
+
+1. Download the selected release here. https://github.com/prometheus-community/windows_exporter/releases. This Dockerfile uses the EXE, so select the appropriate one for your platform. 
+2. Create your container image
+
+    ```bash
+    # this step requires a Windows machine 
+    docker build -t your_repo/windows_exporter:latest
+    ```
+
+3. Push to a container registry. Follow steps to make this available in Docker Hub or your private registry.
+
+## Kubernetes DaemonSet
+
+Deploying this DaemonSet will ensure that the exporter is running on all Windows nodes in your cluster. It is important to ensure that the ```nodeSelector``` label in your Windows cluster matches the yaml below. 
+
+```yaml
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app: win-node-exporter
+  name: win-node-exporter
+  namespace: monitoring
+spec:
+  selector:
+    matchLabels:
+      app: win-node-exporter
+  template:
+    metadata:
+      labels:
+        app: win-node-exporter
+    spec:
+      containers:
+      - args: 
+        - --collectors.enabled=os,iis,container
+        - --telemetry.addr=127.0.0.1:9100
+        name: win-node-exporter
+        image: chzbrgr71/windows_exporter:v1.1
+        ports:
+        - containerPort: 9100
+          hostPort: 9100
+          name: https            
+      nodeSelector:
+        kubernetes.io/os: windows            
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    k8s-app: win-node-exporter
+  name: win-node-exporter
+  namespace: monitoring
+spec:
+  type: NodePort
+  ports:
+  - name: https
+    port: 9100
+    targetPort: https
+  selector:
+    app: win-node-exporter
+```
+
+Deploy: 
+
+```bash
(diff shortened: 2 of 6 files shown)


kubernetes/windows-exporter-daemonset.yaml
```diff
@@ -0,0 +1,28 @@
...
+      containers:
+      - args: 
+        - --collectors.enabled=os,iis,container
+        - --telemetry.addr=127.0.0.1:9100
+        name: win-node-exporter
+        image: chzbrgr71/windows_exporter:v1.1
+        ports:
+        - containerPort: 9100
+          hostPort: 9100
+          name: https
```
iusergii, line 26: Does windows_exporter support TLS?

chzbrgr71: I do think TLS and auth need to be handled for this. I saw that TLS came up in the past for this project: https://github.com/prometheus-community/windows_exporter/issues/57 

Also note that the linux prom node_exporter still has TLS as experimental. https://github.com/prometheus/node_exporter#tls-endpoint

iusergii: Cool. I just was confused a bit when tested this solution.

carlpett:
Hey @chzbrgr71,
Thanks for getting this started! I hadn't had any opportunity to run Windows nodes on Kubernetes yet, but I started up a small cluster to test it out. I think the main issue here will be what metrics work, and if they actually match what would have been seen on the node.
Some findings from my experimental setup, please let me know if you had a different experience 
- The container collector did not work (cannot reach the Host Compute Service, likely due to container isolation)
- CPU metrics did not have core information (`core` label was empty), and _seemed_ to be related to the container CPU usage, not host CPU
- In general, most collector information seemed related to what was going on in the container, not the host.

I suspect the container running the exporter is a bit "too isolated", so we'll just collect data on the container itself, but I haven't been able to validate this from my short time.

Maybe @sachinmsft has some insights, or could ping someone from within the Hyper-V/Windows team with knowledge of how this would work?

sachinmsft:
Hi @chzbrgr71 ,
I am confused with this PR . I have below question.
Are you trying to get the node metrics through running wmi-exporter on each node as daemonset like node-exporter gets for Linux node ?

If yes, then I don't think that you can achieve that by running the wmi-exporter inside the pod. reason being that Windows does not support privilege container so you can not run wmi-exporter in the node context.

You have to run the wmi-exporter.exe on the node to get the node metrics. If you want to run it through K8s then you can use the Wins (https://github.com/rancher/wins) to run the process on the node through container. but it adds the dependency of running wins service on each Windows node beforehand.

kubernetes/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM mcr.microsoft.com/dotnet/framework/aspnet:4.8-windowsservercore-ltsc2019
```
ricardona, line 1: Why use this image with the aspnet installation?

Could we just use the base operating system?
```suggestion
FROM mcr.microsoft.com/windows/servercore:ltsc2019
```

kubernetes/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+
+COPY windows_exporter-0.13.0-amd64.exe C:
```
ricardona, line 3: In order not to depend on local files, you could download it from the Internet:

```suggestion
ADD https://github.com/prometheus-community/windows_exporter/releases/download/v0.13.0/windows_exporter-0.13.0-amd64.exe /windows_exporter-0.13.0-amd64.exe
```

kubernetes/kubernetes.md
```diff
@@ -0,0 +1,134 @@
...
+    matchLabels:
+      app: win-node-exporter
+  template:
+    metadata:
+      labels:
+        app: win-node-exporter
+    spec:
+      containers:
+      - args: 
+        - --collectors.enabled=os,iis,container
```
ricardona, line 62: Why `iis`?

chzbrgr71:
Thanks for taking a look at this. 

@sachinmsft - you are correct that I'm trying to get node metrics in the same fashion as node-exporter. It makes sense that this might not work in a Windows scenario. We're not really able to run the wmi-exporter on the node since this is being deployed on AKS and we're not able (allowed) to install software directly on the VM's. That might prevent us from using the Rancher Wins utility as well. 

If you have any other ideas, please let me know. We're seeing more and more AKS customers looking for this capability.

sachinmsft:
Hi @chzbrgr71,
I don't see any other way. We should not need wins once Windows has privilege containers support.

cloudcafetech:
Hi @chzbrgr71,

I am following your repo https://github.com/chzbrgr71/windows_exporter/tree/br-k8s

Deployed daemonset and services

Getting connection refused error ..

image

keikhara:
Just to add on this.. conversation. The way you can collect the metrics data like other vendors like datadog, azmon is using named pipe.  Here's a url which explains this. 
https://dille.name/blog/2017/11/29/using-the-docker-named-pipe-as-a-non-admin-for-windowscontainers/
Bare in mind that the base metrics will not be as elaborate as linux containers as windows containers are different. You may need to do some calculation to normalize for customers who wants to do both linux and windows containers in the same graph.

casey-robertson:
> Just to add on this.. conversation. The way you can collect the metrics data like other vendors like datadog, azmon is using named pipe. Here's a url which explains this.
> https://dille.name/blog/2017/11/29/using-the-docker-named-pipe-as-a-non-admin-for-windowscontainers/
> Bare in mind that the base metrics will not be as elaborate as linux containers as windows containers are different. You may need to do some calculation to normalize for customers who wants to do both linux and windows containers in the same graph.

Where do you tell the Windows exporter to use the localhost listener as a target?

jsturtevant:
Kubernetes now has support for [HostProcess](https://kubernetes.io/blog/2021/08/16/windows-hostprocess-containers/) Containers so we can move forward with this PR :partying_face: 

I've took this work and add the support for hostprocess here: https://github.com/jsturtevant/windows_exporter/tree/use-hostprocess 

@chzbrgr71 Let me know if you are interested in taking this forward otherwise I am happy to submit a separate PR.

Couple screen shots of this working:

node cpu:
Screenshot from 2021-11-05 09-43-57

containers running:
Screenshot from 2021-11-05 09-37-44

targets:
Screenshot from 2021-11-05 09-38-19

perithompson:
ah you got here before me @jsturtevant 😄

carlpett:
Fantastic news! Let us know if it'll go in this or another PR and we can get it reviewed :)

chzbrgr71:
This is awesome. Great work here. This had dropped off my radar over 12 months ago and I know @jsturtevant and team put a lot of work upstream to make it work. 

Feel free to move forward with a PR on this since I don't have cycles now to look at it.

jsturtevant:
opened a new PR: https://github.com/prometheus-community/windows_exporter/pull/864

/close

Issue #586, opened 2020-08-12
Change metric type process_start_time_seconds


suvika17:
Hi. I'm trying to push metrics from windows_exporter to pushgateway (both latest versions), but I get the error:

```
Invoke-webrequest: Submitted metrics are invalid or incompatible with existing metrics: Collected metric "process_start_time_seconds" {label: label: label: label: counter:} is
not a GAUGE
```

The process_start_time_seconds metric given by windows_exporter has a counter type, and Pushgateway can only accept a gauge type. Could you change the metric type to gauge so that I can push metrics to Pushgateway?

electron2302:
Possibly related to  #206

electron2302:
Similar error by just pulling localhost:9182 
When compiling current main branch (Sha: 6bcaee5) I get:
```
An error has occurred while serving metrics:

collected metric process_start_time_seconds counter:<value:1.599226507e+09 >  should be a Gauge
```

All default config, nothing changed or added, used [readme commands](https://github.com/prometheus-community/windows_exporter#usage)

carlpett:
Hi @electron2302,
I would guess that you built the exporter with another version of the Prometheus client libraries than intended. What version of Go do you have? You need to make sure you are on 1.13 or later, and building with modules enabled, otherwise it'll pull in the absolutely latest version.

jetf:
I ran into the same problem when I pushed metrics from windows exporter (v0.14) and jmx exporter (v0.13) from the same host to a pushgateway. The used windows_exporter was the latest download from this site: branch="master",goversion="go1.13.3",revision="922c08b85b60da8a444a57a84126448d7c993307",version="0.14.0" . 

After some search, I suspect that windows_exporter has hardcoded the metric type of `process_start_time_seconds` to COUNTER (which conflicts with other exporters in premetheus core):

- https://github.com/prometheus-community/windows_exporter/blob/f623c0ed89903413c6fd7cc8ab2d447f3d20ea36/exporter.go#L79-L85

- https://github.com/prometheus-community/windows_exporter/blob/f623c0ed89903413c6fd7cc8ab2d447f3d20ea36/exporter.go#L106-L110

I worked around the problem by changing the metric type to gauge in my scraper script before I send the metric to a pushgateway.
Unfortunately, I have not managed to build windows_exporter and submit a PR (I just start learning go). Could someone change the metric type to GAUGE ?

[...]

PR #587, opened 2020-08-14
Fixed generator for single performance counter

electron2302:
When you wanted to create a generator for an performance counter that had just one counter, you would get this error `panic: json: cannot unmarshal object into Go struct field TemplateData.Members of type []main.Member`.

This was caused because the powershell “unboxed” the single element from the list.


carlpett:
Thanks @electron2302!
I'm not in a position where I can test this right now - could you verify that it now works for both single and multiple counter cases?

electron2302:
Hey, @carlpett 
The script that generates the classes works fine in both cases with this change, but the generated class is wrong because the template is wrong (probably outdated) and while trying to fix that I figured out that the current version of main (Sha: 6bcaee5) is not working  (Issue #586) , so I was actually waiting for a fix of that, bevor further fixing the generator.

I will change this PR to draft till complete.

PR #592, opened 2020-09-01
Use perflib to gather metrics in the mssql collector

mallenLF:
The perflib-based access code replaces WMI and is substantially more efficient
and stable, leading to fewer collection timeouts in the mssql collector.

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -58,6 +58,10 @@ var (
 
 func registerCollector(name string, builder collectorBuilder, perfCounterNames ...string) {
 	builders[name] = builder
+	addPerfCounterDependencies(name, perfCounterNames)
+}
+
+func addPerfCounterDependencies(name string, perfCounterNames []string) {
 	perfIndicies := make([]string, 0, len(perfCounterNames))
 	for _, cn := range perfCounterNames {
 		perfIndicies = append(perfIndicies, MapCounterToIndex(cn))
--- a/collector/mssql.go
+++ b/collector/mssql.go
@@ -3778,19 +3778,19 @@ func (c *MSSQLCollector) collectSQLErrors(ctx *ScrapeContext, ch chan<- promethe
 }
 
 type win32PerfRawDataSqlServerTransactions struct {
-	FreeSpaceintempdbKB            float64
-	LongestTransactionRunningTime  float64
-	NonSnapshotVersionTransactions float64
-	SnapshotTransactions           float64
-	Transactions                   float64
-	Updateconflictratio            float64
-	UpdateSnapshotTransactions     float64
-	VersionCleanuprateKBPers       float64
-	VersionGenerationrateKBPers    float64
-	VersionStoreSizeKB             float64
-	VersionStoreunitcount          float64
-	VersionStoreunitcreation       float64
-	VersionStoreunittruncation     float64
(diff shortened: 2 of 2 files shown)


collector/collector.go
```diff
@@ -58,6 +58,10 @@ var (
 
 func registerCollector(name string, builder collectorBuilder, perfCounterNames ...string) {
 	builders[name] = builder
+	addPerfCounterDependencies(name, perfCounterNames)
+}
+
+func addPerfCounterDependencies(name string, perfCounterNames []string) {
```
carlpett, line 64: I can see why this is necessary, but it is a bit of a layering violation. We should probably - in a separate PR - overhaul the startup order so commandline parsing is done before collector initialization so we don't need to muck around in the internals like this in collectors.

collector/mssql.go
```diff
@@ -1848,465 +1879,461 @@ func (c *MSSQLCollector) Collect(ctx *ScrapeContext, ch chan<- prometheus.Metric
 // - Win32_PerfRawData_MSSQLSERVER_SQLServerAccessMethods
 //   https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-access-methods-object
 type win32PerfRawDataSQLServerAccessMethods struct {
```
carlpett, line 1881: With the WMI lib, we had to match type names to class names, but this is no longer the case, so we can give these types names without all the WMI-isms. So something like `mssqlAccessMethods` in this case, maybe? Same pattern for all the other things mentioning perfraw etc.

carlpett (commented):
In general, looks great, thanks for doing the conversion! I'd just like to see some cleanup of the last bits of WMI still in there :)

collector/collector.go
mallenLF: Please see commit 2d015fa3b7aeb9a1f7d57b243065ba44a9fc8f72 containing renamed structs.

collector/mssql.go
mallenLF: Please see commit 2d015fa3b7aeb9a1f7d57b243065ba44a9fc8f72 containing renamed structs.

collector/collector.go
mallenLF: Performing command-line parsing before collector initialization sounds good -- especially in the case of the mssql collector, as I/O will need to be performed, and deferring I/O and other operations that can fail to code that runs outside of an init context would be ideal. I tried to keep this change as minimal as possible to get this working.

carlpett (approved):
Great work, thanks!
Just one question: I don't have a MSSQL environment to test on, did you validate that the metrics output on the existing code and new code is the same? I know one perflib conversion there was some difference in scaling as compared to WMI.

mallenLF:
> Great work, thanks!
> Just one question: I don't have a MSSQL environment to test on, did you validate that the metrics output on the existing code and new code is the same? I know one perflib conversion there was some difference in scaling as compared to WMI.

I did check and they do appear to be the same for the classes I was able to test. I don't have replication configured, so I was not able to test the "availreplica" and "dbreplica" classes, but the rest match up between the v0.13.0 release and this build. I'm really glad you asked though, because when I went back to double-check I found a problem where I misspelled "perflib" in a field annotation, which I fixed.

carlpett:
Nice catch! :) Then this seems good to go, thanks a lot!

PR #608, opened 2020-09-28
Enable the use of remote ip addresses in Windows Firewall exception

siku4:
I was looking for a way to limit the Windows Firewall exception (added by the installer) to the remote IP address of our prometheus server to prevent uninvolved systems from accessing the metrics endpoint. This could also be of interest to other users who want to roll out windows_exporter on a large number of machines (e.g. SCCM).

I've never used WiX before but found the `<fw:RemoteAddress>` tag and added a CLI property. I have successfully tested the following cases:

- `msiexec /i <path-to-msi-file> REMOTE_ADDR=<ip-address-1>`: Adds one IP address to `Remote IP address` in FW rule
- `msiexec /i <path-to-msi-file> REMOTE_ADDR=<ip-address-1>,<ip-address-2>`: Adds two IP addresses to `Remote IP address` in FW rule
- `msiexec /i <path-to-msi-file>`: Leaves the option `Any IP address` activated

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Name | Description
 `LISTEN_PORT` | The port to bind to. Defaults to 9182.
 `METRICS_PATH` | The path at which to serve metrics. Defaults to `/metrics`
 `TEXTFILE_DIR` | As the `--collector.textfile.directory` flag, provide a directory to read text files with metrics from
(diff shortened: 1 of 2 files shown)


carlpett:
Nice work @siku4!
I'll give this a test run as well as soon as I can, but to make this mergeable, could you sign the DCO? Thanks!

carlpett (approved):
Finally had some time to test this out, and it looks great - thanks @siku4!

PR #615, opened 2020-10-08
Convert the tcp collector to use perflib and support TCPv6 counters

mallenLF:
Using perflib is substantially faster and more reliable than using WMI to
retrieve Windows performance counter data.

TCP over IPv6 is distinguished from IPv4 traffic by means of an "af" label
to denote the address family, which will be "ipv4" for TCP over IPv4,
corresponding to the TCPv4 performance object, and "ipv6" for TCP over
IPv6, corresponding to the TCPv6 performance object.

Signed-off-by: Michael Allen <MAllen@laserfiche.com>

### DIFF
--- a/collector/tcp.go
+++ b/collector/tcp.go
@@ -4,13 +4,13 @@ package collector
 
 import (
 	"errors"
-	"github.com/StackExchange/wmi"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 )
 
 func init() {
-	registerCollector("tcp", NewTCPCollector)
+	registerCollector("tcp", NewTCPCollector, "TCPv4")
 }
 
 // A TCPCollector is a Prometheus collector for WMI Win32_PerfRawData_Tcpip_TCPv4 metrics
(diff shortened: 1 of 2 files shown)


mallenLF:
I also have a commit in my local tree to write TCPv6 counters with labels af="ipv4" and af="ipv6" to distinguish between the TCPv4 and TCPv6 performance objects. However, this would be an incompatible change so I decided to leave it out of this PR. But I can push that commit if interested.

carlpett:
Nice @mallenLF! Regarding TCPv6, I'm fine with adding that in as well with the label. It'll make the results more precise, and those who need to distinguish between protocols can then do so.
Also, could you please update the collector docs with it using Perflib rather than WMI?
Thanks!

mallenLF:
> Nice @mallenLF! Regarding TCPv6, I'm fine with adding that in as well with the label. It'll make the results more precise, and those who need to distinguish between protocols can then do so.
> Also, could you please update the collector docs with it using Perflib rather than WMI?
> Thanks!

I pushed commits to export the TCPv6 counters and to update the docs for the tcp collector.

carlpett (approved):
Thanks!

PR #618, opened 2020-10-09
Fix metrics description

ashumkin:
`windows_mssql_locks_lock_timeouts` and
`windows_mssql_locks_lock_timeouts_excluding_NOWAIT` metrics have
exchanged descriptions.

According to
https://docs.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-locks-object
 ---
    Lock Timeouts     | Number of lock requests per second that
    (timeout > 0)/sec | timed out, but excluding requests for NOWAIT locks.
 ---
    Lock Timeouts/sec | Number of lock requests per second that timed
                      | out, including requests for NOWAIT locks.
 ---

Metrics gathered are correct but descriptions are exchanged.

Signed-off-by: Alexey Shumkin <Alex.Crezoff@gmail.com>


carlpett:
Nice catch, thanks!

PR #620, opened 2020-10-09
Use perflib to gather metrics for netframework collectors

mallenLF:
The netframework_* collectors are updated to use perflib rather than WMI
to read metric data from Windows. In addition, flags named
collector.netframework.whitelist and collector.netframework.blacklist
are added to allow processes to be included and excluded by name,
respectively, in the netframework collectors.

Reported metric values are consistent with the WMI implementation,
except for windows_netframework_clrjit_jit_time_percent and
windows_netframework_clrmemory_gc_time_percent. The values reported
by this implementation are consistent with those reported by Windows
Performance Monitor.

### DIFF
--- a/collector/netframework.go
+++ b/collector/netframework.go
@@ -1,0 +1,40 @@
+package collector
+
+import (
+	"fmt"
+	"regexp"
+
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
+	netframeworkWhitelist = kingpin.Flag(
+		"collector.netframework.whitelist",
+		"Regexp of processes to include. Process name must both match whitelist and not match blacklist to be included.",
+	).Default(".*").String()
+	netframeworkBlacklist = kingpin.Flag(
+		"collector.netframework.blacklist",
+		"Regexp of processes to exclude. Process name must both match whitelist and not match blacklist to be included.",
+	).Default("").String()
+)
+
+// A NETFrameworkFlags stores common flags for the netframework family of collectors.
+type NETFrameworkFlags struct {
+	whitelist       *string
+	blacklist       *string
+	whitelistRegexp *regexp.Regexp
+	blacklistRegexp *regexp.Regexp
+}
+
+var netframeworkFlags NETFrameworkFlags
+
+// GetNETFrameworkFlags provides access to common flags for the netframework family of collectors.
+func GetNETFrameworkFlags() *NETFrameworkFlags {
+	if netframeworkFlags.whitelist == nil {
+		netframeworkFlags.whitelist = netframeworkWhitelist
+		netframeworkFlags.blacklist = netframeworkBlacklist
+		netframeworkFlags.whitelistRegexp = regexp.MustCompile(fmt.Sprintf("^(?:%s)$", *netframeworkWhitelist))
+		netframeworkFlags.blacklistRegexp = regexp.MustCompile(fmt.Sprintf("^(?:%s)$", *netframeworkBlacklist))
+	}
+	return &netframeworkFlags
+}
--- a/collector/netframework_clrexceptions.go
+++ b/collector/netframework_clrexceptions.go
@@ -97,8 +97,8 @@ func (c *NETFrameworkCLRExceptionsCollector) collect(ctx *ScrapeContext, ch chan
 		name := process.Name
 		procnum, exists := names[name]
 		if exists {
-			name = fmt.Sprintf("%s#%d", name, procnum)
 			names[name]++
+			name = fmt.Sprintf("%s#%d", name, procnum)
 		} else {
 			names[name] = 1
 		}
--- a/collector/netframework_clrexceptions.go
+++ b/collector/netframework_clrexceptions.go
@@ -3,27 +3,28 @@
 package collector
 
 import (
-	"github.com/StackExchange/wmi"
+	"fmt"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 )
 
 func init() {
-	registerCollector("netframework_clrexceptions", NewNETFramework_NETCLRExceptionsCollector)
+	registerCollector("netframework_clrexceptions", NewNETFrameworkCLRExceptionsCollector, ".NET CLR Exceptions")
 }
 
-// A NETFramework_NETCLRExceptionsCollector is a Prometheus collector for WMI Win32_PerfRawData_NETFramework_NETCLRExceptions metrics
-type NETFramework_NETCLRExceptionsCollector struct {
+// A NETFrameworkCLRExceptionsCollector is a Prometheus collector for Perflib .NET CLR Exceptions metrics
+type NETFrameworkCLRExceptionsCollector struct {
 	NumberofExcepsThrown *prometheus.Desc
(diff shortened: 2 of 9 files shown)


collector/netframework_clrloading.go
```diff
@@ -83,114 +89,134 @@ func NewNETFramework_NETCLRLoadingCollector() (Collector, error) {
...
+	CurrentAssemblies         float64 `perflib:"Current Assemblies"`
+	CurrentClassesLoaded      float64 `perflib:"Current Classes Loaded"`
+	PercentTimeLoading        float64 `perflib:"% Time Loading"`
+	Rateofappdomains          float64 `perflib:"Rate of appdomains"`
+	Rateofappdomainsunloaded  float64 `perflib:"Rate of appdomains unloaded"`
+	RateofAssemblies          float64 `perflib:"Rate of Assemblies"`
+	RateofClassesLoaded       float64 `perflib:"Rate of Classes Loaded"`
+	RateofLoadFailures        float64 `perflib:"Rate of Load Failures"`
+	TotalAppdomains           float64 `perflib:"Total Appdomains"`
+	Totalappdomainsunloaded   float64 `perflib:"Total appdomains unloaded"`
```
carlpett, line 122: Is the mixed cases of `Appdomains`/`appdomains` correct?

collector/netframework.go
```diff
@@ -0,0 +1,40 @@
...
+	blacklistRegexp *regexp.Regexp
+}
+
+var netframeworkFlags NETFrameworkFlags
+
+// GetNETFrameworkFlags provides access to common flags for the netframework family of collectors.
+func GetNETFrameworkFlags() *NETFrameworkFlags {
+	if netframeworkFlags.whitelist == nil {
+		netframeworkFlags.whitelist = netframeworkWhitelist
+		netframeworkFlags.blacklist = netframeworkBlacklist
```
carlpett, line 35: These two don't seem to be used?

collector/netframework.go
```diff
@@ -0,0 +1,40 @@
...
+	blacklist       *string
+	whitelistRegexp *regexp.Regexp
+	blacklistRegexp *regexp.Regexp
+}
+
+var netframeworkFlags NETFrameworkFlags
+
+// GetNETFrameworkFlags provides access to common flags for the netframework family of collectors.
+func GetNETFrameworkFlags() *NETFrameworkFlags {
+	if netframeworkFlags.whitelist == nil {
```
carlpett, line 33: Is this to avoid compiling the regex multiple times, or is there more to it?

carlpett (commented):
Thanks for tackling this one @mallenLF! Had some questions, see below.

We actually have on the backlog to merge these into a single collector with "sub-collectors", like eg the mssql collector. Would you be up for doing that as well?

collector/netframework_clrexceptions.go
```diff
@@ -48,68 +54,88 @@ func NewNETFramework_NETCLRExceptionsCollector() (Collector, error) {
...
 
+		// Append "#1", "#2", etc., to process names to disambiguate duplicates.
+		name := process.Name
+		procnum, exists := names[name]
+		if exists {
+			names[name]++
+			name = fmt.Sprintf("%s#%d", name, procnum)
+		} else {
+			names[name] = 1
+		}
```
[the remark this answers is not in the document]
mallenLF: This was handled by WMI previously. In my testing this renaming logic achieves the same results.

collector/netframework_clrexceptions.go
```diff
@@ -48,68 +54,88 @@ func NewNETFramework_NETCLRExceptionsCollector() (Collector, error) {
...
-	if err := wmi.Query(q, &dst); err != nil {
+func (c *NETFrameworkCLRExceptionsCollector) collect(ctx *ScrapeContext, ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
+	var dst []netframeworkCLRExceptions
+
+	if err := unmarshalObject(ctx.perfObjects[".NET CLR Exceptions"], &dst); err != nil {
 		return nil, err
 	}
 
+	var names = make(map[string]int, len(dst))
 	for _, process := range dst {
```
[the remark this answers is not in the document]
mallenLF: In my testing on multiple machines it is deterministic and the resulting name matches what WMI assigned, as I note above.

collector/netframework_clrexceptions.go
```diff
@@ -48,68 +54,88 @@ func NewNETFramework_NETCLRExceptionsCollector() (Collector, error) {
...
+		procnum, exists := names[name]
+		if exists {
+			names[name]++
+			name = fmt.Sprintf("%s#%d", name, procnum)
+		} else {
+			names[name] = 1
+		}
+
+		// The pattern matching against the whitelist and blacklist has to occur
+		// after appending #N above to be consistent with other collectors.
```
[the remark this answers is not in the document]
mallenLF: Yeah, I think the behavior is not intuitive and it seems like an artifact of how WMI renames the instances for you when querying performance data. I would suggest changing it for all collectors that are subject to this at the same time, however.

collector/netframework.go
mallenLF: That's right, nothing else to it.

collector/netframework.go
mallenLF: They're not used. I was using them for debugging and kept them there in case they proved handy. But I think if we consolidate the collectors into a single one this is no longer relevant.

collector/netframework_clrjit.go
```diff
@@ -48,70 +54,94 @@ func NewNETFramework_NETCLRJitCollector() (Collector, error) {
...
-	Frequency_PerfTime         uint32
-	ILBytesJittedPersec        uint32
-	NumberofILBytesJitted      uint32
-	NumberofMethodsJitted      uint32
-	PercentTimeinJit           uint32
-	StandardJitFailures        uint32
-	TotalNumberofILBytesJitted uint32
+	Frequency_PerfTime         float64 `perflib:"Not Displayed_Base"`
+	ILBytesJittedPersec        float64 `perflib:"IL Bytes Jitted / sec"`
+	NumberofILBytesJitted      float64 `perflib:"# of IL Bytes Jitted"`
```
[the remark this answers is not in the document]
mallenLF: You're right, and they do seem to be redundant rather than missing from being exported. I'll remove them when consolidating the collectors.

collector/netframework_clrloading.go
mallenLF: Yes, the mixed casing is correct. We could change the name of the struct members, of course, but those are exactly how the performance counters are spelled in Windows.

mallenLF:
> Thanks for tackling this one @mallenLF! Had some questions, see below.
> 
> We actually have on the backlog to merge these into a single collector with "sub-collectors", like eg the mssql collector. Would you be up for doing that as well?

I created a new PR, #624 to combine the netframework_clr* collectors into a single collector with a configurable set of child collectors like mssql. If that code appears to be heading in the right direction I suggest we just abandon and close this PR. I tried to address your comments in this PR in that PR.

carlpett:
I think the new PR looks very good, so let's close this one in favour of that :)

PR #621, opened 2020-10-10
none to yes

madhuripatil2:
yes to none


carlpett:
Hi,
The proposed change take correct documentation and makes it incorrect, and I'm going to close and label it as Hacktoberfest spam for now. Please open an issue if you think there is something wrong with the docs, with clear reasoning.

PR #622, opened 2020-10-10
Clarify supported versions of Windows

### LINKED ISSUE #611: Installer should fail on unsupported versions of Windows
Go 1.13+ do not support Windows versions below 2008 (ie we require 2008R2 or later). Currently the installer still allows installation, but then fails to start the service. We should add a check to the msi so that it is made clear to users that it will not work.

https://wixtoolset.org/documentation/manual/v3/xsd/wix/package.html
https://docs.microsoft.com/en-us/windows/win32/msi/released-versions-of-windows-installer

carlpett:
Installation on versions previous to Win7/Server2008R2 will now fail.

Fixes #611

Ref
https://wixtoolset.org/documentation/manual/v3/xsd/wix/package.html
https://docs.microsoft.com/en-us/windows/win32/msi/released-versions-of-windows-installer


breed808 (approved):
Looks good!

PR #624, opened 2020-10-11
Combine the netframework_clr* collectors into a single collector

mallenLF:
The new collector is named netframework_clr and supports a configurable
set of child collectors, similar to the mssql collector. Other changes
include:

- Perflib is used instead of WMI.
- Process name whitelist and blacklist flags are added to
  specify regexp to use for matching for inclusion and exclusion,
  respectively.

There is no change to metric names or labels.


collector/netframework_clr.go
```diff
@@ -0,0 +1,1271 @@
...
+
+	// Append a "#1", "#2", etc., suffix if a process name appears more than once.
+	// WMI does this automatically, but Perflib does not.
+	procnum, exists := nameCounts[name]
+	if exists {
+		nameCounts[name]++
+		name = fmt.Sprintf("%s#%d", name, procnum)
+	} else {
+		nameCounts[name] = 1
+	}
```
carlpett, line 604: I did some testing on this after the discussions in the other PR - on my machine at least, I do not get a stable naming with this approach. The data returned by perflib appears to come sorted by PID, maybe? Not sure this is the ordering, but it is at least not by start order. So, starting a new process with the same name as existing ones may "displace" them in the naming, so `#2` might be `#3`, and so on. And a process terminating will also cause re-ordering, of course.

This also appears to happen with WMI, so it is not a problem introduced by perflib as such, but we should probably try to fix it if we can. From what I can tell, the performance object for the .NET CLR stuff has a `Process ID` field, so we do have that data to use. Two possible paths:
- Add a `pid` label
- Use the process id as the #-suffix

Any reasons not to use the first option?

docs/collector.netframework_clr.md
```diff
@@ -0,0 +1,102 @@
...
+`Win32_PerfRawData_NETFramework_NETCLRMemory`<br>
+`Win32_PerfRawData_NETFramework_NETCLRRemoting`<br>
+`Win32_PerfRawData_NETFramework_NETCLRSecurity`<br>
+Enabled by default? | No
+
+## Flags
+
+### `--collector.netframework_clr.collectors-enabled`
+
+Comma-separated list of child collectors to use. Supported values are `exceptions`, `interop`,
```
carlpett, line 23: We should mention that the default value is all of them

carlpett (commented):
Thanks for getting this going! Did a first pass over it, with some comments

breed808:
@mallenLF I've created [my own branch](https://github.com/prometheus-community/windows_exporter/compare/master...breed808:netframework-combined-fixes) with the fixes that Calle requested. Would you like to merge it to your branch for this Pull Request?
Note that I've been unable to test if the `pid` label is working correctly, so someone will need to confirm that before anything is merged.

breed808:
Nevermind, I've been able to push the commits to the PR source branch. @carlpett can you have a look at this when you're free?

collector/netframework_clr.go
```diff
@@ -0,0 +1,1218 @@
...
+	case "locksandthreads":
+		return ".NET CLR LocksAndThreads"
+	case "memory":
+		return ".NET CLR Memory"
+	case "remoting":
+		return ".NET CLR Remoting"
+	case "security":
+		return ".NET CLR Security"
+	default:
+		return ""
```
carlpett, line 92: We should probably fail for invalid configuration here?

collector/netframework_clr.go
```diff
@@ -0,0 +1,1218 @@
...
+	}
+
+	netclrCollector.netclrCollectors = netclrCollector.getNETCLRCollectors()
+	if *netclrPrintCollectors {
+		fmt.Printf("Available .NET CLR sub-collectors:\n")
+		for name := range netclrCollector.netclrCollectors {
+			fmt.Printf(" - %s\n", name)
+		}
+		os.Exit(0)
+	}
```
carlpett, line 532: Perhaps move this to the top of the initialization?

collector/netframework_clr.go
```diff
@@ -0,0 +1,1218 @@
...
+			"windows_exporter: Whether a netframework_clr child collector was successful.",
+			[]string{"collector"},
+			nil,
+		),
+
+		// .NET CLR Exceptions metrics
+		NumberofExcepsThrown: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, exceptionsSubsystem, "exceptions_thrown_total"),
+			"Displays the total number of exceptions thrown since the application started. This includes both .NET exceptions and unmanaged exceptions that are converted into .NET exceptions.",
+			[]string{"process"},
```
carlpett, line 214: This was changed to `pid`, right?

collector/netframework_clr.go
```diff
@@ -0,0 +1,1218 @@
...
+
+		err := c.execute(ctx, name, function, ch)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *netclrCollector) netclrMapProcessName(name string, nameCounts map[string]int) (string, bool) {
```
carlpett, line 576: This isn't used any more, right?

collector/netframework_clr.go
breed808: Yes, I've set the function to return `(string, error)` and handled that with each call.
`default` case now returns an error.

collector/netframework_clr.go
breed808: Resolved in latest push. I've moved it to the top of the `newNETCLRCollector()` function.

collector/netframework_clr.go
breed808: Good spot, that was my mistake forgetting to set the correct flag name. Thanks for that :+1:

collector/netframework_clr.go
breed808: Correct, I've removed it now.

docs/collector.netframework_clr.md
```diff
@@ -0,0 +1,104 @@
...
+### `--collector.netframework_clr.blacklist`
+
+Regexp of processes to exclude.  Process name must both match whitelist and not
+match blacklist to be included.
+
+## Metrics
+
+Name | Description | Type | Labels
+-----|-------------|------|-------
+`windows_netframework_clrexceptions_exceptions_thrown_total` | Displays the total number of exceptions thrown since the application started. This includes both .NET exceptions and unmanaged exceptions that are converted into .NET exceptions. | counter | `process`
```
carlpett, line 47: Labels need to be updated

carlpett (commented):
LGTM, just one missing update in the docs.
So we just need to do some testing?

carlpett:
Ran a test on this now. Sadly, it does not work, there is no `Process ID` field on the performance counter (meaning all of the processes get a label `pid=""`, which is invalid and errors the scrape).
There appears to be a value you can [set in the registry](https://web.archive.org/web/20190420052123/https://support.microsoft.com/en-us/help/281884/the-process-object-in-performance-monitor-can-display-process-ids-pids) to affect the suffix (`_PID` instead of `#N`) , but we shouldn't be injecting stuff into users' registries, IMO.

Another option I found mentioned, but don't have time to look at right now is PerformanceCounterCategories, which apparently has a mapping. It's mentioned [here](https://weblog.west-wind.com/posts/2014/Sep/27/Capturing-Performance-Counter-Data-for-a-Process-by-Process-Id) with some C# code. No idea how viable it'd be to extend our perflib implementation to support this yet.

breed808:
That's disappointing, I was hoping the `Process ID` field would be present.
A quick search found [go-dotnet](https://github.com/matiasinsaurralde/go-dotnet) which _might_ be used with PerformanceCounterCategories, but some discussion would be required to determine if options like this are worth the complexity.

I don't have much C# experience so it's possible there are better tools to achieve this.

daveaowen:
@carlpett @breed808 could we possibly keep the old netframework* collectors and have the combined Perflib collector as an alternative, with documentation stating the need for the registry change? 

We could also check for the registry key and if it doesn't exist, default to WMI. The IIS collector changes the behaviour based on a registry key as well:

``` golang
func getIISVersion() simple_version {
	k, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\InetStp\`, registry.QUERY_VALUE)
	if err != nil {
		log.Warn("Couldn't open registry to determine IIS version:", err)
		return simple_version{}
	}
	defer func() {
		err = k.Close()
		if err != nil {
			log.Warnf("Failed to close registry key: %v", err)
		}
	}()

	major, _, err := k.GetIntegerValue("MajorVersion")
	if err != nil {
		log.Warn("Couldn't open registry to determine IIS version:", err)
		return simple_version{}
	}
	minor, _, err := k.GetIntegerValue("MinorVersion")
	if err != nil {
		log.Warn("Couldn't open registry to determine IIS version:", err)
		return simple_version{}
	}

	log.Debugf("Detected IIS %d.%d\n", major, minor)

	return simple_version{
		major: major,
		minor: minor,
	}
}
```

mallenLF:
Closing due to `Process ID` not being available and the other code changing a lot since this PR was originally made.

Issue #583, opened 2020-08-05
Possible memory leak on v0.11.0?


dcrn:
We're running wmi_exporter v0.11.0 on about 90+ VMs and have noticed wmi_exporter using up to 1GB of memory in some cases. Further investigation revealed that memory usage is growing over time, which may indicate a memory leak. See the image below:
![wmi_exporter](https://user-images.githubusercontent.com/1385653/89429590-a59d2800-d735-11ea-85b4-eec71cda79ea.png)

Could this be caused by a lack of filtering on the process/service collectors?

Version:
```
wmi_exporter, version 0.11.0 (branch: master, revision: 1d7747b4d1c1fc91961a3e8ce7d5b6b0cd642576)
  build user:       appvyr-win\appveyor@APPVYR-WIN
  build date:       20200401-19:13:52
  go version:       go1.13.3
```

Startup parameters:
```
--log.format logger:eventlog?name=wmi_exporter 
--collectors.enabled os,cs,cpu,logical_disk,logon,memory,net,process,service,system,textfile,vmware 
--telemetry.addr :9182   
```

carlpett:
Hi @dcrn,
This _might_ be the same as in #446. Is this repeatable in your environment? I do not believe we got around to updating the library since we couldn't see if it helped, so if we can see it on your environment that'd be very helpful.

dcrn:
Hi @carlpett,
Thanks for the advice. I grabbed a minidump from the process and saw every thread was waiting. The event viewer also showed that several of the collectors were constantly timing out. This seems to fit with the thread lock/unlock theory.

I built the project with the wmi library commit you linked in the other issue, will get back to you if it works out.

Thanks.

dcrn:
I've left the new build running on 22 servers for the past week, and haven't noticed any problems yet. The memory usage does slowly increase over time, but it will take another few weeks to determine if it will level out:
![image](https://user-images.githubusercontent.com/1385653/90409212-c4d67680-e0a0-11ea-842f-816da81768ee.png)

One of the other servers which I hadn't updated had wmi_exporter using over 1GB of memory, but it didn't show up in the "process" data. I stopped wmi_exporter, loaded the new build onto it. There was still no "process" data and soon jumped up to 500MB usage. I tracked down a bunch more servers like this that had `wmi_exporter_collector_timeout{collector="process"} 1`, and saw that restarting the WMI service resolved the problem.

[...]

Issue #446, opened 2019-12-10
wmi.(*SWbemServices).Query / syscall.Syscall9 sometimes gets stuck indefinitely


GiedriusS:
With `0.8.3` we've run into a situation where thousands of goroutines were stuck on:

```
goroutine 280 [semacquire, 279 minutes]:
sync.runtime_SemacquireMutex(0xc00010fd3c, 0x0)
	C:/go/src/runtime/sema.go:71 +0x44
sync.(*Mutex).Lock(0xc00010fd38)
	C:/go/src/sync/mutex.go:134 +0x110
github.com/martinlindhe/wmi_exporter/vendor/github.com/StackExchange/wmi.(*SWbemServices).Query(0xc00010fd10, 0xc000384060, 0x22, 0x9269a0, 0xc0004ac080, 0x0, 0x0, 0x0, 0x0, 0x0)
	c:/gopath/src/github.com/martinlindhe/wmi_exporter/vendor/github.com/StackExchange/wmi/swbemservices.go:142 +0x52
github.com/martinlindhe/wmi_exporter/vendor/github.com/StackExchange/wmi.Query(0xc000384060, 0x22, 0x9269a0, 0xc0004ac080, 0x0, 0x0, 0x0, 0x3c515e8, 0x1b02)
	c:/gopath/src/github.com/martinlindhe/wmi_exporter/vendor/github.com/StackExchange/wmi/wmi.go:78 +0x90
github.com/martinlindhe/wmi_exporter/collector.(*CSCollector).collect(0xc00010a540, 0xc0001d03c0, 0x4cb895, 0x0, 0xc000026070)
	c:/gopath/src/github.com/martinlindhe/wmi_exporter/collector/cs.go:63 +0x91
github.com/martinlindhe/wmi_exporter/collector.(*CSCollector).Collect(0xc00010a540, 0xc0005d61e0, 0xc0001d03c0, 0x50, 0x50)
	c:/gopath/src/github.com/martinlindhe/wmi_exporter/collector/cs.go:46 +0x43
main.execute(0xc0000a6008, 0x2, 0xae7ba0, 0xc00010a540, 0xc0005d61e0, 0xc0001d03c0, 0x6a3e83)
	c:/gopath/src/github.com/martinlindhe/wmi_exporter/exporter.go:205 +0x8f
[...]

PR #631, opened 2020-10-19
Upgrade WMI library

dcrn:
This upgrades the wmi library as described in #583 and #446.
It has helped to some extent with high memory usage, but whether this solves all causes of it is unknown.
Regardless, upgrading the library has proven to be stable while running on 22 production servers for the past 9-10 weeks (See comments in #583 for further details).

@carlpett


carlpett:
Thanks! Could you sign the DCO? Apart from that it looks great to :)

dcrn:
> 
> 
> Thanks! Could you sign the DCO? Apart from that it looks great to :)

All done now. I had troubles because I resolved conflicts before amending.

PR #636, opened 2020-10-21
Set mssql transactions_active to Gauge.

### LINKED ISSUE #634: Consider changing `windows_mssql_transactions_active_total` to type `gauge`
Disclaimer: I'm new to Prometheus, so I don't understand all the ins and outs of it. Also, I don't understand how the metrics for SQL server are collected yet.

When monitoring the `windows_mssql_transactions_active_total` metric I don't see monotonically increasing numbers as described in the [relevant part in the Prometheus docs](https://prometheus.io/docs/concepts/metric_types/#counter). This can be also seen in this screenshot:

transactions_active_total

From that I conclude this metric should be of type `prometheus.GaugeValue` [here](https://github.com/prometheus-community/windows_exporter/blob/78918f70349b8b65c642a84c7f0ef18bc6320a26/collector/mssql.go#L3836).

breed808:
Change also renames metric to avoid confusion, so will need to be mentioned as a breaking change for next release.
Fixes #634.

### DIFF
--- a/collector/mssql.go
+++ b/collector/mssql.go
@@ -387,7 +387,7 @@ type MSSQLCollector struct {
 	TransactionsLongestTransactionRunningSeconds *prometheus.Desc
 	TransactionsNonSnapshotVersionActiveTotal    *prometheus.Desc
 	TransactionsSnapshotActiveTotal              *prometheus.Desc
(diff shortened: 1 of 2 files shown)


carlpett (approved):
Looks good to me!

PR #639, opened 2020-10-25
Add Windows Time Service collector

### LINKED ISSUE #532: Time metrics
Since node_exporter has time metrics, why does wmi_exporter not have them yet?
I've found `Win32_PerfRawData_MicrosoftWindowsW32TimePerf_WindowsTimeService` which has the following data:
- ClockFrequencyAdjustment
- ComputedTimeOffset
- NTPClientTimeSourceCount
- NTPRoundtripDelay
- NTPServerIncomingRequests
- NTPServerOutgoingResponses

The only thing stopping me from generating a collector is that the most important metric `ComputedTimeOffset` is a `ulong`, which is a 64-bit integer. (https://docs.microsoft.com/en-us/dotnet/visual-basic/language-reference/data-types/ulong-data-type)

breed808:
Will resolve #532.

Note that despite testing the collector on a Windows 10 device with the NTP client enabled, I have not noted any change in the `windows_time_clock_frequency_adjustment_ppb_total` metric. Further testing of this metric by others would be welcome.

### DIFF
--- a/collector/time.go
+++ b/collector/time.go
@@ -1,0 +1,124 @@
+// +build windows
+
+package collector
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+)
+
+func init() {
+	registerCollector("time", newTimeCollector, "Windows Time Service")
+}
+
+// TimeCollector is a Prometheus collector for Perflib counter metrics
+type TimeCollector struct {
+	ClockFrequencyAdjustmentPPBTotal *prometheus.Desc
+	ComputedTimeOffset               *prometheus.Desc
+	NTPClientTimeSourceCount         *prometheus.Desc
+	NTPRoundtripDelay                *prometheus.Desc
+	NTPServerIncomingRequestsTotal   *prometheus.Desc
+	NTPServerOutgoingResponsesTotal  *prometheus.Desc
+}
+
+func newTimeCollector() (Collector, error) {
+	const subsystem = "time"
+
+	return &TimeCollector{
+		ClockFrequencyAdjustmentPPBTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "clock_frequency_adjustment_ppb_total"),
+			"Total adjustment made to the local system clock frequency by W32Time in Parts Per Billion (PPB) units.",
+			nil,
+			nil,
+		),
+		ComputedTimeOffset: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "computed_time_offset_seconds"),
+			"Absolute time offset between the system clock and the chosen time source, in seconds",
+			nil,
+			nil,
+		),
+		NTPClientTimeSourceCount: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "ntp_client_time_source_count"),
+			"Active number of NTP Time sources being used by the client",
+			nil,
+			nil,
+		),
+		NTPRoundtripDelay: prometheus.NewDesc(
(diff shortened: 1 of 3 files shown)


collector/time.go
```diff
@@ -0,0 +1,124 @@
...
+
+	return &TimeCollector{
+		ClockFrequencyAdjustmentPPBTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "clock_frequency_adjustment_ppb_total"),
+			"Total adjustment made to the local system clock frequency by W32Time in Parts Per Billion (PPB) units.",
+			nil,
+			nil,
+		),
+		ComputedTimeOffset: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "computed_time_offset"),
```
carlpett, line 35: Having the unit suffix would be nice here
```suggestion
			prometheus.BuildFQName(Namespace, subsystem, "computed_time_offset_seconds"),
```

collector/time.go
```diff
@@ -0,0 +1,124 @@
...
+			nil,
+		),
+		NTPClientTimeSourceCount: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "ntp_client_time_source_count"),
+			"Active number of NTP Time sources being used by the client",
+			nil,
+			nil,
+		),
+		NTPRoundtripDelay: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "ntp_round_trip_delay"),
```
carlpett, line 47: Add unit
```suggestion
			prometheus.BuildFQName(Namespace, subsystem, "ntp_round_trip_delay_seconds"),
```

docs/collector.time.md
```diff
@@ -0,0 +1,45 @@
+# time collector
+
+The time collector exposes the Windows Time Service metrics. Note that the Windows Time Service must be running, else metric collection will fail.
+If the Windows Time Service is stopped after collection has started, collector metric values will reset to 0.
+
+|||
+-|-
+Metric name prefix  | `time`
+Data source         | Perflib
+Enabled by default? | No
```
carlpett, line 10: This would make sense to me to enable by default, any reason we shouldn't?

docs/collector.time.md
```diff
@@ -0,0 +1,45 @@
...
+### Example metric
+_This collector does not yet have explained examples, we would appreciate your help adding them!_
+
+## Useful queries
+_This collector does not yet have any useful queries added, we would appreciate your help adding them!_
+
+## Alerting examples
+**prometheus.rules**
+```yaml
+# Alert on hosts with more than 80% CPU usage over a 10 minute period
```
carlpett, line 36: Copy/paste remains? :)

docs/collector.time.md
```diff
@@ -0,0 +1,45 @@
...
+- alert: NTPClientDelay
+  expr: ntp_round_trip_delay_total > 1
```
carlpett, line 38: Typo? Also add the unit if changed above
```suggestion
  expr: windows_time_ntp_round_trip_delay_seconds > 1
```

carlpett (commented):
Very nice! :cake: Some missing units-related comments

docs/collector.time.md
breed808: Collector is dependant on the Windows Time Service, which wasn't enabled by default on my Windows 10 VM.
Would that be a blocker to setting this as a default collector?

breed808:
Thanks for picking up on my typos & missing unit suffixes Calle :+1:

docs/collector.time.md
carlpett: Ah, okay. I had it enabled on the systems I checked quickly, but if it isn't a default then let's keep this one non-default as well.

carlpett (approved):
:tada:

Issue #614, opened 2020-10-07
Exchange: "Error in ADAccessProcesses: counter not found"


Biomains:
2 error messages in Application eventlog.

collector exchange failed after 0.000999s: counter not found
Error in ADAccessProcesses: counter not found

This is on Windows 2016 Std, with MS Exchange server 2016.

I had this on both wmi_exporter v.01.12.0 and on the latest windows_exporter v.014.0

carlpett:
@rmyhren @eikaas Either of you seen this not being present before? I suppose a workaround here could be allowing to disable individual sub-collectors if they are somehow not available.

timweissenfels:
I'm afraid the issue still persists @fischerman

Issue #569, opened 2020-07-10
Exchange collector returns subset of metrics with Exchange 2010


scluff:
## Expected Behavior
Only counters which are not found should be omitted (HttpProxy and WorkloadManagement)

## Actual Behavior
Only metrics sent to the channel before an error is encountered are scraped

## Steps to Reproduce the Problem
```
[user@prometheus ~]# for i in {1..10}; do curl -s http://exchange.example.com:9182/metrics | grep windows_exchange | grep -v '# '| wc -l; sleep 1; done
0
160
0
0
1
6
170
3
0
0
```

## Specifications
  - Version: v0.13.0
  - Platform: Exchange 2010 SP3 RU17 on Windows Server 2008 R2 SP1

scluff:
If I remove the following lines, I no longer encounter see a varying number of returned metrics.
https://github.com/prometheus-community/windows_exporter/blob/6bcaee58853ce4700b820058237501e6727530be/collector/exchange.go#L172
https://github.com/prometheus-community/windows_exporter/blob/6bcaee58853ce4700b820058237501e6727530be/collector/exchange.go#L177

carlpett:
Hi @scluff,
Thanks for reporting! That is indeed a bug, we shouldn't be returning here, since that gives partial results:
https://github.com/prometheus-community/windows_exporter/blob/6bcaee58853ce4700b820058237501e6727530be/collector/exchange.go#L180-L183
What are the returned errors you get? Depending on what is going wrong, there's different ways to approach this.

scluff:
I see one of these in the log:
> level=error msg="Error in HttpProxy: counter not found" source="exchange.go:181"
> level=error msg="Error in WorkloadManagement: counter not found" source="exchange.go:181"

Followed by:
> level=error msg="collector exchange failed after 0.000000s: counter not found" source="exporter.go:206"

carlpett:
Ok, thanks. It's been many moons since I last managed Exchange servers - would I be correct in guessing that you haven't installed roles corresponding to those? What I'm looking for is really if these are expected conditions in some installs, or a sign that something is broken?
Generally speaking, if something is broken it'd be more in line with Prometheus standard behaviours to _always_ serve no metrics at all from this collector, while if it is expected, then maybe we shouldn't even log an error.

scluff:
I am not an Exchange expert by any stretch, but from what I have read HttpProxy may be related to a feature called Outlook Anywhere; we do not have this enabled.

[...]

PR #642, opened 2020-10-29
adds whitelist for exchange collectors

fischerman:
related to #614 and #569 

The flag enables the windows_exporter to work with different exchange installations where not all performance counters are available. Detecting the installed roles would be even nicer, but is much more work. Either way this is a nice way of customizing the exporter.

As a side effect of this PR, the order in which the collectors are called is now fixed. As a result, if one of the sub collectors fails the returned metrics are now consistent.

### DIFF
--- a/collector/exchange.go
+++ b/collector/exchange.go
@@ -82,10 +82,17 @@ var (
 		"RpcClientAccess",
 	}
 
+	collectorWhitelist = make([]string, 0, len(exchangeAllCollectorNames))
+
 	argExchangeListAllCollectors = kingpin.Flag(
 		"collectors.exchange.list",
 		"List the collectors along with their perflib object name/ids",
 	).Bool()
+
+	argExchangeCollectorsWhitelist = kingpin.Flag(
+		"collectors.exchange.whitelist",
+		"Comma-separated list of collectors to use. Defaults to all, if not specified.",
+	).Default("").String()
 )
 
 // newExchangeCollector returns a new Collector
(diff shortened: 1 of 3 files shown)


breed808 (changes requested):
Hi Björn, thanks for submitting this PR! I've added a couple review comments, have a look at them and let me know if there's any issues.
Whitelisting of the Exchange collectors might assist with performance in some scenarios, so this is welcome.

fischerman:
@breed808 I've implemented your feedback.

breed808 (approved):
Looks good! @carlpett do you want to have a look at this before it's merged?

collector/exchange.go
```diff
@@ -607,3 +631,12 @@ func (c *exchangeCollector) toLabelName(name string) string {
...
 }
+
+func find(slice []string, val string) bool {
+	for _, item := range slice {
+		if item == val {
+			return true
+		}
+	}
+	return false
+}
```
carlpett, line 642: Bit of a nit, but we've tried to keep generic utility functions in collector.go, since they are shared across the entire package.

collector/exchange.go
```diff
@@ -82,10 +82,17 @@ var (
 		"RpcClientAccess",
 	}
 
+	collectorWhitelist = make([]string, 0, len(exchangeAllCollectorNames))
```
carlpett, line 85: Could we make this a field on the `exchangeCollector` type? If not, it at least needs a more exchange-specific name, since it is a package level var.

carlpett (commented):
Thanks! Looks good functionally, just some structural comments

fischerman:
@carlpett Done! I've moved the whitelist in the struct. I don't have permissions to merge.

carlpett:
Great! One minor thing that struck me now that I took a final pass was if whitelist is the right term here, since there is no corresponding blacklist? I think we usually just have `enabled` in similar places (eg `collectors.enabled` and `collector.mssql.enabled`)
Sorry for not thinking about it in the previous check!

fischerman:
Agreed! I've renamed it. Ready to merge from my side. Looking forward to v0.15 :)

carlpett (approved):
:tada: Thanks!

rmyhren:
Wow, this is great. Thanks for the work, @fischerman. We have not had the time to dig into this yet. Another benefit of providing upstream commits 👍

PR #651, opened 2020-11-03
Fix exporter failing to start without config flag

### LINKED ISSUE #647: No config present
The Windows exporter fails to start when no config file is present.

```
C:\Users\Administrator\go\src\github.com\prometheus-community\windows_exporter>.\windows_exporter.exe
time="2020-11-02T14:30:02+01:00" level=info msg="Loading configuration file: C:\\Program Files\\windows_exporter\\config.yml" source="config.go:37"
time="2020-11-02T14:30:02+01:00" level=fatal msg="could not load config file: CreateFile C:\\Program Files\\windows_exporter\\config.yml: The system cannot find the path specified.\n" source="exporter.go:316"
```

According to docs this should work.

Version: master (https://github.com/prometheus-community/windows_exporter/commit/f623c0ed89903413c6fd7cc8ab2d447f3d20ea36)

/cc @breed808

breed808:
Resolves #647

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -273,8 +273,8 @@ func main() {
 		app        = kingpin.New("windows_exporter", "")
 		configFile = app.Flag(
 			"config.file",
-			"YAML configuration file to use. Values set in this file will override flags.",
-		).Default(os.ExpandEnv("$ProgramFiles\\windows_exporter\\config.yml")).String()
+			"YAML configuration file to use. Values set in this file will be overriden by CLI flags.",
+		).String()
 		listenAddress = app.Flag(
 			"telemetry.addr",
 			"host:port for exporter.",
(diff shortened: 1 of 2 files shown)


exporter.go
```diff
@@ -273,8 +273,8 @@ func main() {
 		app        = kingpin.New("windows_exporter", "")
 		configFile = app.Flag(
 			"config.file",
-			"YAML configuration file to use. Values set in this file will override flags.",
-		).Default(os.ExpandEnv("$ProgramFiles\\windows_exporter\\config.yml")).String()
```
carlpett, line 277: There's a note in the readme about the default file that we need to remove as well

breed808: Good spot, I'll fix that up

carlpett (approved):
:+1:

PR #656, opened 2020-11-07
Fix collector flags not being registered

### LINKED ISSUE #648: Collector-specific flags missing
The help section suggests, that the collector-specific flags are not available.
```
usage: windows_exporter [<flags>]

Flags:
  -h, --help                    Show context-sensitive help (also try
                                --help-long and --help-man).
      --config.file="C:\\Program Files\\windows_exporter\\config.yml"  
                                YAML configuration file to use. Values set in
                                this file will override flags.
      --telemetry.addr=":9182"  host:port for exporter.
      --telemetry.path="/metrics"  
                                URL path for surfacing collected metrics.
      --telemetry.max-requests=5  
                                Maximum number of concurrent requests. 0 to
                                disable.
      --collectors.enabled="cpu,cs,logical_disk,net,os,service,system,textfile"  
                                Comma-separated list of collectors to use. Use
                                '[defaults]' as a placeholder for all the
                                collectors enabled by default.
      --collectors.print        If true, print available collectors and exit.
      --scrape.timeout-margin=0.5  
                                Seconds to subtract from the timeout allowed by
                                the client. Tune to allow for overhead or high
                                loads.
      --log.level="info"        Only log messages with the given severity or
                                above. Valid levels: [debug, info, warn, error,
                                fatal]
      --log.format="logger:stderr"  
                                "logger:stdout?json=true"
      --version                 Show application version.
```
For examples:
```
C:\Users\Administrator\go\src\github.com\prometheus-community\windows_exporter>.\windows_exporter.exe --collectors.exchange.list
time="2020-11-02T15:28:34+01:00" level=fatal msg="unknown long flag '--collectors.exchange.list'\n" source="exporter.go:312"
```
The corresponding config file fields are ignored.

Version: master (https://github.com/prometheus-community/windows_exporter/commit/f623c0ed89903413c6fd7cc8ab2d447f3d20ea36)

carlpett:
FIxes #648.
I also tried the approach we initially discussed, passing the `app` to the collector package, by having a `registerFlags(regFn func(app *kingpin.Application))` in the collectors package and having each collector call that as part of `init`. That ended up being a lot less clear in my opinion, however (since flag variable definition and actual instantiation need to be split into the registration function and a var block).

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -270,61 +270,55 @@ func initWbem() {
 
 func main() {
 	var (
-		app        = kingpin.New("windows_exporter", "")
-		configFile = app.Flag(
+		configFile = kingpin.Flag(
 			"config.file",
 			"YAML configuration file to use. Values set in this file will be overriden by CLI flags.",
 		).String()
-		listenAddress = app.Flag(
+		listenAddress = kingpin.Flag(
 			"telemetry.addr",
 			"host:port for exporter.",
 		).Default(":9182").String()
-		metricsPath = app.Flag(
+		metricsPath = kingpin.Flag(
 			"telemetry.path",
 			"URL path for surfacing collected metrics.",
 		).Default("/metrics").String()
-		maxRequests = app.Flag(
+		maxRequests = kingpin.Flag(
 			"telemetry.max-requests",
 			"Maximum number of concurrent requests. 0 to disable.",
 		).Default("5").Int()
-		enabledCollectors = app.Flag(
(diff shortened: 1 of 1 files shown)


breed808 (approved):
Looks good! I was looking into making the `app` a global variable, which worked but was far less neater than this.

carlpett:
:+1: Did some testing and everything looks good. I'll cut a 0.15 from this!

PR #657, opened 2020-11-08
Add SMTP collector

### LINKED ISSUE #654: How can check IIS Server status ?
Dear everyone, 
https://community.spiceworks.com/topic/2036533-how-to-monitor-a-smtp-virtual-server-in-iis-6-0
I have question like this friend. 
How can I have metrics about status of IIS server. So i will get alert if it stopping and restart service again.

breed808:
This implements an IIS SMTP Collector, exposing metrics for each SMTP virtual server configured on server.

As I don't have access to an in-use SMTP server at my work, I've downloaded the Windows Server 2019 evaluation ISO and installed the SMTP server on it. As such, the metrics aren't tested, and I can't confirm if their type (gauge/counter) is correct. This will need to be tested before it is merged.

Resolves #654

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Name     | Description | Enabled by default
 [process](docs/collector.process.md) | Per-process metrics |
 [remote_fx](docs/collector.remote_fx.md) | RemoteFX protocol (RDP) metrics |
 [service](docs/collector.service.md) | Service state metrics | &#10003;
+[smtp](docs/collector.smtp.md) | IIS SMTP Server |
 [system](docs/collector.system.md) | System calls | &#10003;
 [tcp](docs/collector.tcp.md) | TCP connections |
 [time](docs/collector.time.md) | Windows Time Service |
--- a/collector/smtp.go
+++ b/collector/smtp.go
@@ -1,0 +1,692 @@
+// +build windows
+
+package collector
+
+import (
+	"fmt"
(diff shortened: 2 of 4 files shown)


carlpett:
Nice! I don't have any IIS-SMTP servers either (to be honest, I had to dig pretty deep in memory to remember there even was such a functionality in IIS).
The code looks good, so my only question is really if we should either be more specific about what smtp server we are collecting from (ie `iis-smtp` or such), or do it as a sub-collector of the existing iis collector?

breed808:
There's details for and against the current `smtp` name. The IIS SMTP server appears to be the only SMTP server with performance counters exposed by Windows, and therefore the only one to fall under this exporter's scope.
However, the SMTP server is managed through the IIS 6 Management console, and is therefore related to IIS in some manner.

The SMTP Server is installed with the SMTP feature, which pulls in some IIS management components in the IIS role. I think keeping the collector separate would be best, as one can have an SMTP server running, or a webserver, or both.

If we're not able to find testers, perhaps the collector could be merged with documentation stating it is in pre-release or beta state?

carlpett:
I'd have thought it'd collide with Exchange? But if not, let's keep it as-is. I'm also fine with adding it marked as experimental in the docs or similar.

breed808:
OK, I've added an experimental note in the docs.

carlpett (approved):
:+1:

Issue #659, opened 2020-11-11
validation issues in the metric format when checking against promtool


kajanth:
Prometheus version : 2.22.1
Platform : Windows

curl -s http://localhost:9184/metrics | c:\prom\prometheus-2.22.1.windows-amd64\promtool.exe check metrics

process_start_time_seconds counter metrics should have "_total" suffix
windows_container_available counter metrics should have "_total" suffix
windows_container_count non-histogram and non-summary metrics should not have "_count" suffix
windows_container_cpu_usage_seconds_kernelmode counter metrics should have "_total" suffix
windows_container_cpu_usage_seconds_usermode counter metrics should have "_total" suffix
windows_net_packets_outbound_discarded counter metrics should have "_total" suffix
windows_net_packets_outbound_errors counter metrics should have "_total" suffix
windows_net_packets_received_discarded counter metrics should have "_total" suffix
windows_net_packets_received_errors counter metrics should have "_total" suffix
windows_net_packets_received_unknown counter metrics should have "_total" suffix
windows_process_handle_count non-histogram and non-summary metrics should not have "_count" suffix
windows_process_thread_count non-histogram and non-summary metrics should not have "_count" suffix

breed808:
Hi @kajanth, thanks for reporting this. It looks like some of these metrics (particularly the `windows_container` metrics) may be gauges, and need to be changed.
Others appear to be counters that need the suffix corrected. I don't have the ability to confirm the container metrics, but I'll look into the others.

mjtrangoni:
Update for _windows_exporter v0.17.0_ and _prometheus v2.32.1_ for the following collectors enabled
`cache,cpu,logical_disk,net,memory,os,process,system,tcp,time`

```bash
# curl -s http://xxxx:9182/metrics | promtool check metrics
windows_cache_copy_read_hits_total non-counter metrics should not have "_total" suffix
windows_memory_cache_faults_total non-counter metrics should not have "_total" suffix
windows_memory_demand_zero_faults_total non-counter metrics should not have "_total" suffix
windows_memory_page_faults_total non-counter metrics should not have "_total" suffix
windows_memory_pool_nonpaged_allocs_total non-counter metrics should not have "_total" suffix
windows_memory_pool_nonpaged_bytes_total non-counter metrics should not have "_total" suffix
windows_memory_pool_paged_allocs_total non-counter metrics should not have "_total" suffix
[...]

PR #667, opened 2020-11-24
Add missing "_total" suffix to ADFS and Net counters

breed808:
This addresses some of the counters mentioned in #659, and should make a future transition to [OpenMetrics](https://www.robustperception.io/openmetrics-is-released) easier.


carlpett (approved):
LGTM. The `net` change will likely break a lot of dashboards, so we'll need to remember to point it out clearly in the release notes.

PR #669, opened 2020-11-26
Update prometheus client library to v1.8.0

### LINKED ISSUE #668: process_resident_memory_bytes absent from metrics
I was trying to monitor memory used by exporters (https://www.robustperception.io/analysing-prometheus-memory-usage) and i found out that the metric "process_resident_memory_bytes" is absent from the exporter.

It is normally a go_client standard metric (https://github.com/prometheus/client_golang/blob/master/prometheus/process_collector.go#L101), why it is absent from windows_exporter ?
And how to monitor ram used by the exporter at the moment since "go_memstats_alloc_bytes" is not the real memory used by the system for that process.

breed808:
Updating the Prometheus client library from v0.9.2 to the latest version (v1.8.0).
The new library version exposes the `process_resident_memory_bytes` metric, and resolves #668

This is thanks to @carlpett's previous [work in the Prometheus client library](https://github.com/prometheus/client_golang/pull/596).

There's been a number of changes between client library versions, so testing of this change should be undertaken prior to merging or the next release.


carlpett:
Fantastic! I don't exactly recall what held off upgrading at the time the upstream PR was merged, but as I recall it was a fairly obvious issue, so if scrapes work, I believe that specific issue should be fixed at least :)
We should try to build some basic smoke test for CI. We can probably take some inspiration from how node_exporter does it, ie more or less you check that we get the same basic output structure as we expect.

carlpett (approved):
:tada:

breed808:
Agreed, adding a smoke test to the would be rather nice. I'll look into it and see if I can get working example in Appveyor.

PR #671, opened 2020-11-27
Add DFSR collectors

### LINKED ISSUE #666: [FeatureRequest] DFSR monitoring
Hello,

I found this project very useful! Thanks a lot for making this.

I would like to raise my hand and ask for a feature request: DFSR monitoring. Is it possible to include it into roadmap?

Thanks,
Maciej

breed808:
Implements DFSR collectors, though no metric documentation is present yet. Will resolve #666 when merged.
@mjozefcz are you comfortable testing this PR, or would you like me to provide some build instructions?

### DIFF
--- a/collector/dfsr.go
+++ b/collector/dfsr.go
@@ -122,7 +122,7 @@ func NewDFSRCollector() (Collector, error) {
 
 		// Connection
 		ConnectionBandwidthSavingsUsingDFSReplicationTotal: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "bandwidth_savings_using_dfs_replication_total"),
+			prometheus.BuildFQName(Namespace, subsystem, "bandwidth_savings_using_dfs_replication_bytes_total"),
 			"Total amount of bandwidth savings using DFS Replication for this connection, in bytes",
 			[]string{"name"},
 			nil,
@@ -143,14 +143,14 @@ func NewDFSRCollector() (Collector, error) {
 		),
 
 		ConnectionFilesReceivedTotal: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "files_received_total"),
+			prometheus.BuildFQName(Namespace, subsystem, "received_files_total"),
 			"Total number of files receieved for connection",
 			[]string{"name"},
 			nil,
 		),
 
 		ConnectionRDCBytesReceivedTotal: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "rdc_bytes_received_total"),
+			prometheus.BuildFQName(Namespace, subsystem, "rdc_received_bytes_total"),
 			"",
 			[]string{"name"},
 			nil,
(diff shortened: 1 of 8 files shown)


collector/dfsr.go
```diff
@@ -0,0 +1,777 @@
...
+const (
+	dfsrVolumeSubsystem     = "dfsr_volume"
+	dfsrConnectionSubsystem = "dfsr_connection"
+	dfsrFoldersSubsystem    = "dfsr_folder"
+)
+
+func init() {
+	registerCollector(dfsrConnectionSubsystem, NewDFSRConnectionCollector, "DFS Replication Service Connections")
+	registerCollector(dfsrFoldersSubsystem, NewDFSRConnectionCollector, "DFS Replication Service Folders")
+	registerCollector(dfsrVolumeSubsystem, NewDFSRConnectionCollector, "DFS Replication Service Volumes")
```
carlpett, line 19: I'm not sure we want three distinct collectors for this - can't we use the subcollector pattern here?

docs/collector.dfsr.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Flags
+
+None
+
+## Metrics
+
+Name | Description | Type | Labels
+-----|-------------|------|-------
+`dfsr_connection_bandwidth_savings_using_dfs_replication_total` | | counter | None
```
carlpett, line 19: The code seems to have a `name` label, for most metrics, at least?

carlpett (commented):
Nice! I don't have a DFS environment, but the code looks mostly fine. 
Re test binaries, just want to point out that we get them in the CI build, eg https://ci.appveyor.com/project/prometheus-community/windows-exporter/builds/36550282/artifacts

docs/collector.dfsr.md
breed808: My mistake, I'll add the label for each entry.

collector/dfsr.go
breed808: Do we have a subcollector example? Are we thinking a separate namespace I.E. `dfsr.volume` rather than the current `dfsr_volume`?

breed808:
I've reworked the collector to support child collectors, similar to the `mssql` collector.
It's more complex, though I do like the ability to enable/disable child collectors.

It compiles and the unit tests are successful, but this requires proper testingon a server running DFSR.

breed808:
Small change: The collector init() perflib dependencies weren't set correctly, which I've fixed.

collector/dfsr.go
```diff
@@ -0,0 +1,868 @@
...
+		dfsrScrapeSuccessDesc: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "collector_success"),
+			"windows_exporter: Whether a dfsr child collector was successful.",
+			[]string{"collector"},
+			nil,
+		),
+
+		// Connection
+		ConnectionBandwidthSavingsUsingDFSReplicationTotal: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "bandwidth_savings_using_dfs_replication_total"),
```
carlpett, line 134: Bit of a nitpick, but we should have the unit in the name, as a suffix for gauges or just before the _total suffix for counters.
Goes for a lot of the names below as well, unfortunately.

collector/dfsr.go
```diff
@@ -0,0 +1,868 @@
...
+	wg := sync.WaitGroup{}
+
+	for name, function := range c.dfsrChildCollectors {
+		wg.Add(1)
+		go c.execute(ctx, name, function, ch, &wg)
+
+	}
+	wg.Wait()
+
+	if c.dfsrChildCollectorFailure > 0 {
```
carlpett, line 451: This looks like it'll keep returning an error if we ever had one, or am I missing a reset somewhere? Regardless, this will be problematic if we have concurrent scrapes.
Since you use perflib in this collector, I'd expect each child to finish very quickly, and we might not need to fan out? Then you could just make a regular call to each child serially and return an error.

carlpett (commented):
Found a boring naming nitpick, but also think that the subcollector error handling has a race condition :(
Apart from that, looks good! Since this is the second new collector recently which we can't test ourselves, perhaps we should discuss how to deal with beta/experimental collectors in a good way.

collector/dfsr.go
breed808: I think you're correct: there's no reset of this internal counter. This is also the case in the `mssql` collector, so I'll fix that too.
Running the collectors serially would be simpler than keeping the concurrent scrapes and trying to fix with `atomic/sync`, so I'll change that too.

breed808:
I've corrected most of the metric names; the remaining names will need to be tested to confirm the applicable unit (I suspect most are "bytes").

Child collectors have also been modified, and are now run serially. `mssql` collector has had the error counter issue fixed too.

carlpett (approved):
LGTM! Do we want to add something about experimental status in the docs?

breed808:
Yes, I can add a note in the collector documentation.
Do you think a log entry on collector initialisation for this and the `smtp` collector would be suitable? That way, if a user hasn't seen the collector docs there will still be a log entry notifying them of the collector's experimental status.

carlpett:
Yes, that sounds like a nice touch! :+1:

PR #672, opened 2020-12-01
End-to-end testing script

breed808:
Add an end-to-end testing script for use in CI, to compare changes in windows_exporter output.
I've tested this in a local Windows 10 VM, so I'm unsure how Appveyor would handle this.

A downside of this change is that the `e2e-output.txt` file would need to be modified on any metric change to the default exporters.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 export GOOS=windows
 
-build:
+.PHONY: build
+build: windows_exporter.exe
+windows_exporter.exe: **/*.go
 	promu build -v
 
 test:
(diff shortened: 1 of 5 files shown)


carlpett:
Super nice! I think we can just plop the new make target into the `after_test` section in appveyor.yml and it should work

breed808:
You were right, that's worked very well in Appveyor. I can see the `windows_textfile_scrape_error` value has changed, but all else is well. Is this expected behaviour in the Appveyor CI system?
I'll amend the end-to-end script to raise an error if the output differs.

breed808:
Script is now raising an error, just need to discover why there's a textfile scrape error in Appveyor

breed808:
Got it, CI is happy now.

carlpett (approved):
Great stuff :clap:

PR #680, opened 2020-12-10
Add initial HTTPS support for windows_exporter

breed808:
HTTPS support is required for environments with policy and/or regulatory
requirements which require internal traffic to be encrypted.

This has been rather liberally taken from the `node_exporter` project,
with some minor modifications made to support the current
windows_exporter logging library.

Addition of the code and test files will increase maintenance overhead
as we'll need to keep an eye on HTTP changes in `node_exporter`.
Unfortunately due to differences in platforms and logging libraries the
package couldn't be directly imported.


brian-brazil:
Please do not copy&paste or change this code, rather depend on the version in the exporter-toolkit repo. It's explicitly a library so that there's exactly one copy of this security-sensitive code, rather than multiple drifting copies. common/log is also deprecated, and should not be used - so if that's your problem I you recommend a prompt migration as that code is destined to be deleted.

@SuperQ @roidelapluie

breed808:
Thanks Brian, I wasn't aware of the exporter-toolkit repo. I'll rework the PR to include that.
This may be blocked until we can move away from common/log.

roidelapluie:
exporter-toolkit is not related to common, it is explicitly meant for third parties

carlpett:
Re common/log, we should probably just fork the bits we use (or port the Windows Event log stuff I added there to some other log framework, but I'm not sure we care all that much?).

breed808:
I think it's worth trying to port it over; I've found the Event Log entries rather helpful when diagnosing issues with the `windows_exporter` service installed from MSI.

I can see some code [specific to syslog](https://github.com/go-kit/kit/blob/master/log/syslog/syslog.go) in the `go-kit/kit/log` library. Do you think we could do the same with your event log code?

carlpett:
Yeah, we definitely need ot keep the capability. My main question is if we should keep as-is, just moving the existing logrus code over to an existing package, or if we see some benefit of eg go-kit.

breed808:
Closing in favour of @SuperQ's PR.

PR #689, opened 2020-12-30
Document version requirement for time collector

### LINKED ISSUE #658: Time collector fails on Windows Server 2012R2
Getting errors when enabling the [time collector](https://github.com/prometheus-community/windows_exporter/blob/master/docs/collector.time.md) on v0.15.0 

Windows application event errors every 30 seconds:

- Event 102 - collector time failed after 0.000000s: counter not found

and

- Event 102 - failed collecting time metrics:<nil> counter not found

### Testing 

- Errors are repeatable on Windows Server 2012R2 virtual machines
- Errors do not occur on Windows 10 or Windows Server 2016

breed808:
Windows Server 2016 or newer is required for the collector, as the
perflib counters were not exposed in previous versions.

Resolves #658

### DIFF
--- a/collector/time.go
+++ b/collector/time.go
@@ -3,6 +3,8 @@
 package collector
 
 import (
+	"errors"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 )
@@ -22,6 +24,10 @@ type TimeCollector struct {
 }
 
 func newTimeCollector() (Collector, error) {
+	if getWindowsVersion() <= 6.1 {
+		return nil, errors.New("Windows version older than Server 2016 detected. The time collector will not run and should be disabled via CLI flags or configuration file")
+
+	}
 	const subsystem = "time"
 
 	return &TimeCollector{
--- a/docs/collector.time.md
+++ b/docs/collector.time.md
@@ -3,6 +3,8 @@
 The time collector exposes the Windows Time Service metrics. Note that the Windows Time Service must be running, else metric collection will fail.
 If the Windows Time Service is stopped after collection has started, collector metric values will reset to 0.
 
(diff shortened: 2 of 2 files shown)


collector/time.go
```diff
@@ -22,6 +22,9 @@ type TimeCollector struct {
 }
 
 func newTimeCollector() (Collector, error) {
+	if getWindowsVersion() < 10 {
+		log.Warn("Windows version older than Server 2016 detected. The time collector is likely to fail!")
```
carlpett, line 26: I'm considering if we shouldn't fail outright here? We know with high certainty that it will fail, and I'd personally rather have the system reject my settings in that case?

carlpett: Also, a vague recollection about Windows version numbers - does this actually work? IIRC you need to mark applications with a compatibility manifest to get the "real" version number, otherwise it'd lie and report an older version (8.1?).
I'm not 100% sure this applies here though, but I've been bitten by it in the past.

breed808: >I'm considering if we shouldn't fail outright here?

Are we able to fail individual collectors? If so I imagine `log.Error()` would be suitable here?

Regarding the version numbers, I'm unfortunately unable to test this as I don't have a server older than 2016 available to me. Perhaps we could hold off on merging until someone's able to test this?

carlpett: Well, we can return an error, which would exit the exporter. I'd argue it is an invalid configuration to enable this collector on non-compatible machines?

Re version numbers, if we get back something higher than 6.1 on modern systems, we should be fine. Looks like the version masking might only relate to calling `GetVersion[Ex]` in the Windows APIs, while we read from the registry.

breed808: Thanks Calle. I've updated the commit, could you let me know if it's alright?

carlpett (approved):

PR #691, opened 2020-12-30
Rename mssql instance label to mssql_instance

### LINKED ISSUE #690: windows_mssql metrics use 'instance' label like Pormetheus
Hi,

The MSSQL metrics use the label 'instance'.
But, this label is automatically attached when Prometheus scrape a target.
https://prometheus.io/docs/concepts/jobs_instances/#jobs-and-instances

In Grafana, if I do query_result({job="myjob"}), and filter on label 'instance', I have targets names and MSSQL instance in the list.
It's complicated to extract the targets list.

is using 'instance' for MSSQL a good idea ?

breed808:
Instance label is conflicting with default Prometheus "instance" label
which is added when scraping targets.

Resolves #690


carlpett (approved):
Yikes, how did I miss that in the reviews? :open_mouth: Interesting as well that there haven't been any reports, I suspect people aren't setting `honor_labels: true` very often and simply accept the `exported_` prefix Prometheus adds.

Anyway, this is the right thing to do. Not a huge fan of the new label name either, but I dont' have a better suggestion.

Issue #677, opened 2020-12-08
Replace WMI in OS collector


benridley:
We're experiencing performance issues with the OS collector timing out on our production nodes. This seems in line with other collectors having performance issues when using WMI, so I'd like to suggest we replace the OS collector's metrics with compatible metrics from more performant sources. 

I've found [sysinfoapi](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/) which contains much of the same information as the WMI Operating System class. Perhaps this would make a good candidate for replacement?

My colleagues and I are doing some PoCs on this, so I thought I'd create this issue to track our efforts and seek feedback from maintainers.

breed808:
This sounds similar to #649 where sysinfo is replacing WMI for the process collector.
Personally, I'm eager to replace any remaining usage of WMI if we can show the replacement is stable and fast, which isn't difficult as WMI sets such a low bar.

Would you be keen to open a Pull Request for this?

benridley:
Thanks @breed808, 

Yeah absolutely. We will write an implementation and some benchmarks, and come back here with results and PR once we confirm it's all working as intended. 

Cheers!

benridley:
Wow, well I decided to benchmark the changes. My branch is [here](https://github.com/benridley/windows_exporter/tree/dev_cs_collector)

Current benchmark using WMI:
```goos: windows
goarch: amd64
pkg: github.com/prometheus-community/windows_exporter/collector
BenchmarkCsCollect-16    	     110	  10972730 ns/op	    4102 B/op	     152 allocs/op
PASS
ok  	github.com/prometheus-community/windows_exporter/collector	3.267s
```

and new benchmark using sysinfoapi:

```
goos: windows
goarch: amd64
pkg: github.com/prometheus-community/windows_exporter/collector
BenchmarkCsCollect-16    	   28776	     42153 ns/op	   25440 B/op	      36 allocs/op
PASS
ok  	github.com/prometheus-community/windows_exporter/collector	2.704
```

So a speedup of 262x... Not bad! I'll get a PR ready soon.

breed808:
Performance is looking very nice, I've run the benchmarks and have a similar result on a testing Windows 10 VM.
Looking forward to the PR :smiley:

benridley:
PR#702

PR #702, opened 2021-01-20
Replace WMI in cs and os collectors

benridley:
This PR addresses #677 by replacing WMI calls within the above collectors with direct winapi calls, perflib calls, and registry reads. 

A 'header' folder has been introduced that contains Golang wrappers for parts of the Windows API that the exporter requires for exposing metrics. We figured this would be useful going forward to wrap more of the direct winapi calls as new collectors are added and older ones migrated away from WMI. 

Performance benchmarks have shown ~20x improvement on the os collector and ~250x improvement in the cs collector.


breed808 (commented):
This is looking very good. I'm particularly impressed with the documentation in the comments, well done there!

I'm also rather happy with the introduction of benchmarks for the collectors.
I think this is a good start to introducing benchmarks so we can identify laggard collectors. The question is, do we want to introduce a separate `_test.go` file for each collector containing benchmarks or centralize them in `collector_test.go`?

I'd be happy approving this myself, but I'll give @carlpett the opportunity to review & approve; his low-level knowledge of Windows & this exporter surpasses my own.

carlpett:
Thank you @benridley and @retryW, this is a great step forward!
From a high level, I think the main thing I'd like to discuss is if we could/should wrap the details a bit more? The `Collect` functions are now a quite long series of somewhat involved and low-level operations. I'm not completely certain where a good tradeoff lies, so happy to have a discussion there.
For example, we could have the APIs into two layers, one with the one-to-one mapping from Windows structs into Go, and then a a Go-friendly API that is what we actually expose to the rest of the exporter. This would of course mean some amount of more code, but I'm a bit concerned with how complex it looks.
Thoughts?

retryW:
Cheers for the positive feedback!
I'm not opposed to adding another layer, however feel like for the most part it would add extra obfuscation for little benefit.
Maybe we wrap the windows registry calls up into their own function as they're causing the majority of the clutter/readability issues within the os.go `collect()` func ?

benridley:
Thank you both for the feedback!
@breed808 Good question. I think I'm in favour of separate files, only because if we start adding unit tests as well as benchmarks it'd be nice to have unit tests isolated per collector, as a single file could get really messy with multiple tests per collector for all collectors.
@carlpett I definitely see the concern, but I'm not sure how much we can abstract it in a meaningful way. I think a lot of the low level clutter is just a consequence of working with a low-level Windows API, and we shouldn't try and hide that when it's core to how the exporter operates. I am open to the idea if we can find a meaningful abstraction though, but what would that look like? Perhaps commenting might improve readability also without adding a layer of indirection?

carlpett:
Extracting a Registry lookup function sounds good @retryW, I think we might have some other places that could use that.

@benridley Agree, there's definitely a fine line between too much and too little "implementation context". I'd subjectively find it easier to read and reason about if we'd separate the "load the data" and "export the data" concerns, and let the loading part scrub away some of the Win32isms (eg the `Ull` data type prefixes, or the `Wki102` one). As an example, we'd then have a new package `sysinfoapi` wrapping the `headers/sysinfo` package, with a `MemoryStatus` struct with a `TotalPhysicalMemoryBytes` field (and the others needed).
Hiding the mechanisms of this will of course require stepping through a couple of files when understanding the deeper bits, and there will be a bit of boilerplate code writing required.

The main benefit of this would be making the code base a bit more welcoming to new contributors looking to add more functionality in these files, not having to find the right spot to inject more logic (although, writing this I realize there's of course the case to be made that these collectors are probably quite static in features).

What do you think, reasonable?

benridley:
@carlpett Yeah that seems reasonable. We'll start creating some more idiomatic wrappers for the API and see how it looks. Cheers!

benridley:
@carlpett @breed808 Apologies for the delay folks!

Let me know what you think now. Basically I've hidden all the nasty low level WinAPI operations in private methods and structs, and the 'header' files expose idiomatic wrappers that can be used externally.  Certainly makes the code a lot cleaner to use, although it does involve a bit of double handling. 

Let me know what you think and if this is what you had in mind, otherwise we can keep iterating :)

headers/sysinfoapi/sysinfoapi.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+}
+
+// GetComputerName wraps the GetComputerNameW function in a more Go-like way
+// https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getcomputernamew
+func GetComputerName(f WinComputerNameFormat) (string, error) {
+	// 1kb buffer to accept computer name. This should be more than enough as the maximum size
+	// returned is the max length of a DNS name, which this author believes is 253 characters.
+	size := 1024
+	var buffer [4096]uint16
```
carlpett, line 156: 1kb or 8kb? :)
I this as you mention in the comment that 256 (UTF-16) characters should be the limit of the how much can be returned.

collector/os.go
```diff
@@ -146,41 +157,105 @@ type Win32_OperatingSystem struct {
...
+		return nil, err
+	}
+	pagingFiles, _, err := memManKey.GetStringsValue("ExistingPageFiles")
+	if err != nil {
+		return nil, err
+	}
+
+	if err := memManKey.Close(); err != nil {
+		return nil, err
+	}
```
carlpett, line 186: This is a possible leak, since it isn't executed if `GetStringsValue` errors. You'll need to put this in a defer :( 
As an alternative, perhaps pull the openkey/read value/close into a helper?

collector/os.go
```diff
@@ -146,41 +157,105 @@ type Win32_OperatingSystem struct {
...
+
+	bn, _, err := ntKey.GetStringValue("CurrentBuildNumber")
+	if err != nil {
+		return nil, err
+	}
+
+	if err := ntKey.Close(); err != nil {
+		return nil, err
+	}
```
carlpett, line 205: Same as above, needs a defer to avoid possible leak

carlpett (commented):
Sorry for the late review, but this is looking good! A few comments/questions below

headers/netapi32/netapi32.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// NetApiBufferFree frees the memory other network management functions use internally to return information.
+// https://docs.microsoft.com/en-us/windows/win32/api/lmapibuf/nf-lmapibuf-netapibufferfree
+func netApiBufferFree(buffer *wKSTAInfo102) {
+	procNetApiBufferFree.Call(uintptr(unsafe.Pointer(buffer)))
+}
+
+// NetWkstaGetInfo returns information about the configuration of a workstation.
+// WARNING: The caller must call netApiBufferFree to free the memory allocated by this function.
```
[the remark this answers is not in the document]
benridley: Yeah, it was more relevant when the function was public. Now that it's not exposed, this should be irrelevant. I've cleaned it up.

headers/sysinfoapi/sysinfoapi.go
benridley: My bad, I've made size a const 1024 and used that throughout the function. Much clearer now.

collector/os.go
benridley: Good catch... I got a bit tied up trying to figure out we would handle an error in Close(), because if we defer it will be ignored. They do this in the documentation for the package however, and from reading the winapi stuff it seems it will only error if we try and close a key that's not open/already closed. So I think in our case we can safely ignore it and just defer. Something to be wary of though.

carlpett (approved):
Looks *great* to me! Thanks a lot!

collector/os.go
benridley: Thanks @carlpett, I did find the named returns workaround in my travels but I was cautious because I thought we could end up in a situation where we e.g. fail to open a key, then fail to close it, and the close error overwrites the open failure which would lead to confusion. Directly logging the error seems like a great workaround, so I'll keep that in mind. Cheers!

PR #706, opened 2021-01-26
Remove "total" suffix from dns_memory_used_bytes

### LINKED ISSUE #284: DNS collector memory metric should not be a counter
The DNS collector reports a metric `wmi_dns_memory_used_bytes_total` as a counter, but it should actually be a gauge, and not have the `_total` suffix.

breed808:
Metric is a gauge and therefore should not have the "total" suffix.

Resolves #284


carlpett (approved):
LGTM!

PR #710, opened 2021-01-29
Add missing dfsr metric descriptions

### LINKED ISSUE #707: dfsr collector returned error
Hello, 

After build windows_exporter.exe from the source code and run it with dfsr collector enabled, it returned below error  from the metric output:
----quote for the error-------

3 error(s) occurred:
* collected metric windows_dfsr_received_files_total label:<name:"name" value:"dfsr-test-{C9BCA8F8-76DB-4B79-BF13-6AD6D5CF8C31}" > counter:<value:1 >  has help "" but should have "Total number of files receieved for connection"
* collected metric windows_dfsr_rdc_received_files_total label:<name:"name" value:"dfsr-test-{C9BCA8F8-76DB-4B79-BF13-6AD6D5CF8C31}" > counter:<value:0 >  has help "" but should have "Total number of Remote Differential Compression files received"
* collected metric windows_dfsr_files_received_bytes_total label:<name:"name" value:"dfsr-test-{C9BCA8F8-76DB-4B79-BF13-6AD6D5CF8C31}" > counter:<value:435 >  has help "" but should have "Total size of files received, in bytes"
-----unquote for the error---------------------------
Could you help to check what's the issue?

breed808:
Reviewing these metrics has made me realise how sloppy my initial implementation was :disappointed:

@bds006 this should resolve your error with the empty metric descriptions, but keep in mind these metrics are still untested so there may be other bugs for you to discover.

Resolves #707

### DIFF
--- a/collector/dfsr.go
+++ b/collector/dfsr.go
@@ -23,10 +23,6 @@ func init() {
 
 // DFSRCollector contains the metric and state data of the DFSR collectors.
 type DFSRCollector struct {
-	// Meta
-	dfsrScrapeDurationDesc *prometheus.Desc
-	dfsrScrapeSuccessDesc  *prometheus.Desc
-
 	// Connection source
 	ConnectionBandwidthSavingsUsingDFSReplicationTotal *prometheus.Desc
 	ConnectionBytesReceivedTotal                       *prometheus.Desc
@@ -108,20 +104,6 @@ func NewDFSRCollector() (Collector, error) {
 	addPerfCounterDependencies(subsystem, perfCounters)
 
 	dfsrCollector := DFSRCollector{
-		// meta
-		dfsrScrapeDurationDesc: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "collector_duration_seconds"),
-			"windows_exporter: Duration of an dfsr child collection.",
-			[]string{"collector"},
-			nil,
-		),
-		dfsrScrapeSuccessDesc: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "collector_success"),
-			"windows_exporter: Whether a dfsr child collector was successful.",
-			[]string{"collector"},
-			nil,
-		),
-
 		// Connection
 		ConnectionBandwidthSavingsUsingDFSReplicationTotal: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "bandwidth_savings_using_dfs_replication_bytes_total"),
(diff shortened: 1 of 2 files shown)


collector/dfsr.go
```diff
@@ -124,8 +124,8 @@ func NewDFSRCollector() (Collector, error) {
 
 		// Connection
 		ConnectionBandwidthSavingsUsingDFSReplicationTotal: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "bandwidth_savings_using_dfs_replication_bytes_total"),
-			"Total amount of bandwidth savings using DFS Replication for this connection, in bytes",
+			prometheus.BuildFQName(Namespace, subsystem, "bandwidth_savings_using_dfs_replication_percentage"),
```
carlpett, line 127: I have some vague memory of this being changed *to* bytes from percentages in the original PR? Exposing the bytes if possible would be more useful, IMO (and most frequently, that is what we actually get in the data?)

collector/dfsr.go
```diff
@@ -153,21 +153,21 @@ func NewDFSRCollector() (Collector, error) {
...
-			"",
+			"Total bytes received on the connection while replicating files using Remote Differential Compression",
 			[]string{"name"},
 			nil,
 		),
 
 		ConnectionRDCCompressedSizeOfFilesReceivedTotal: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "rdc_compressed_size_of_files_received_total"),
-			"",
+			prometheus.BuildFQName(Namespace, subsystem, "rdc_compressed_size_of_received_files_total"),
```
carlpett, line 162: This would be `_bytes_total`, right?

collector/dfsr.go
```diff
@@ -188,57 +188,57 @@ func NewDFSRCollector() (Collector, error) {
 
 		// Folder
 		FolderBandwidthSavingsUsingDFSReplicationTotal: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "bandwidth_savings_using_dfs_replication_total"),
-			"",
+			prometheus.BuildFQName(Namespace, subsystem, "bandwidth_savings_using_dfs_replication_percentage"),
```
carlpett, line 191: Same percentage/bytes question as above here

carlpett (commented):
Some small nits, otherwise LGTM! :+1:

collector/dfsr.go
breed808: Good spot, I've amended that metric.

collector/dfsr.go
breed808: I think you're right: I've reverted the change to these two bandwidth metrics.

carlpett (approved):
:tada:

PR #716, opened 2021-02-02
Set container collector prefix based on owner property

jsturtevant:
This updates the `container_id` label for the metrics reported by the container [collector](https://github.com/prometheus-community/windows_exporter/blob/master/docs/collector.container.md) to use the owner ref which is set by docker and containerd respectively.

When running the windows_exporter with the `container` collector running with a containerd runtime the output of the metrics prefixes the metrics `container_id` label with `docker://` but it should be `containerd://` to match other tooling.

Current output:
```
windows_container_available{container_id="docker://130d82553c53d017649b7115a4e58a8007eef0641ce8804bfb6eb7dcc70ef18b", instance="10.240.0.65:5000", job="windows-exporter"}
```

If you use hcsdiag list on the host to see the owner ref it specifies the containerd shim as the owner:

```
hcsdiag list
130d82553c53d017649b7115a4e58a8007eef0641ce8804bfb6eb7dcc70ef18b
    Windows Server Container,   Running,                                     , containerd-shim-runhcs-v1.exe
```

This is important for jobs that match against `container_id` to gather more stats from tools like `kube-state-metrics` which specifies `containerd` for the `contianerid`:

```
kube_pod_container_info{container="iis", container_id="containerd://130d82553c53d017649b7115a4e58a8007eef0641ce8804bfb6eb7dcc70ef18b", image="mcr.microsoft.com/windows/servercore/iis:windowsservercore-ltsc2019", image_id="mcr.microsoft.com/windows/servercore/iis@sha256:4763a2b3029259a9fd05168d3963493cbf245b535e02dc80e1eaf1dc51645a05", instance="10.240.0.39:8443", job="kube-state-metrics", namespace="default", pod="iis-1809-668c5b9c87-t5z66"}
```

For example, a promethues rule from the [kubernetes-mixin project](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/3d4f68cead695d2717d606f80246ffcfd7b1f08b/rules/windows.libsonnet#L183):

```
windows_container_available{%(wmiExporterSelector)s} * on(container_id) group_left(container, pod, namespace) max(kube_pod_container_info{%(kubeStateMetricsSelector)s}) by(container, container_id, pod, namespace)
            ||| % $._config,
```

### DIFF
--- a/collector/container.go
+++ b/collector/container.go
@@ -168,70 +168,67 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
 	}
 
 	for _, containerDetails := range containers {
-		containerId := containerDetails.ID
-
-		container, err := hcsshim.OpenContainer(containerId)
+		container, err := hcsshim.OpenContainer(containerDetails.ID)
 		if container != nil {
 			defer containerClose(container)
 		}
 		if err != nil {
-			log.Error("err in opening container: ", containerId, err)
+			log.Error("err in opening container: ", containerDetails.ID, err)
 			continue
 		}
 
 		cstats, err := container.Statistics()
 		if err != nil {
-			log.Error("err in fetching container Statistics: ", containerId, err)
+			log.Error("err in fetching container Statistics: ", containerDetails.ID, err)
 			continue
 		}
-		// HCS V1 is for docker runtime. Add the docker:// prefix on container_id
-		containerId = "docker://" + containerId
+		containerIdWithPrefix := getContainerIdWithPrefix(containerDetails)
 
 		ch <- prometheus.MustNewConstMetric(
 			c.ContainerAvailable,
 			prometheus.CounterValue,
 			1,
-			containerId,
+			containerIdWithPrefix,
 		)
 		ch <- prometheus.MustNewConstMetric(
 			c.UsageCommitBytes,
 			prometheus.GaugeValue,
 			float64(cstats.Memory.UsageCommitBytes),
-			containerId,
+			containerIdWithPrefix,
 		)
 		ch <- prometheus.MustNewConstMetric(
 			c.UsageCommitPeakBytes,
 			prometheus.GaugeValue,
 			float64(cstats.Memory.UsageCommitPeakBytes),
-			containerId,
+			containerIdWithPrefix,
 		)
 		ch <- prometheus.MustNewConstMetric(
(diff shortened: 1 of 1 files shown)


collector/container.go
```diff
@@ -184,8 +184,7 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
 			log.Error("err in fetching container Statistics: ", containerId, err)
 			continue
 		}
-		// HCS V1 is for docker runtime. Add the docker:// prefix on container_id
-		containerId = "docker://" + containerId
+		containerId = getContainerIdWithPrefix(containerDetails)
```
breed808, line 187: This wasn't introduced by your change, but I think we should avoid mutating the state of the `containerId` variable.
Perhaps a `containerPrefix := getContainerIdWithPrefix(containerDetails)` could be used instead?

collector/container.go
```diff
@@ -280,3 +279,15 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
...
+func getContainerIdWithPrefix(containerDetails hcsshim.ContainerProperties) string {
+	switch containerDetails.Owner {
+	case "docker":
+		return "docker://" + containerDetails.ID
+	case "containerd-shim-runhcs-v1.exe":
+		return "containerd://" + containerDetails.ID
+	default:
+		// If not set fall back prior behavior
+		return "docker://" + containerDetails.ID
+	}
```
breed808, line 292: Minor nitpick, but should the `"docker"` case exist if it's already covered by the `default` case?

breed808 (commented):
Thanks for submitting this James!
I've included a couple minor comments, have a look and let me know if there's any issues.

collector/container.go
jsturtevant: I had the same thought :smile:   Will make this update.

collector/container.go
jsturtevant: I was going for being overly explicit in the owners,  Happy to go either way.

collector/container.go
jsturtevant: Updated

collector/container.go
breed808: Lets omit the `"docker"` case for now, we can add it in the future if required.

breed808 (approved):
Looks good!

PR #717, opened 2021-02-02
Use correct metric name for windows_net_packets_outbound_discarded

jsturtevant:
The docs call the metric `windows_net_packets_outbound_discarded` but it is exported as `windows_net_packets_outbound_discarded_total`

https://github.com/prometheus-community/windows_exporter/blob/b9b60f1ea0e6c66314761f74b006814241cbe2ba/collector/net.go#L72-L77


breed808 (approved):
Thanks for fixing this one :thumbsup:

PR #723, opened 2021-02-12
Cache collector

breed808:
Implements collector for Windows Cache. This is using the commits provided by @wbagdon provided in #455.
Due to the age of the PR, I've rebased the commits from current master and pushed to a new branch.

### DIFF
--- a/collector/cache.go
+++ b/collector/cache.go
@@ -1,103 +1,104 @@
+// +build windows
+
 package collector
 
 import (
-	"github.com/StackExchange/wmi"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 )
 
 func init() {
-	Factories["cache"] = NewCacheCollector
+	registerCollector("cache", newCacheCollector, "Cache")
 }
 
-// A CacheCollector is a Prometheus collector for WMI Win32_PerfFormattedData_PerfOS_Cache  metrics
+// A CacheCollector is a Prometheus collector for Perflib Cache metrics
 type CacheCollector struct {
-	AsyncCopyReadsPersec         *prometheus.Desc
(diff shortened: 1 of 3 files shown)


collector/cache.go
```diff
@@ -0,0 +1,452 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+)
+
+func init() {
+	registerCollector("cache", newCacheCollector, "Cache")
+}
+
+// A CacheCollector is a Prometheus collector for WMI Win32_PerfFormattedData_PerfOS_Cache  metrics
```
carlpett, line 14: Old :)

collector/cache.go
```diff
@@ -0,0 +1,452 @@
...
+func (c *CacheCollector) Collect(ctx *ScrapeContext, ch chan<- prometheus.Metric) error {
+	if desc, err := c.collect(ctx, ch); err != nil {
+		log.Error("failed collecting cache metrics:", desc, err)
+		return err
+	}
+	return nil
+}
+
+// Win32_PerfFormattedData_PerfOS_Cache  docs:
+// - https://docs.microsoft.com/en-us/previous-versions/aa394267(v=vs.85)
```
carlpett, line 238: Do we have a better link?

carlpett (approved):
Nice work! Looks like it addresses all the comments I had on the draft PR as well.
There are two minor nits on comments, but LGTM! :tada:

collector/cache.go
breed808: Good spot, I've amended the commit to fix this.

collector/cache.go
breed808: Unfortunately no. This is the only official documentation for the class I've been able to find, so @wbagdon had it right when using this link for documentation.

PR #726, opened 2021-02-16
Add a cpu_info collector

higels:
The cpu_info collector exposes a single metric per socket with some information on installed CPUs, somewhat like the node_cpu_info metric from node-exporter.


carlpett (approved):
Neat, LGTM! @breed808 any thoughts?

breed808:
Initially I was thinking this could be merged into the `cpu` collector, but keeping it separate is probably best due to the difference in metrics sources (Perflib for `cpu` vs WMI for `cpu_info`).

Minor nitpick, but the WMI source should be mentioned in the documentation. Besides that it's looking good :+1:

higels:
Hopefully the last commit addressed your nitpick @breed808 - I may have misunderstood.

breed808:
Thanks @higels, that's what I was looking for.

Issue #650, opened 2020-11-03
cpu collector dose not support fetching cup temperature metrics


zhongxinlei:
I was using windows exporter to collect monitoring data to find out that it dose not support collecting cup temperature information. Since cpu temperature is an important element to evaluate and monitor the system, and some other exporters(like node_exporter) already included this feature to collect cpu temperature, it might be nice and necessary for windows exporter to also have this feature?

carlpett:
Hi @zhongxinlei,
We expose temperature (on supported hardware, not every machine exposes it to Windows in a way such that we can read it) via the `thermalzone` collector.

Issue #560, opened 2020-06-27
thermalzone does not seem to work properly for me


Jared87:
I installed the exporter because my CPU fan was really bad in cooling, i activated the thermalzone plugin, but it shows constantly `16.850000000000023` in the `windows_thermalzone_temperature_celsius` field.
```
# HELP windows_thermalzone_percent_passive_limit (PercentPassiveLimit)
# TYPE windows_thermalzone_percent_passive_limit gauge
windows_thermalzone_percent_passive_limit{name="\\_TZ.TZ10"} 100
windows_thermalzone_percent_passive_limit{name="\\_TZ.X570"} 100
# HELP windows_thermalzone_temperature_celsius (Temperature)
# TYPE windows_thermalzone_temperature_celsius gauge
windows_thermalzone_temperature_celsius{name="\\_TZ.TZ10"} 16.850000000000023
windows_thermalzone_temperature_celsius{name="\\_TZ.X570"} 16.850000000000023
# HELP windows_thermalzone_throttle_reasons (ThrottleReasons)
# TYPE windows_thermalzone_throttle_reasons gauge
windows_thermalzone_throttle_reasons{name="\\_TZ.TZ10"} 1
windows_thermalzone_throttle_reasons{name="\\_TZ.X570"} 1
```

The system is running a Ryzen 7 3700X on a GIGABYTE X570 AORUS ELITE, other Tools like CoreTemp showing the correct temperatures.

Did i do something wrong or am i completly on the wrong way?

carlpett:
Hi @Jared87,
The driver support for WMI (which is how we're reading the temperatures) is quite mixed, and we've seen many questions on lack of data, or like your case, incorrect data. The only advice we can give unfortunately is checking with your vendor if they have better drivers available.
I haven't used CoreTemp, but I would expect such tools to use some other mechanism to read the data. We'd be very open to help improving the collector if that mechanism is possible to use for us.

Jared87:
Thank you for the explanation. They habe a super short description how they get  the temperatures on their website, but nothing technical. Changing drivers seems Not to ne an option for me, it was just a try to geht the numbers from my desktop in a crashsafe way.

Issue #539, opened 2020-06-05
no temperature metrics exposed from wmi_exporter when enabled thermalzone in Windows Server 2016


godocean:
I downloaded wmi_exporter-0.9.0-amd64.exe in my server which is Windows Server 2016, and started it as below:
./wmi_exporter-0.9.0-amd64.exe --collectors.enabled "thermalzone"
when I request from http://localhost:9182/metrics, it didn't return any temperature metrics.

And then I open powershell with administrator privilege in this server, and run command as below:
![image](https://user-images.githubusercontent.com/17870432/83869994-0ae99280-a760-11ea-81dd-b987a6f936d9.png)

So my question is if "thermalzone" is supported in Windows Server 2016? or should I do some work to open the feature gate in Server 2016? can anyone give some feedback? Thanks ahead.

carlpett:
Hi @godocean,
First off, I'd suggest running the latest version (0.13) rather than 0.9. Regarding the thermalzone collector, is is dependent on both compatible hardware and drivers being installed. As long as that is in place, the metrics will work.

godocean:
> Hi @godocean,
> First off, I'd suggest running the latest version (0.13) rather than 0.9. Regarding the thermalzone collector, is is dependent on both compatible hardware and drivers being installed. As long as that is in place, the metrics will work.

godocean:
@carlpett 
Thanks for your reply!

I've followed your advice and tried latest version(0.13) in Windows Server 2016, but it returned the same result without any temperature metrics,

You mentioned that for thermalzone collector, "it is dependent on both compatible hardware and drivers being installed", what kind of compatible hardware and drivers should be installed? Could you give me more details about it? 

Because I've tried this windows exporter in my PC whose OS is Windows 10 Enterprise, and it worked out temperature metrics. So I'm little confused about it, maybe there should be some feature gate is open by default in Windows 10 Enterprise but not in Windows Server 2016, I'm looking forward to your kindly guidance. Thanks ahead!

carlpett:
Hi,
>You mentioned that for thermalzone collector, "it is dependent on both compatible hardware and drivers being installed", what kind of compatible hardware and drivers should be installed? Could you give me more details about it?

The temperature needs to be measured by the hardware, and there needs to be drivers installed to interpret it. This will be very dependent on your system setup. And, just to be very clear, if the server is a virtual machine, there are no temperature metrics (this is the most common case where it doesn't work).

[...]

Issue #495, opened 2020-03-31
wmi_thermalzone_temperature_celsius get -273.15


CasparChou:
![image](https://user-images.githubusercontent.com/10682620/78058838-78252400-73bb-11ea-90e8-e51729ec85a0.png)

issue as title

[V] Powershell check
    powershell > Get-WMIObject Win32_PerfRawData_Counters_ThermalZoneInformation
    (return Temperature: 301)

[V] reinstall service
[V] restart service
[V] Administrator permission

carlpett:
Hey @CasparChou,
The field we use is `HighPrecisionTemperature`, rather than `Temperature`. Do you get a value there?
In general, the thermalzone collector is dependent on good drivers to report correctly, so absent that there will be issues (although we might want to thinking about if we should just return an error if this field is not initialized, rather than reporting a very cold temperature...)

CasparChou:
it seems no field named `HighPrecisionTemperature`, so it might be sensor not support or driver issues..

carlpett:
Yes, if you don't get that field, it is not supported by the hardware/driver. We should fail instead of reporting -273.15, though.

zhongxinlei:
Hey @carlpett , about the windows_thermalzone_temperature_celsius metrics, in description it says "Temperature of the thermal zone, in degrees Celsius.", could I understand the thermal zone is cpu or ti's something else? and I also could only get value of -237.15, may I ask what specific hardware/driver do I need to have it report right?

carlpett:
Your hardware may support many thermal zones, which may or may not include the CPU. I've seen systems with temperature readings from memory and hard drives, for example.
It is very hard to give recommendations on what may be missing when the temperature is not reported, it depends on so very many factors. I would advice you to contact your hardware manufacturer.

PR #727, opened 2021-02-16
Add coretemp collector

martinlindhe:
Hello!
I have been digging around trying to get proper CPU temperature readings.

Here's what I learned:
- thermalzone collector mostly does not export anything. No driver seems to be available by default and where is even one for Intel CPU:s?
- Intel and AMD cpu:s report core temperatures in different ways. Implementations exist but most is closed source. I did find some information about how to read this, for example openhardwaremonitor (C#) has some code here: https://github.com/openhardwaremonitor/openhardwaremonitor/tree/master/Hardware/CPU
- Reading CPU temperatures directly requires program to run as Administrator, complicating setup
- Reading CPU temperatures directly requires writing a bunch of code I'm not committed enough to put more time into.

Eventually I figured out that the software Core Temp (https://www.alcpu.com/CoreTemp/) has an API of sorts, and instead built a collector directly using it.
This has the following advantages:
- no administrator required
- CoreTemp reports temperature & core load data in a uniform way for both Intel and AMD cpu:s

And the following disadvantages:
- CoreTemp must also be installed and running on the machine to monitor
- No direct golang implementation of temperature scraping
- Coretemp is not free for commercial use

And probably more I didn't think about.

Anyway, this is sort of a hack but works well and exports cpu data such as this:

```
# HELP windows_coretemp_load (Load)
# TYPE windows_coretemp_load gauge
windows_coretemp_load{core="0",name="Intel Core i7 6700K (Skylake)"} 4
windows_coretemp_load{core="1",name="Intel Core i7 6700K (Skylake)"} 3
windows_coretemp_load{core="2",name="Intel Core i7 6700K (Skylake)"} 5
windows_coretemp_load{core="3",name="Intel Core i7 6700K (Skylake)"} 2
# HELP windows_coretemp_temperature_celsius (Temperature)
# TYPE windows_coretemp_temperature_celsius gauge
windows_coretemp_temperature_celsius{core="0",name="Intel Core i7 6700K (Skylake)"} 36
windows_coretemp_temperature_celsius{core="1",name="Intel Core i7 6700K (Skylake)"} 39
windows_coretemp_temperature_celsius{core="2",name="Intel Core i7 6700K (Skylake)"} 35
windows_coretemp_temperature_celsius{core="3",name="Intel Core i7 6700K (Skylake)"} 35
```

So lets have a discussion about if this is a reasonable approach at all.

I also see the lack of proper temperature reporting has been discussed a number of times already: #495, #560, #650, #539

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ Name     | Description | Enabled by default
 [cpu](docs/collector.cpu.md) | CPU usage | &#10003;
 [cs](docs/collector.cs.md) | "Computer System" metrics (system properties, num cpus/total memory) | &#10003;
 [container](docs/collector.container.md) | Container metrics |
+[coretemp](docs/collector.coretemp.md) | Core Temp metrics |
 [dfsr](docs/collector.dfsr.md) | DFSR metrics |
 [dhcp](docs/collector.dhcp.md) | DHCP Server |
 [dns](docs/collector.dns.md) | DNS Server |
@@ -75,7 +76,7 @@ Flag     | Description | Default value
 `--telemetry.path` | URL path for surfacing collected metrics. | `/metrics`
 `--telemetry.max-requests` | Maximum number of concurrent requests. 0 to disable. | `5`
 `--collectors.enabled` | Comma-separated list of collectors to use. Use `[defaults]` as a placeholder for all the collectors enabled by default." | `[defaults]`
-`--collectors.print` | If true, print available collectors and exit. | 
+`--collectors.print` | If true, print available collectors and exit. |
 `--scrape.timeout-margin` | Seconds to subtract from the timeout allowed by the client. Tune to allow for overhead or high loads. | `0.5`
 
 ## Installation
--- a/collector/coretemp.go
+++ b/collector/coretemp.go
@@ -1,0 +1,154 @@
+package collector
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+	"syscall"
+	"unsafe"
+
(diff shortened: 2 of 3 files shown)


carlpett:
Wow, it'd be fantastic to have a working solution for this! But I'm split. On the one hand, there's some precedence in the node_exporter for exposing details about other programs (eg ntp, supervisord), and to some extent this exporter as well (vmware collector is dependent on the VMware guest agent being installed as a WMI provider).
On the other hand, it seems like it is somehow a bit different, and we'd be somewhat committed to maintaining a 3rd party integration.

So not sure, really. Would this be something people would typically be installing on their server fleet?

martinlindhe:
> So not sure, really. Would this be something people would typically be installing on their server fleet?

I doubt it. It's just a quick and dirty solution for the time being.
I guess properly porting the cpu monitoring code from openhardwaremonitor is a more feasible approach.

JackGarbiec:
Just to clarify, do we need to run the actual coretemp binary or download the sdk?

JackGarbiec:
windows_exporter is licensed under MIT, Core Temp seems to only be free for personal use, it needs a commercial license for commercial purposes.

martinlindhe:
> Just to clarify, do we need to run the actual coretemp binary or download the sdk?

Just need the coretemp binary.

martinlindhe:
> windows_exporter is licensed under MIT, Core Temp seems to only be free for personal use, it needs a commercial license for commercial purposes.

Good point, updated original post

PR #728, opened 2021-02-18
Typo in tcp doc

basroovers:
I corrected a metric from counter to gauge.

You can see on line 129 in /collector/tcp.go that it should indeed be a gauge.


carlpett:
Thanks! The change itself looks good, but we need a DCO (click the "Details" link in the checks), but there also seems to be something odd about the PR including a merge of the very first pull request to the repo? This appears to break git history in some interesting way which breaks some tooling... Can you rid of that commit somehow?

PR #730, opened 2021-02-19
Move experimental warnings from init()

breed808:
`init()` causes warnings to be emitted despite collectors not being enabled.


carlpett (approved):
:+1:

PR #736, opened 2021-02-24
Typo in tcp doc

basroovers:
I corrected a metric from counter to gauge.

You can see on line 129 in /collector/tcp.go that it should indeed be a gauge.


breed808 (approved):
Looks good!

PR #744, opened 2021-03-07
Add benchmark for each collector

breed808:
Benchmarks will allow for easier identification of slow collectors.
Additionally, they increase test coverage of the collectors, with some
collectors now reaching 80-95% coverage with this change.

Collector benchmarks have been structed so that common functionality is
present in `collector/collector_test.go` as is done with non-test
functionality in `collector/collector.go`.
Test logic that is specific to individual collectors is present in the
collector test file (E.G. `collector/process_test.go` for the Process
collector).

Signed-off-by: Ben Reedy <breed808@breed808.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ windows_exporter.exe: **/*.go
 test:
 	go test -v ./...
 
+bench:
+	go test -v -bench='benchmark(cpu|logicaldisk|logon|memory|net|process|service|system|tcp|time)collector' ./...
+
 lint:
 	golangci-lint -c .golangci.yaml run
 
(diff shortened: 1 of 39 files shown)


carlpett:
Looks good from a code point of view, just wondering if we would intend to run these as part of CI or more ad-hoc?

breed808:
My intention was for these to be run in an ad-hoc manner, primarily to detect slow collectors.

We could integrate these into the CI with [cob](https://github.com/knqyf263/cob). If you're interested I can get that working too.

carlpett:
I think having them available for ad-hoc usage is sufficient for now. The build agents don't have all the services installed anyway.

carlpett (approved):

breed808:
FYI I've rebased this to resolve the merge conflicts introduced by #702. There's no functional difference in the change so I'll proceed with the merge.

PR #746, opened 2021-03-10
Soft interface

mattdurham:


carlpett:
Hey @mattdurham,
It's great that you are experimenting, but for such large changes as you are proposing, it is wise to open an issue to discuss first. In fact, this idea, making the exporter also available as a library, has been suggested before, and we've been clear that it is something we're considering a non-goal.
Our focus is ensuring the default use-case of the exporter as a binary works well, and we want to avoid spreading ourselves thin supporting other cases.

(You haven't mentioned your specific need, but I assume you want to embed this into another application, and I'll also note that the Prometheus community in general advices against "mega-exporters" merging many different ones into a single binary, for reliability reasons)

rfratto:
Hey @carlpett, this is part of our work building the [Grafana Agent](https://github.com/grafana/agent), which bundles in popular exporters as integrations. We recognize that this is typically a Prometheus anti-pattern but decided to add it as a feature anyway: it's been a very commonly requested feature from Grafana Cloud users to install one binary for collecting and shipping common metrics. 

We've been doing these types changes to other exporters over the last year, see prometheus/consul_exporter#198  ncabatoff/process-exporter#179 for two that I've done. 

It's completely understandable that upstream maintainers aren't always going to want to accept the changes we make. Our goal is to try to contribute everything we can upstream, but otherwise maintain forks that synchronize with the main repo. It's totally fine that you'd prefer to keep this as a non-goal upstream. In cases like these, we'll continue to engage upstream for all other enhancements and bugfixes we might have, but keep maintaining this specific change in our fork. 

To avoid confusion, we'll do our best to make it clear in the Grafana Agent's docs that our integration is _based_ on the windows_exporter but diverges from the project's goals to make it work as a library.

RichiH:
Not commenting on the merit of the PR, just giving context:

From Prometheus-team point of view, we're trying to carefully allow re-use of various components. We're seeing a new wave of adoption which will likely require more parts to be re-usable. e.g. in the context of OpenTelemetry potentially adopting the Prometheus Agent.

This is strictly in the realm of "Prometheus ecosystem", not "Prometheus reference architecture", though.

carlpett:
@rfratto I see, thanks for the explanation! To some extent, this information would change my stance a bit. In previous questions, it has been for some company-internal product where it'd be very unlikely to receive ongoing maintenance.
If you'd be okay with "committing" to keeping this working, I think we might be more likely to be able to find a way to do this. Not least because no matter how clear the wording in documentation, there will likely be significant overspill in issues, so in this respect it'd probably be easier for everyone if there's a single implementation.

@RichiH I think it'd be beneficial to publish this change in stance a bit more widely, along with reasoning describing the background and when to/when not to consider merged exporters.

So, in summary - I'd rather collaborate on this rather than having a forked-and-slightly-different-version on your side. I don't know how much you've progressed during the two weeks (day job has been a bit too hectic lately), but perhaps a PR isn't the best way to work on this? Any suggestions?

rfratto:
We're more than happy to collaborate! We're eager to keep both sides of the project (windows_exporter itself and Grafana Agent's usage of it) working properly, and either @mattdurham or I can help out with maintenance. 

We've progressed a bit over the last two weeks, mainly around minimizing the blast radius and keeping the changes simple and avoiding any code that windows_exporter wouldn't use itself. It'll probably still be a non-trivial PR, so I agree that it's not the best route for collaboration. What's the best way we can reach you offline? Are you on the CNCF Slack?

RichiH:
@carlpett: There's no design doc or anything yet, @SuperQ plans to work on that in a month or two.

The intended use case is one of pipelines and agents.

As to "when" the answer is likely to go along the lines of "Never in the default case. Only if you consider Prometheus a detail of your operations and/or gain significant benefit from deploying a single binary"

RichiH:
https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#heading=h.lhssx3myb0gz and the recording at https://youtu.be/F3e8j42BKV have a discussion covering both intended goals and technical approach.

PR #753, opened 2021-03-24
Update CI to install tools with go install rather than go get

carlpett:


breed808 (approved):
Looks good :+1:
Not sure what's going on with the CI, is that unrelated to the change?

carlpett:
AppVeyor seems to have bumped to Go 1.16, where non-modules mode is no longer supported. And this meant that how we installed promu broke. 
Apparently my first attempted patch ran into Powershell interpreting stderr as actual errors. Trying again with running in cmd instead...

carlpett:
Looks better now. The fact that the environment can change under our feet isn't great though.

PR #759, opened 2021-04-18
Fix textfile crashes with duplicate metrics

### LINKED ISSUE #755: Invalid textfile collector file brings down whole exporter
Certain malformed .prom files are handled gracefully by the exporter
However, there are some cases that will bring down the whole exporter endpoint (HTTP 500) and show an exception.

**Example:**
Add this as `oh-oh.prom` to the textfile collector directory:

```text
# HELP windows_sometest This is a Test.
# TYPE windows_sometest gauge
windows_sometest{display_name="foobar",display_version="13.4.0"} 1
windows_sometest{display_name="foobar",display_version="13.4.0"} 1
```
This brings down the whole exporter:
```powershell
❯ curl.exe http://localhost:9182/metrics --verbose
HTTP/1.1 500 Internal Server Error
collected metric "windows_sometest" { label:<name:"display_name" value:"foobar" > label:<name:"display_version" value:"13.4.0" > gauge:<value:1 > } was collected before with the same name and label values
```

There are several possible cases, where duplicate/mixed format metrics cause this error. This can happen quite easily when using scripting and `textfile` collector.

**Can this be fixed?**

Expected behavior is that `windows_textfile_scrape_error` returns `1` but other metrics should continue to work.

breed808:
Resolves #755

### DIFF
--- a/collector/textfile.go
+++ b/collector/textfile.go
@@ -66,9 +66,9 @@ func NewTextFileCollector() (Collector, error) {
 	}, nil
 }
 
-// Given a metric family, determine if any two entries are duplicates.
+// Given a slice of metric families, determine if any two entries are duplicates.
 // Duplicates will be detected where the metric name, labels and label values are identical.
-func duplicateMetricEntry(metricFamilies map[string]*dto.MetricFamily) bool {
+func duplicateMetricEntry(metricFamilies []*dto.MetricFamily) bool {
 	uniqueMetrics := make(map[string]map[string]string)
 	for _, metricFamily := range metricFamilies {
 		metric_name := *metricFamily.Name
@@ -249,6 +249,10 @@ func (c *textFileCollector) Collect(ctx *ScrapeContext, ch chan<- prometheus.Met
 		error = 1.0
 	}
 
+	// Create empty metricFamily slice here and append parsedFamilies to it inside the loop.
+	// Once loop is complete, raise error if any duplicates are present.
+	// This will ensure that duplicate metrics are correctly detected between multiple .prom files.
+	var metricFamilies = []*dto.MetricFamily{}
 fileLoop:
 	for _, f := range files {
 		if !strings.HasSuffix(f.Name(), ".prom") {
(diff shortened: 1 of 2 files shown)


carlpett:
Just making sure I read this correctly: This PR will check for duplicates within each file (but not across multiple files), and drop the entire file in case there are duplicates? We would still expose those files that do not have duplicates, but set `collector_success` to 0?
There was some work on this a couple of years ago in #274 which might have some valuable historical context as well.

carlpett:
The strictest/"most correct" would probably be to merge all textfiles, and then check the complete set for validity. If it is not valid, we reject the entire textfile collector results, and emit no metrics at all.
Do you think that'd be reasonably easy to implement? I guess it'd involve a bunch of buffering, or trying to do something clever...

breed808:
It might be possible by stuffing each `MetricFamily` into a slice defined before the file loop, then checking for duplicates :thinking:
I'll have a look and see if I can get it working.

breed808:
I've got this working and passing CI, have a look and let me know if there's any concerns.

collector/textfile.go
```diff
@@ -65,6 +66,31 @@ func NewTextFileCollector() (Collector, error) {
 	}, nil
 }
 
+// Given a metric family, determine if any two entries are duplicates.
```
carlpett, line 69: Nitpick, docs haven't kept up with the code changes :) 
```suggestion
// Given a slice of metric families, determine if any two entries are duplicates.
```

collector/textfile.go
```diff
@@ -271,10 +301,20 @@ fileLoop:
...
+
+	if duplicateMetricEntry(metricFamilies) {
+		log.Errorf("Duplicate metrics detected in files")
+		error += 1.0
+	} else {
+		for _, mf := range metricFamilies {
 			convertMetricFamily(mf, ch)
 		}
 	}
+	// If crash is caused by convertMetricFamily above, move to an else {} block here
```
carlpett, line 317: What does this mean?

collector/textfile.go
```diff
@@ -271,10 +301,20 @@ fileLoop:
...
 
-		for _, mf := range parsedFamilies {
+		for _, metricFamily := range parsedFamilies {
+			metricFamilies = append(metricFamilies, metricFamily)
+		}
+	}
+
+	if duplicateMetricEntry(metricFamilies) {
+		log.Errorf("Duplicate metrics detected in files")
+		error += 1.0
```
carlpett, line 311: Hm, this would mean we could end up with `windows_textfile_scrape_error == 2`, if there are any other errors? Is there a reason to not just set it to 1 instead?

collector/textfile.go
```diff
@@ -271,10 +301,20 @@ fileLoop:
...
+	} else {
+		for _, mf := range metricFamilies {
 			convertMetricFamily(mf, ch)
 		}
 	}
+	// If crash is caused by convertMetricFamily above, move to an else {} block here
 
 	c.exportMTimes(mtimes, ch)
```
carlpett, line 319: Should we still export the mtimes if there are duplicates? I'd expect us to only emit the scrape_error metric in that case?

carlpett (commented):
Nice! Some minor questions

collector/textfile.go
breed808: Oops, this was a leftover TODO when I was writing the change. I've removed that now.

collector/textfile.go
breed808: Good spot! I've fixed this in a recent push.

collector/textfile.go
breed808: Yes, no reason to add here. I changed this to `=` rather than the current `+=`.

collector/textfile.go
breed808: Another good spot; I've moved this so it's only run if no duplicates are detected.

carlpett (approved):
Sorry, missed the push! Looks good, very nice work!

Issue #570, opened 2020-07-11
NIC name missing ?


SupraOva:
Hi,

Everything is in the title, the name of the NIC interface seems to be missing, or I couldn't find it.

## TYPE windows_net_bytes_received_total counter
windows_net_bytes_received_total{nic="**vmxnet3_Ethernet_Adapter**"} 4.7875466e+07
windows_net_bytes_received_total{nic="**vmxnet3_Ethernet_Adapter__2**"} 4.255629273e+09

The metrics shows the interface description instead name. Is there a way to display it ?

carlpett:
This is the name as far as Windows is concerned. What field/data did you expect instead?

SupraOva:
I rather expect the name in this command :
`Get-WmiObject Win32_NetworkAdapter | Select-Object -ExpandProperty NetConnectionID`

See https://docs.microsoft.com/en-us/windows/win32/cimwin32prov/win32-networkadapter

> NetConnectionID
>     Data type: string
>     Access type: Read/write
>     **Name of the network connection** as it appears in the Network Connections Control Panel program.
>

scluff:
NetConnectionID is not a member of the [Win32_PerfRawData_Tcpip_NetworkInterface](https://technet.microsoft.com/en-us/security/aa394340(v=vs.80)) class, which is what this collector uses.

https://github.com/prometheus-community/windows_exporter/blob/b64ccbe68307385d9d9f2dcdffdcdd776216406e/collector/net.go#L148-L164

SupraOva:
Is there a way to match the name that appears in class Win32_PerfRawData_Tcpip_NetworkInterface with the name in Win32_NetworkAdapter class ?

JDA88:
I agree, without this label the NET metrics are almost useless on any computer that have more than one NIC.

You can match `Win32_NetworkAdapter` Name property with the `Win32_NetworkAdapter` Name property and add the NetConnectionID as a `name` label?

```
Get-WmiObject Win32_PerfRawData_Tcpip_NetworkInterface -Property Name                  | Select-Object Name                  | FT -Autosize
Get-WmiObject Win32_NetworkAdapter                     -Property Name, NetConnectionID | Select-Object Name, NetConnectionID | FT -Autosize
```

SupraOva:
Up. 

Any news about the nic name matching ?

carlpett:
[...]

PR #764, opened 2021-04-19
Adds netConnectionId label to the net collector

chriswiggins:
This PR addresses an issue raised in #570 which we've also encountered - we'd like to be able to identify the network adapters based on their "Windows" name rather than the hardware NIC name. 

We often see Hyper-V switching which NIC is first/second when provisioning a new VM, so by adding this we're able to identify them based on name.

This is the first time I've ever written any go, so I've borrowed concepts from the `cpu` collector executing a wmi query. Happy to make whatever changes necessary to make this able to be merged 👍

### DIFF
--- a/collector/net.go
+++ b/collector/net.go
@@ -3,17 +3,23 @@
 package collector
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
 
 	"github.com/prometheus-community/windows_exporter/log"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/StackExchange/wmi"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 func init() {
 	registerCollector("net", NewNetworkCollector, "Network Interface")
 }
+
+const (
+	win32NetworkQuery = "SELECT Name, NetConnectionID FROM Win32_NetworkAdapter"
+)
 
 var (
 	nicWhitelist = kingpin.Flag(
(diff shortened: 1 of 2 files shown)


chriswiggins:
Apologies for the mess trying to get the sign-off sorted. I think I've tackled that now 😄

carlpett:
Hi @chriswiggins,
Thanks for the PR! There's a (very) significant performance overhead in using WMI, which is why we're trying to move to performance counters instead. Is it possible to find this data that way instead?

chriswiggins:
@carlpett unfortunately it doesn't appear so - I've done a lot of digging but the resources aren't so good 😞 

What about if we cached the value on startup? Or cached it for a certain time period?

chriswiggins:
The other option is to make it available via a command-line toggle? We're happy to have the performance hit because it's so crucial to us being able to gather meaningful stats

carlpett:
Ok, unfortunate that the easy way out wasn't possible. Thanks for checking!
I think there's two options, then:
- Caching, as you mentioned. I'd guess it'd be enough to refresh the cache at any time you get a cache miss. My main concern with this approach would be how it handles interfaces coming and going, could we end up confusing them?
- Make calls to the Windows APIs. From a quick search, I think [GetInterfaceInfo](https://docs.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getinterfaceinfo) might be suitable? The drawback here is that it can be somewhat finicky to do the interop bits. There's a bit of inspiration to be had in #702 in terms of how to structure it.

chriswiggins:
The second option looks great and from first glance I think easily possible. 

I'll investigate and get back to you! Thanks

PR #783, opened 2021-05-16
Remove hard-coded "Computer Queues" filter

### LINKED ISSUE #769: MQMQ collector is only fetching computer queues and ignoring rest
Hi we are using wmi exporter extensively and monitoring more than 20k servers and use standard system collectors.
We are now testing to use the msmq collector but found that it is only collecting the computer queues.
The version we use is v0.11.0
Is there any any enhancements going around msmq collector?

breed808:
msmq collector would only collect from a hard-coded "Computer
Queues" queue. Removal of filter allows other queues to be queried with
the collector.msmq.msmq-where flag.

Resolves #769

### DIFF
--- a/collector/msmq.go
+++ b/collector/msmq.go
@@ -93,29 +93,27 @@ func (c *Win32_PerfRawData_MSMQ_MSMQQueueCollector) collect(ch chan<- prometheus
 	}
 
 	for _, msmq := range dst {
-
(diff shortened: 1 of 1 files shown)


breed808:
Yes, in the release notes I'll need to include a filter that can be used to match the pre-merge behavior.
Thanks for reviewing!

PR #788, opened 2021-05-20
Correct layout of SystemInfo structs

benridley:
This corrects the layout of the SystemInfo structs which previously had a field that was too small, causing the struct to return the wrong values after being modified by the WinAPI function. 

Signed-off-by: Ben Ridley <benridley29@gmail.com>


carlpett (approved):
I have a hard time verifying this at the moment, but looks like it lines up with the docs at least :)

ravpatiltibco:
I have downloaded windows exporter version 0.16.0  from assets section. Few matrices like windows_system_system_up_time getting no values after installing. Please advise.

benridley:
Hi @ravpatiltibco, the System uptime metric is collected differently to the metrics concerned in this PR. From first glance it appears its using `perflib` to find the metrics. Are you getting any values for other perflib metrics like processor_queue_length?

Issue #795, opened 2021-06-03
No monitoring for disks without mounted partitions


rob-scheepens:
While doing fio testing on \\.\PhysicalDrive1 I was under the impression that this corresponded to HarddiskVolume1, sample:

`windows_logical_disk_free_bytes{volume="HarddiskVolume1"} 1.1010048e+08`

However, when I looked at the numbers more closely, none of them made sense. The sample above for example shows some 100MB free, whereas the disk is a RAW device of 24GB. As an experiment I created and mounted a partition on \\.\PhysicalDrive1 , and shortly thereafter I got metrics for HarddiskVolume3, sample:

`windows_logical_disk_free_bytes{volume="HarddiskVolume3"} 2.5697452032e+10`

This matches 24GB much more closely. Thus, it appears that no monitoring objects are created for disks that have no (mounted) partitions, as only HarddiskVolumeX objects are monitored. This makes the windows_exporter unusuable for my projects, as I typically run IO performance tests on raw devices so that results are not skewed by filesystem mechanisms, such as filesystem level caching. Since I expect this to be related to the fact that no Logical Disk objects are created by Windows for non-mounted volumes/disks, I would like to propose to also implement metrics of the [Win32_PerfFormattedData_PerfDisk_PhysicalDisk class](https://docs.microsoft.com/en-us/previous-versions/aa394262(v=vs.85)), which are identical to their LogicalDisk counterparts.

Was there a conscious decision to implement LogicalDisk metrics instead of PhysicalDisk?

rob-scheepens:
I created an experimental implementation of physical_disk.go at https://github.com/rob-scheepens/windows_exporter/commit/4576fc0675cc8cc4148dc483949787665daca27b, but I am not seeing metrics yet. The collector is working though:
```
C:\>windows_exporter.exe --collectors.enabled="physical_disk" --log.level=debug
time="2021-06-07T23:17:10-07:00" level=debug msg="Initializing SWbemServices" source="exporter.go:248"
time="2021-06-07T23:17:10-07:00" level=info msg="Enabled collectors: physical_disk" source="exporter.go:347"
time="2021-06-07T23:17:10-07:00" level=info msg="Starting windows_exporter (version=, branch=, revision=)" source="exporter.go:399"
time="2021-06-07T23:17:10-07:00" level=info msg="Build context (go=go1.15.5, user=, date=)" source="exporter.go:400"
[...]

Issue #799, opened 2021-06-09
Latency from logical disk collector


monismonther:
Hi, first of all I would like to thank you for the great effort put into this exporter

We have a problem with disk latency metrics. How are these numbers supposed to be interpreted, when reading them from perfmon or directly from powershell, the numbers come up in milli or micro seconds , for example five samples show as follows

Avg. Disk sec/Read 0.004
Avg. Disk sec/Read 0.0006
Avg. Disk sec/Read 0.0008
Avg. Disk sec/Read 0.002
Avg. Disk sec/Read 0.005

Note that these are in seconds, so the numbers are 4 millisecond 600 micro, 800 micro, 2 milli and 5 milli respectively

However under grafana we use rate(windows_logical_disk_read_seconds_total[1m]) but get numbers that are in hundreds of milliseconds or 1 -2 seconds sometimes. Checking the source code you read the value and multiply it by the ticksToSecondsScaleFactor which is 1/ 1e7 , why do you do that? and how are these number coming like that

We think disk latency is very crucial and there should be some solution to this
Are we missing something very obvious here? Hope someone can help

rob-scheepens:
I suspect this is related to Frequency_Sys100NS, but I fail to see the relation here. It appears to me that to get ms instead of s, we should multiply by 1,000, not by 10,000,000. To get some additional understanding, I had a look at the raw perfdata counter on a 2019 Core system, but that confused me even more:
```
PS C:\Users\Administrator> Get-CimInstance Win32_PerfRawData_PerfDisk_PhysicalDisk | ? { $_.Name -like "1" } | ft Name,AvgDiskSecPerWrite

Name AvgDiskSecPerWrite
---- ------------------
1             617492156

PS C:\Users\Administrator> Get-CimInstance Win32_PerfRawData_PerfDisk_PhysicalDisk | ? { $_.Name -like "1" } | ft Name,AvgDiskSecPerWrite

Name AvgDiskSecPerWrite
---- ------------------
1             769101384
```
Surely these values cannot be seconds. In my case they should be close to 1ms (SSD disks). It looks to me that the author(s) of the collector knew of this issue with the raw data and mitigated it. Perhaps different Windows versions use different reporting metrics, despite the name still suggesting it in seconds.

@monismonther : what OS were you using to obtain the numbers you mention above?

rob-scheepens:
I've created a support request with Microsoft to get clarification.

monismonther:
@rob-scheepens 

[...]

PR #803, opened 2021-06-14
First implementation of physical_disk collector

rob-scheepens:
This PR is to implement a first incarnation of the physical_disk collector, as derived from the logical_disk collector. This as a result from the request in https://github.com/prometheus-community/windows_exporter/issues/795#issuecomment-860029644.


collector/physical_disk.go
```diff
@@ -0,0 +1,269 @@
...
+		SplitIOs: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "split_ios_total"),
+			"The number of I/Os to the disk were split into multiple I/Os (PhysicalDisk.SplitIOPerSec)",
+			[]string{"disk"},
+			nil,
+		),
+
+		ReadLatency: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "read_latency_seconds_total"),
+			"Shows the average time, in seconds, of a read operation from the disk (PhysicalDisk.AvgDiskSecPerRead)",
```
carlpett, line 118: This (and for `WriteLatency` as well) is actually total latency for all reads, correct?

collector/physical_disk.go
```diff
@@ -0,0 +1,269 @@
...
+	AvgDiskSecPerTransfer  float64 `perflib:"Avg. Disk sec/Transfer"`
+}
+
+func (c *PhysicalDiskCollector) collect(ctx *ScrapeContext, ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
+	var dst []PhysicalDisk
+	if err := unmarshalObject(ctx.perfObjects["PhysicalDisk"], &dst); err != nil {
+		return nil, err
+	}
+
+	for _, disk := range dst {
```
carlpett, line 176: The `disk` label currently looks like this: `"0 C:"`. Since I remember your use-case was in non-mounted disks, what does it look like under those circumstances @rob-scheepens?
I'm not super happy about having both disk number **and** mountpoint, to be honest. Depending on what the label is for non-mounted, perhaps we can do something that'll look reasonable for both cases.

carlpett (commented):
Nice! Some feedback below. Also the CI is failing because you added it as a default collector, which means the end-to-end tests need to be updated (see [tools/e2e-output.txt](https://github.com/prometheus-community/windows_exporter/blob/master/tools/e2e-output.txt))

collector/physical_disk.go
rob-scheepens: For non-mounted disks you will only see the disk number. I agree not to have the label and number, I was actually surprised to see C: show up. ;)

collector/physical_disk.go
rob-scheepens: Not sure, the MSFT description says "Average time, in seconds, of a read operation of data from the disk.". I think this is averaged against the previous sample, not in total. But again, I'm not sure.

rob-scheepens:
> Nice! Some feedback below. Also the CI is failing because you added it as a default collector, which means the end-to-end tests need to be updated (see [tools/e2e-output.txt](https://github.com/prometheus-community/windows_exporter/blob/master/tools/e2e-output.txt))

What has your preference: remove it from the defaults, or shall I update the end-to-end tests?

rob-scheepens:
@carlpett : I've modified e2e-output and included it in this PR. Can you check if all is ok now?

collector/physical_disk.go
carlpett: Was this fixed?

carlpett:
Thanks @rob-scheepens, apart from clarification on if one of the old review comments has been addressed, this looks good! Can you make sure to sign the DCO, though?

carlpett:
Thanks! The e2e test appear to fail now though, although it's not clear to me why. Do they work locally for you?

collector/physical_disk.go
rob-scheepens: Nope, not yet.

rob-scheepens:
> Thanks! The e2e test appear to fail now though, although it's not clear to me why. Do they work locally for you?

When I execute "C:\git\rob-scheepens\windows_exporter>powershell -NonInteractive -ExecutionPolicy Bypass -File .\tools\end-to-end-test.ps1" that works fine, but when I trigger the script through make, it fails:

```
C:\git\rob-scheepens\windows_exporter>make e2e-test
powershell -NonInteractive -ExecutionPolicy Bypass -File .\tools\end-to-end-test.ps1
 
InputObject                                                                                                                                                     SideIndicator
-----------                                                                                                                                                     -------------
windows_exporter_collector_success{collector="physical_disk"} 1                                                                                                 <=
windows_exporter_collector_timeout{collector="physical_disk"} 0                                                                                                 <=
# HELP windows_physical_disk_idle_seconds_total Seconds that the disk was idle (PhysicalDisk.PercentIdleTime)                                                   <=
# TYPE windows_physical_disk_idle_seconds_total counter                                                                                                         <=
# HELP windows_physical_disk_read_bytes_total The number of bytes transferred from the disk during read operations (PhysicalDisk.DiskReadBytesPerSec)           <=
# TYPE windows_physical_disk_read_bytes_total counter                                                                                                           <=
# HELP windows_physical_disk_read_latency_seconds_total Shows the average time, in seconds, of a read operation from the disk (PhysicalDisk.AvgDiskSecPerRead)  <=
# TYPE windows_physical_disk_read_latency_seconds_total counter                                                                                                 <=
# HELP windows_physical_disk_read_seconds_total Seconds that the disk was busy servicing read requests (PhysicalDisk.PercentDiskReadTime)                       <=
# TYPE windows_physical_disk_read_seconds_total counter                                                                                                         <=
# HELP windows_physical_disk_read_write_latency_seconds_total Shows the time, in seconds, of the average disk transfer (PhysicalDisk.AvgDiskSecPerTransfer)     <=
# TYPE windows_physical_disk_read_write_latency_seconds_total counter                                                                                           <=
# HELP windows_physical_disk_reads_total The number of read operations on the disk (PhysicalDisk.DiskReadsPerSec)                                               <=
# TYPE windows_physical_disk_reads_total counter                                                                                                                <=
# HELP windows_physical_disk_requests_queued The number of requests queued to the disk (PhysicalDisk.CurrentDiskQueueLength)                                    <=
# TYPE windows_physical_disk_requests_queued gauge                                                                                                              <=
# HELP windows_physical_disk_split_ios_total The number of I/Os to the disk were split into multiple I/Os (PhysicalDisk.SplitIOPerSec)                          <=
# TYPE windows_physical_disk_split_ios_total counter                                                                                                            <=
# HELP windows_physical_disk_write_bytes_total The number of bytes transferred to the disk during write operations (PhysicalDisk.DiskWriteBytesPerSec)          <=
# TYPE windows_physical_disk_write_bytes_total counter                                                                                                          <=
# HELP windows_physical_disk_write_latency_seconds_total Shows the average time, in seconds, of a write operation to the disk (PhysicalDisk.AvgDiskSecPerWrite) <=
# TYPE windows_physical_disk_write_latency_seconds_total counter                                                                                                <=
# HELP windows_physical_disk_write_seconds_total Seconds that the disk was busy servicing write requests (PhysicalDisk.PercentDiskWriteTime)                    <=
# TYPE windows_physical_disk_write_seconds_total counter                                                                                                        <=
# HELP windows_physical_disk_writes_total The number of write operations on the disk (PhysicalDisk.DiskWritesPerSec)                                            <=
# TYPE windows_physical_disk_writes_total counter                                                                                                               <=
 
 
make: *** [Makefile:19: e2e-test] Error 1
```
I've attached a make -d log but that didn't show anything obvious to me. Also running ProcMon during make did not show an obvious problem, log also attached.

Is the exit code a standard Windows code? If so it means "file not found", but as said, I didn't see that yet in the ProcMon file or make debug log. Thoughts @carlpett ?

[make -d log.txt](https://github.com/prometheus-community/windows_exporter/files/7704427/make.-d.log.txt)
[make_e2e-test.zip](https://github.com/prometheus-community/windows_exporter/files/7704475/make_e2e-test.zip)

rob-scheepens:
@carlpett Looking at the various logs generated I noticed that the physical_disk collector is not started during the e2e-test. Looks like I missed a few places where it should be included, working on that.

Could that be the reason of the error code 2? if so, then why would things work ok when running the PS script directly, without make?

rob-scheepens:
Still no go @carlpett , after the three changes I just made. What am I still missing?
```
time="2021-12-13T07:56:38-08:00" level=info msg="Enabled collectors: system, textfile, cpu, cs, logical_disk, net, os, service" source="exporter.go:348"
```

rob-scheepens:
@carlpett : looks like the three additional changes and a rebuild fixed it - can you verify? I am now able to run make e2e-test:
```CreateProcess(C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe,powershell -NonInteractive -ExecutionPolicy Bypass -File .\tools\end-to-end-test.ps1,...)
Putting child 025bcc08 (e2e-test) PID 40258792 on the chain.
Live child 025bcc08 (e2e-test) PID 40258792
Main thread handle = 00000154
Reaping winning child 025bcc08 PID 40258792
Removing child 025bcc08 PID 40258792 from chain.
Successfully remade target file 'e2e-test'.```

**Comment by rob-scheepens** [comment]
@carlpett : build and test succeeded now. :)

File: collector/physical_disk.go (hunk 2 above)
Comment (carlpett) [reply collector/physical_disk.go:176]: I'd like this to be resolved before we merge, since once this is released it's breaking to change the labels. I can see three different cases that can occur, maybe missing some?
- Not mounted
- Mounted as disk root
- Mounted as a directory inside another (eg D:\Foo\Bar)

Some possible resolutions:
- Always use disk number, skip the mountpoint
- Use mountpoint if available (I dislike this, since it'd mean you'd get a label change by mounting the disk)
- Label with disk number as `disk`, create a separate `_info` metric which has the mountpoint

What do you think?

**Review by carlpett — COMMENTED** [review]
Peak season delays in reviews right now... Thanks for fixing the E2E!
Let's get the final label question resolved and then we can merge :cake:

File: collector/physical_disk.go (hunk 2 above)
Comment (rob-scheepens) [reply collector/physical_disk.go:176]: I'd vote for a fix that shows Raw as label for disks that do not have a partition, and thus are also not mounted. We could combine this with the option of always using disk number, so that in case of a raw disk we would see "1 Raw", and in the case of a mounted partition, we would see the behavior as it currently is. Does that make sense?

Comment (rob-scheepens) [reply collector/physical_disk.go:176]: Testing Labels
==============
Test vm's disks:
```
DISKPART> lis dis

  Disk ###  Status         Size     Free     Dyn  Gpt
  --------  -------------  -------  -------  ---  ---
  Disk 0    Online          100 GB      0 B        *
  Disk 1    Offline          20 GB    20 GB
  Disk 2    Offline          20 GB    20 GB
  Disk 3    Offline          20 GB    20 GB
  Disk 4    Offline          20 GB    20 GB
  Disk 5    Offline          20 GB    20 GB
  Disk 6    Offline          20 GB    20 GB
  Disk 7    Online           50 GB    50 GB
  Disk 8    Online           50 GB    50 GB

windows_physical_disk_idle_seconds_total{disk="1"} 1742.6185977999999
windows_physical_disk_idle_seconds_total{disk="2"} 1742.6185988
windows_physical_disk_idle_seconds_total{disk="3"} 1742.6185944
windows_physical_disk_idle_seconds_total{disk="4"} 1742.6185948999998
windows_physical_disk_idle_seconds_total{disk="5"} 1742.6185957999999
windows_physical_disk_idle_seconds_total{disk="6"} 1742.6185965
windows_physical_disk_idle_seconds_total{disk="7"} 40.4170576
windows_physical_disk_idle_seconds_total{disk="8"} 27.7063651
```
Note: disk 0 is excluded by blacklisting C:. None of disks 1-8 contain any partition.

Created partition on disk 7 using "crea par pri". No change in labels:
```
windows_physical_disk_idle_seconds_total{disk="1"} 1871.906501
windows_physical_disk_idle_seconds_total{disk="2"} 1871.9065021
windows_physical_disk_idle_seconds_total{disk="3"} 1871.9064982
windows_physical_disk_idle_seconds_total{disk="4"} 1871.9064985999998
windows_physical_disk_idle_seconds_total{disk="5"} 1871.9064994999999
windows_physical_disk_idle_seconds_total{disk="6"} 1871.9065008
windows_physical_disk_idle_seconds_total{disk="7"} 169.68967189999998
windows_physical_disk_idle_seconds_total{disk="8"} 156.99427
```
Assigned letter X using "assign letter=x", no change in labels still. 
Also after formatting no change in labels. Only after reboot the X: volume shows up in the metrics list:
```
windows_physical_disk_idle_seconds_total{disk="1"} 7.8161184
windows_physical_disk_idle_seconds_total{disk="2"} 7.8161192999999995
windows_physical_disk_idle_seconds_total{disk="3"} 7.8161203
windows_physical_disk_idle_seconds_total{disk="4"} 7.8161213
windows_physical_disk_idle_seconds_total{disk="5"} 7.8161219
windows_physical_disk_idle_seconds_total{disk="6"} 7.8161236999999995
windows_physical_disk_idle_seconds_total{disk="7 X:"} 7.816124599999999
windows_physical_disk_idle_seconds_total{disk="8"} 7.8161261
```
Is this an issue with the exporter? The interesting thing is that when I did the same workflow for Z: on disk 8 (crea par pri, format ntfs quick, assign letter z), it did show up immediately:
```
windows_physical_disk_idle_seconds_total{disk="1"} 0.00039109999999999997
windows_physical_disk_idle_seconds_total{disk="2"} 0.00039139999999999997
windows_physical_disk_idle_seconds_total{disk="3"} 0.0003917
windows_physical_disk_idle_seconds_total{disk="4"} 0.0003915
windows_physical_disk_idle_seconds_total{disk="5"} 0.00039139999999999997
windows_physical_disk_idle_seconds_total{disk="6"} 0.0003915
windows_physical_disk_idle_seconds_total{disk="7 X:"} 0.00039139999999999997
windows_physical_disk_idle_seconds_total{disk="8 Z:"} 0.0003917
```
Deleting the partion was also immediately reflected:
```
windows_physical_disk_idle_seconds_total{disk="1"} 90.2290264
windows_physical_disk_idle_seconds_total{disk="2"} 90.2290273
windows_physical_disk_idle_seconds_total{disk="3"} 90.2290281
windows_physical_disk_idle_seconds_total{disk="4"} 90.22902839999999
windows_physical_disk_idle_seconds_total{disk="5"} 90.229029
windows_physical_disk_idle_seconds_total{disk="6"} 90.2290299
windows_physical_disk_idle_seconds_total{disk="7 X:"} 90.2290305
windows_physical_disk_idle_seconds_total{disk="8"} 90.1772701
```
Creating two partitions on disk 8 also was immediately reflected:
```
windows_physical_disk_idle_seconds_total{disk="1"} 157.879834
windows_physical_disk_idle_seconds_total{disk="2"} 157.8798352
windows_physical_disk_idle_seconds_total{disk="3"} 157.8798363
windows_physical_disk_idle_seconds_total{disk="4"} 157.8798369
windows_physical_disk_idle_seconds_total{disk="5"} 157.87983739999999
windows_physical_disk_idle_seconds_total{disk="6"} 157.8798382
windows_physical_disk_idle_seconds_total{disk="7 X:"} 157.8798388
windows_physical_disk_idle_seconds_total{disk="8 Y: Z:"} 157.77886379999998
```
To me, this labeling looks consistent and in line with what one would expact. You agree @carlpett? If so, we can put this conversation to rest and merge the PR, right?

**Comment by rob-scheepens** [comment]
Sorry for the delay @carlpett, for some reason I was sure I had already replied... :( 

Regarding the labeling: I suggest we use the PhysicalDisk number in combination with a mountpoint if appropriate. Technically speaking a disk (PhysicalDisk) is never mounted, only partitions (LogicalDisks) are. An example label for C: would then look like:
```
0 C:
```
An example label for a raw disk without any partitions would be:
```
1
```
This seems to match your third possible resolution mentioned above.

Would you agree?

rob-scheepens:
@carlpett Just had another look at the labels on a test machine, and to me this looks good. Are you ok with this too?
image

breed808:
@rob-scheepens I don't think Calle's available at the moment.
I could step in and review based on the existing comments.

Could you rebase the feature branch on `master` so the CI passes? There's also a merge conflict with the Go module files.

rob-scheepens:
> @rob-scheepens I don't think Calle's available at the moment. I could step in and review based on the existing comments.
> 
> Could you rebase the feature branch on `master` so the CI passes? There's also a merge conflict with the Go module files.

Hi @breed808 , I cleaned up the git commit history and backed out the unused import, so CI should be able to finish ok now.

collector/physical_disk.go
breed808: From a brief review of this conversation I assume option 2 has been chosen?

> Use mountpoint if available (I dislike this, since it'd mean you'd get a label change by mounting the disk)

If so I'd recommend documenting this labeling behaviour, in order to avoid user surprise. Could you add an entry for this collector in the `docs/` directory?

rob-scheepens: As mentioned in https://github.com/prometheus-community/windows_exporter/pull/803#issuecomment-1198869937, I'd like the behavior where the disk number is always shown, whether or not in combination with a partition letter or mount point, in case there is a mounted partition. If you agree, let me know and I'll create the doc.

breed808: I think either the disk number only or having the disk number and drive letter as separate labels would be fine.
E.G. `windows_physical_disk_idle_seconds_total{disk="7", drive_letter="X:"} 157.8798388`

Keeping the two pieces of information in a single label would have users resorting to regexes for label matching: `windows_physical_disk_idle_seconds_total{disk=".*X:"}` or `windows_physical_disk_idle_seconds_total{disk="7.*"}`

rob-scheepens:
Closing this since we now have a working version of physical_disk collector using PDH instead of Perflib, as mentioned in https://github.com/prometheus-community/windows_exporter/issues/799#issuecomment-1406568524. A PR for that will follow shortly.

rob-scheepens:
Eventually I could reopen this PR, due to GitHub issues. I closed the duplicate I created of this, PR 1211.

image

rob-scheepens:
@breed808 @carlpett : could you please (re)review this?

cbwest3-ntnx:
@breed808 & @carlpett, this MR is tagged for milestone [v0.21.0](https://github.com/prometheus-community/windows_exporter/milestone/4), however [v0.22.0](https://github.com/prometheus-community/windows_exporter/releases/tag/v0.22.0) is already released. Is it possible for this PR to go into the next release? tia!

breed808:
Looks like the CI isn't happy due to the logging dependency in the collector. Have a look at #1192 for context on the logging change.

breed808:
Apologies for the delay in getting back to this. I've re-run the CI and resolved the issues it's found.

Are we happy with using the disk number in the `disk` flag for the `physical_disk` collector metrics?

rob-scheepens:
@breed808 thanks for checking, and yes, we're happy with the disk number.

breed808 (approved):

PR #804, opened 2021-06-14
Export CurrentBandwidth as bytes

### LINKED ISSUE #800: net collector: windows_net_current_bandwidth name and unit don't follow best practices
1. As per [Prometheus best practices](https://prometheus.io/docs/practices/naming/#metric-names) metric names should **have a suffix describing the unit**; looking up metric units in docs is cumbersome: `windows_net_current_bandwidth_bits`

2. As per [Prometheus best practices](https://prometheus.io/docs/practices/naming/#base-units) metrics should **use base units**, allowing for easy comparison: `windows_net_current_bandwidth_bytes`
    > To avoid confusion combining different metrics, always use bytes, even where bits appear more common.

3. There should be a **consensus with [*NIX node exporter](https://github.com/prometheus/node_exporter)** regarding the metric name, either `speed` or `bandwidth`. I'd vote for the latter; surely it needs further discussion.

max-len:
From https://prometheus.io/docs/practices/naming/:
To avoid confusion combining different metrics, always use bytes, even
where bits appear more common.

Fixes #800

### DIFF
--- a/collector/net.go
+++ b/collector/net.go
@@ -118,7 +118,7 @@ func NewNetworkCollector() (Collector, error) {
 			nil,
 		),
 		CurrentBandwidth: prometheus.NewDesc(
-			prometheus.BuildFQName(Namespace, subsystem, "current_bandwidth"),
+			prometheus.BuildFQName(Namespace, subsystem, "current_bandwidth_bytes"),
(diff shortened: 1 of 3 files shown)


carlpett (approved):
Thanks!
_Note to release: This is a breaking change_

PR #809, opened 2021-06-18
Add missing Process Collector metrics

### LINKED ISSUE #797: Working Set - Private process counter is collected but not exposed as a metric
We were using windows_exporter to collect process metrics and noticed the Working Set - Private counter is not one of the metrics exposed. However, it is collected here: https://github.com/prometheus-community/windows_exporter/blob/72c46664dbd73185f19b5b8b2619ea2a17b2e706/collector/process.go#L170

Is there any reason why this is not exposed? Could we contribute this by adding a metric following the pattern of the other process metrics?

breed808:
This PR adds the `windows_process_working_set_private_bytes` and `windows_process_working_set_peak_bytes` metrics, which were previously being collected but not exposed.

I've also moved the collector metric documentation to the collector's markdown file, leaving a less-verbose description in each metric definition.

Finally, the `windows_process_working_set` metric has been renamed to `windows_process_working_set_bytes` to keep in line with the Prometheus metric naming best practices.

Resolves #797


carlpett (approved):
LGTM. We'll need to note the breaking change in the next release notes.

Issue #771, opened 2021-04-26
Service metrics - replace WMI by direct API calls


carlossscastro:
Hi Team,

We are running some tests with the exporter to obtain the list of services running.
While this is running ok we have run into some issues in smaller VM's with limited resources.

We were wondering if you would consider replacing the collection of those metrics using direct API calls instead of WMI which is quite more costly for the vm's.
A direct API approach would be possible either by calling advapi32.dll directly or by using GO native implementation `x/sys/windows/svc/mgr` 

One thing to note is that we would probably lose the status of the service as it is not directly exposed by the api
https://github.com/prometheus-community/windows_exporter/blob/master/collector/service.go#L113-L126

We would be able to PR this, just wanted to know your opinion beforehand.

Cheers,

carlpett:
Hey @carlossscastro,
Yes, we'd be happy to take a contribution for this! Ideally we aim to use perflib as much as possible, but it tends to lack statuses or textual information quite frequently. But please do submit a proposal :)

carlossscastro:
Thanks @carlpett!

We'll work on the code changes in the next few days and submit our proposal for review.
Appreciate your quick reply.

carlossscastro:
Hi @carlpett,

We have submitted a PR for review (https://github.com/prometheus-community/windows_exporter/pull/812).
We'd greatly appreciate some feedback on the approach!

Thanks!

PR #812, opened 2021-07-01
Services collection using API (no WMI)

carlossscastro:
Team,

As we have mentioned before in #771, we are currently facing some issues while trying to run the Windows Exporter on machines under certain circumstances, with those being specially memory and CPU pressure. After doing some research, we have traced these issues back to the use of the WMI query interface. WMI is known to be not very well performant and to have issues when the machine is under this kind of resource pressure, as reported for example [here](https://answers.microsoft.com/en-us/windows/forum/all/solved-wmi-provider-host-high-cpu-usage-and-laptop/44cb3953-a883-404a-a94f-ee4f363d3a4b), [here](https://social.technet.microsoft.com/Forums/en-US/ea71875c-0a86-42eb-abd2-2429837c60f4/wmi-[…]n-windows-server-2012-datacenter?forum=winserverhyperv) and [here](https://www.howtogeek.com/332838/what-is-the-wmi-provider-host-wmiprvse.exe-and-why-is-it-using-so-much-cpu/).

In this PR, we implement an alternative way of collecting windows service metrics, which is the use case we currently face the most challenging, through the windows APIs rather than WMI. We have tested this solution and it does solve most of the problems we faced when running the exporter in "hot" machines. Use of the new mechanism is opt-in, by specifying the `collector.service.useAPI` flag.

Unfortunately, when this new mechanism is used, the `collector.service.services-where` flag, based on WQL, which was previously appended to the WMI query, can no longer be honored, and if both are present we just log a warning to inform the user about this. Additionally, the `status` field of the service is not reported either, as this was also produced by WMI. Users can instead look at the `state` property which should have the same information, although formatted in a different way.

We'd greatly appreciate some feedback on the approach! Any suggestions on both the code and the UX of the new behavior would be more than welcome.


roobre:
Hi there! It's been a while since we opened this PR, is there anything we can do to get some feedback on it? We're open to make changes in the design and/or implementation if the current patch is not aligned with what you're looking for.

If you could take a look @carlpett @breed808 it would be fantastic!
And apologies for pinging you directly, I saw you were active reviewers in this repo and I am not sure if the team mentioned in the `CODEOWNERS` file is up to date 😕 

Many thanks in advance!

breed808:
Hi all, apologies for the delayed response with this. I've had some personal matters taking away from my free time in the past couple months.

I'm quite happy with the code, particularly that it's behind a disabled-by-default flag, so this can be introduced with minimal impact for existing users :+1: 

I've run some synthetic benchmarks between the two service collector sources, and I'm getting a ~15% increase in collector time compared to the WMI source. Is this expected?
It's not a big deal as the the anticipated gains in stability more than outweigh a slight performance decrease.

roobre:
Thanks a lot for the feedback @breed808!

This surprises me as well, I'm not 100% sure about the exact benchmarks we have made in our end, but I'd also expect the API collection to be faster than WMI.

I'll check with the team and see if we can reproduce and track down that slowdown to something concrete.

breed808:
Thanks @roobre! I wouldn't consider this a blocker for merging, but it would be nice to see if you can reproduce it before we continue :+1:

carlossscastro:
@breed808 

We were able to spot the issue. There was a logic problem on the code where the exporter was actually collecting metrics twice.
We just submitted the fix so if you could now run your tests against it would be great.

Let us know if any other changes are needed.

Thanks again for your feedback!

breed808 (approved):
Very nice! I can see a ~80% decrease in query time with the API method, with the recent changes.
You'll need to sign the [DCO](https://github.com/prometheus-community/windows_exporter/pull/812/checks?check_run_id=3402595265) in order for the CI to pass, then we're good to merge.

carlossscastro:
@breed808 

Thanks for testing. Great to see some nice performance improvements.
I pushed again with a signed commit.

Let us know if anything else is needed!

PR #817, opened 2021-07-14
WDDM metrics

rubu:
I'm opening a PR for feature that collects GPU metrics via WDDM in the perflib collector (initially the idea was to collect NVIDIA GPU metrics via NVML but that does not work with cards in WDDM mode, but that functionality is in this PR as well).

### DIFF
--- a/collector/nvidia.go
+++ b/collector/nvidia.go
@@ -1,0 +1,249 @@
+// +build windows
+
+package collector
+
+/*
+#include "nvml.h"
+#include <windows.h>
+
+// definition of required import symbols
+nvmlReturn_t DECLDIR (*nvmlInit_v2Ptr)() = NULL;
+const DECLDIR char* (*nvmlErrorStringPtr)(nvmlReturn_t result) = NULL;
+nvmlReturn_t DECLDIR (*nvmlDeviceGetCount_v2Ptr)(unsigned int *deviceCount) = NULL;
+nvmlReturn_t DECLDIR (*nvmlDeviceGetHandleByIndex_v2Ptr)(unsigned int index, nvmlDevice_t *device);
+nvmlReturn_t DECLDIR (*nvmlDeviceGetNamePtr)(nvmlDevice_t device, char *name, unsigned int length);
+nvmlReturn_t DECLDIR (*nvmlDeviceGetGraphicsRunningProcesses_v2Ptr)(nvmlDevice_t device, unsigned int *infoCount, nvmlProcessInfo_t *infos);
+nvmlReturn_t DECLDIR (*nvmlSystemGetProcessNamePtr)(unsigned int pid, char *name, unsigned int length);
+
+nvmlReturn_t invoke_nvmlInit_v2()
+{
+	return (*nvmlInit_v2Ptr)();
+}
+
+const char *invoke_nvmlErrorString(nvmlReturn_t error)
+{
+	return (*nvmlErrorStringPtr)(error);
+}
+
+nvmlReturn_t invoke_nvmlDeviceGetCount_v2(unsigned int *deviceCount)
+{
+	return (*nvmlDeviceGetCount_v2Ptr)(deviceCount);
+}
+
+nvmlReturn_t invoke_nvmlDeviceGetHandleByIndex_v2(unsigned int index, nvmlDevice_t *device)
+{
+	return (*nvmlDeviceGetHandleByIndex_v2Ptr)(index, device);
+}
+
+nvmlReturn_t invoke_nvmlDeviceGetName(nvmlDevice_t device, char *name, unsigned int length)
+{
+	return (*nvmlDeviceGetNamePtr)(device, name, length);
+}
+
+nvmlReturn_t invoke_nvmlDeviceGetGraphicsRunningProcesses_v2(nvmlDevice_t device, unsigned int *infoCount, nvmlProcessInfo_t *infos)
+{
+	return (*nvmlDeviceGetGraphicsRunningProcesses_v2Ptr)(device, infoCount, infos);
+}
+
+nvmlReturn_t invoke_nvmlSystemGetProcessName(unsigned int pid, char *name, unsigned int length)
+{
+	return (*nvmlSystemGetProcessNamePtr)(pid, name, length);
+}
+*/
+import "C"
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"syscall"
+	"unsafe"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+func init() {
+	registerCollector("nvidia", newNvidiaCollector, "NVIDIA")
+}
+
+var (
+	gpuProcessWhitelist = kingpin.Flag(
+		"collector.nvidia.process_whitelist",
+		"Regexp of processes to include. Process name must both match whitelist and not match blacklist to be included.",
+	).Default(".*").String()
+	gpuProcessBlacklist = kingpin.Flag(
+		"collector.nvidia.process_blacklist",
+		"Regexp of processes to exclude. Process name must both match whitelist and not match blacklist to be included.",
+	).Default("").String()
+	nvidiaGpuWhitelist = kingpin.Flag(
+		"collector.nvidia.gpu_whitelist",
+		"Regexp of NVIDIA GPUs to include in per process metrics.",
+	).Default(".*").String()
+)
+
+var nvmlValueNotAvailable = C.longlong(C.NVML_VALUE_NOT_AVAILABLE)
+
+type nvidiaCollector struct {
+	TotalGpuMemoryUsed *prometheus.Desc
+
+	gpuProcessWhitelistPattern *regexp.Regexp
+	gpuProcessBlacklistPattern *regexp.Regexp
+
+	nvmlLibrary C.HINSTANCE
+
+	devices []C.nvmlDevice_t
+}
+
+func (error C.nvmlReturn_t) Error() string {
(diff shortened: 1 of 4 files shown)


rubu:
What is the best practice - should I squash the whole feature branch, merge it in my master, rebase it and then open a PR?

collector/nvidia.go
```diff
@@ -0,0 +1,305 @@
+// +build windows
```
carlpett, line 1: I think we might want to use some extra tag (eg nvidia) here, since most people probably won't have the necessary libraries to build this locally.

collector/nvidia.go
```diff
@@ -0,0 +1,305 @@
...
+
+package collector
+
+/*
+#include "nvml.h"
+#include <windows.h>
+
+// definition of required import symbols
```
carlpett, line 9: We have a pattern of putting Cgo-bits in the `headers/` directory, and expose a "nice" Go interface from there, so the collector logic gets easier to read

collector/nvidia.go
```diff
@@ -0,0 +1,305 @@
...
+
+	err := C.invoke_nvmlInit_v2()
+	if err != C.NVML_SUCCESS {
+		return nil, err
+	}
+
+	const subsystem = "nvidia"
+
+	if *processWhitelist == ".*" && *processBlacklist == "" {
+		log.Warn("No filters specified for nvidia collector. This will generate a very large number of metrics!")
```
carlpett, line 190: Is this a copy/paste, or will there actually be a lot of metrics?

collector/nvidia.go
```diff
@@ -0,0 +1,305 @@
...
+			}
+			if nvidiaGpuWhitelistPatttern.MatchString(name) {
+				devices = append(devices, nvdiaGpuDevice{device: device, driverModel: currentDriverModel, name: name})
+			}
+		}
+	}
+
+	return &nvidiaCollector{
+		TotalProcessGpuMemoryUsed: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "total_process_gpu_memory_used"),
```
carlpett, line 229: It is good to indicate the unit of measure in the name, so eg `total_process_gpu_memory_used_bytes` (I guess)

collector/nvidia.go
```diff
@@ -0,0 +1,305 @@
...
+			"Total gpu memory used per process aggregated over all gpus.",
+			[]string{"process", "process_id"},
+			nil,
+		),
+		TotalGpuMemoryUsed: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "total_gpu_memory_used"),
```
carlpett, line 235: Unit here too

collector/nvidia.go
```diff
@@ -0,0 +1,305 @@
...
+				continue
+			}
+			name := string(nameBuffer[:clen(nameBuffer)])
+			var currentDriverModel, pendingDriverModel C.nvmlDriverModel_t
+			err = C.invoke_nvmlDeviceGetDriverModel(device, &currentDriverModel, &pendingDriverModel)
+			if err != C.NVML_SUCCESS {
+				continue
+			}
+			if currentDriverModel == C.NVML_DRIVER_WDDM {
+				log.Warnf("Gpu %s is using WWDM driver that does not allow collecting per process memory usage\n", name)
```
carlpett, line 219: Shouldn't we just `continue` then?

collector/nvidia.go
```diff
@@ -0,0 +1,305 @@
...
+	"syscall"
+	"unsafe"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+func init() {
+	registerCollector("nvidia", newNvidiaCollector, "NVIDIA")
```
carlpett, line 79: You don't seem to unmarshal the `NVIDIA` counters you request here?

collector/process.go
```diff
@@ -2,11 +2,78 @@
 
 package collector
```
carlpett, line 3: Given the amount of code and requirement for extra libraries, I'd be inclined to suggest we put the per-process GPU metrics in a separate collector. What do you think?

collector/process.go
```diff
@@ -2,11 +2,78 @@
...
 
+/*
+#cgo LDFLAGS: -lDXGI
```
carlpett, line 6: As mentioned previously, this'd be nice to put in headers/ with a wrapper

carlpett (commented):
Thanks for the PR @rubu!
There's some code-organizational parts I've mentioned that we need to talk about. Also, I suppose our CI currently lacks a lot of libraries to build this. Any idea what we need to install?

collector/nvidia.go
rubu: This module does not link against any static libs (since the NVML package from NVIDIA simply does not have them) and loads the needed dll via C.LoadLibrary so if it is not there it is just not there. No actual libraries are needed for building.

collector/nvidia.go
rubu: ok, i can rework that.

collector/nvidia.go
rubu: That is not a copy/paste, since my initial thought was that basically any process that is using the GPU (thus for minimum any process that has a GUI) will be exposed here - ofc there won't be as many per process metrics as in the process collector, but the actual entries will be as many as there are processes using the GPU. Again, something we can discuss and work out.

collector/nvidia.go
rubu: Ok noted, will fix.

collector/nvidia.go
rubu: There is also the TotalGpuMemoryUsed, which does not care about WDDM / TCM mode. That is why the continue is not there.

collector/nvidia.go
rubu: This collector does not take anything from perflib, all the WDDM stuff (GPU usage from perflib) has been put in the process collector.

rubu: Sorry, this is literally just an error, since I don't think I initially understood what that parameter does, will remove it.

collector/process.go
rubu: Hmm, well, as for the extra libs, this only uses DXGI.lib, which is present in Windows/System32. The data comes from perflib, the extra code is just to match the GPU luids with the corresponding processes, so since I perceive the process collector as something that takes data from perflib, makes sense for me that this stuff lives here, but again we can talk about it.

rubu:
@carlpett thanks for getting back, so let's try and resolve the points and I'll fix the stuff that needs fixing :) Since I already made some bugfixes it actually became annoying that I have to build this from source every time and redistribute the binary, if the proper way would be to use the official releases. Btw one thing I stumbled upon last week - the promu build command actually silently excluded all the cgo stuff since the yaml file was missing:
`
go:
    cgo: true
`
which I added. So up to this point windows_exporter did not anyhow depend on any C binding?

collector/nvidia.go
rubu: So, at this point the repo does not have a `headers/` directory. Thus I have no idea what exactly should go in there. Can you give me a link to any other project/repo that has such a structure for an example? Sorry, I'm primarily a C++ dev and really do not know the guidelines for Go/C interop.

collector/nvidia.go
carlpett: Ah, ok then :+1:

collector/nvidia.go
carlpett: Hm, perhaps you need to rebase? Looks like this branch started off in February and it might not have existed then, but it does now. There's some background in [this PR](https://github.com/prometheus-community/windows_exporter/pull/702#issuecomment-769785337)

collector/nvidia.go
carlpett: Okay, then no worries! It's just been a rather frequent occurrence historically that contributors start off by copy/pasting one of the collects that generate a lot of metrics and gets this log line by chance, so to speak. From your description it sounds warranted then, I wasn't sure how many matches it would result in.

collector/nvidia.go
carlpett: Alright!

carlpett:
>So up to this point windows_exporter did not anyhow depend on any C bindings?

Hm. It does, eg for the `container` collector (although the cgo stuff itself is in a separate library). I wonder why this would work? Could we be using different versions of `promu`? I believe we pinned the version quite a while ago due to some bug in Promu, there might have been upstream changes since then :grimacing:

okopanja:
Just a short question: is this pull request still active or not?

rubu:
@okopanja I didn't manage to refactor/split all the stuff, and it seems like there is not much need for it, so I can just keep it as is in my local fork. Are you interested in it or just tidying up? In case of the former, I can try and finally make everything nice and repush, otherwise I guess this can be tidied up.

rubu:
@ChandonPierre Hey, sorry for abandoning work on this - simply lack of time and it was easier to live with a local fork since the WDDM/GPU metrics were all that I needed, but can try to find time and give the refactor another go, but I can't make any promises on the timeline. Anyone else maybe interested in helping out?

collector/nvidia.go
```diff
@@ -0,0 +1,305 @@
...
+import "C"
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"syscall"
+	"unsafe"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/log"
```
ChandonPierre, line 74: I believe this should be `"github.com/prometheus-community/windows_exporter/log"`

kczauz:
> 

Hi @rubu - do you have some time to finish this up?  I am interested in having some gpu metrics on windows endpoints.

Issue #808, opened 2021-06-17
Adding Scheduled Tasks Collector Proposal 


mousavian:
Scheduled Tasks Collector
----

## Abstract

Add a new collector called `scheduled_tasks` that can be enabled to provide metrics about registered tasks configured in the Windows Task Scheduler Service.

## Background

Task Scheduler is a component of Microsoft Windows that provides the ability to schedule the launch of computer programs or scripts at pre-defined times or after specified time intervals. [<sup>1</sup>](https://en.wikipedia.org/wiki/Windows_Task_Scheduler)

The importance of having metrics for these scheduled tasks for the sake of monitoring and aletring particularly on production environments is undeniable.

## Proposal

A new collector named `scheduled_tasks` will be added to `collector` directory.

This collector will utilise [go-ole](https://github.com/go-ole/go-ole) library which uses Windows COM technology under the hood to communicate with Scheduler Service, as opposed to Windows Management Instrumentation (WMI) package. The reason for this is that:
> The `Win32_ScheduledJob` class is internally using the AT protocol, which is bound to deprecation starting with Windows 8 and Windows Server 2012. As a first step the AT protocol is disabled by default. [<sup>2</sup>](https://docs.microsoft.com/en-gb/windows/win32/cimwin32prov/win32-scheduledjob?redirectedfrom=MSDN)

Therefore, the `collect` function will initialise COM library at first using the `CoInitialize`[<sup>3</sup>](https://docs.microsoft.com/en-us/windows/win32/api/objbase/nf-objbase-coinitialize) function provided in the `go-ole` package.

A single object of the class associated with a Task Scheduler's CLSID will be created next by calling `CreateInstance`[<sup>4</sup>](https://docs.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-cocreateinstance).

After all the connections are initialised and established successfully, the collector will attemp to enumerate all the folders and collect all the registered tasks via `GetFolder` and `GetTasks` methods. [<sup>5</sup>](https://docs.microsoft.com/en-us/windows/win32/api/taskschd/nn-taskschd-itaskfolder)

As each task comes with bunch of attributes[<sup>5</sup>](https://docs.microsoft.com/en-us/windows/win32/api/taskschd/nn-taskschd-iregisteredtask), this is the time we have to decide which attribute to keep.
My suggestion is the following struct for a task:
```go
type ScheduledTask struct {
[...]

PR #821, opened 2021-07-17
Adding Scheduled Tasks Collector

mousavian:
Adding a new collector called `scheduled_task` that can be enabled to provide metrics about registered tasks configured in the Windows Task Scheduler Service. ( #808 )

### DIFF
--- a/collector/scheduled_task.go
+++ b/collector/scheduled_task.go
@@ -59,6 +59,8 @@ type ScheduledTask struct {
 	LastTaskResult  TaskResult
 }
 
+type ScheduledTasks []ScheduledTask
+
 func init() {
 	registerCollector("scheduled_task", NewScheduledTask)
 }
@@ -161,10 +163,7 @@ const SCHEDULED_TASK_PROGRAM_ID = "Schedule.Service.1"
 // S_FALSE is returned by CoInitialize if it was already called on this thread.
 const S_FALSE = 0x00000001
 
-func getScheduledTasks() ([]ScheduledTask, error) {
-	var err error
-	scheduledTasks := []ScheduledTask{}
-
+func getScheduledTasks() (scheduledTasks ScheduledTasks, err error) {
 	err = ole.CoInitialize(0)
 	if err != nil {
 		code := err.(*ole.OleError).Code()
@@ -176,13 +175,11 @@ func getScheduledTasks() ([]ScheduledTask, error) {
 
 	schedClassID, err := ole.ClassIDFrom(SCHEDULED_TASK_PROGRAM_ID)
 	if err != nil {
-		ole.CoUninitialize()
 		return scheduledTasks, err
 	}
 
 	taskSchedulerObj, err := ole.CreateInstance(schedClassID, nil)
 	if err != nil || taskSchedulerObj == nil {
-		ole.CoUninitialize()
 		return scheduledTasks, err
 	}
 	defer taskSchedulerObj.Release()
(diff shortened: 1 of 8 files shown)


mousavian:
Any idea why CI failed? Not sure why it's still trying to download `go-ole v1.2.1` while this PR suggests higher version in mod file.

breed808 (commented):
This looks rather nice! I'll admit my knowledge of the Win32 API is lacking, so I'm not able to adequately review most of the `getScheduledTasks()` function. @carlpett could you take a look?
Documentation is also very good, thanks for putting in the effort for that :+1:

You're right about the CI, I've been able to compile this PR locally with no issues. We may need to re-trigger the CI build to see if it's a repeatable error.

collector/scheduled_task.go
```diff
@@ -0,0 +1,283 @@
...
+const SCHEDULED_TASK_PROGRAM_ID = "Schedule.Service.1"
+
+// S_FALSE is returned by CoInitialize if it was already called on this thread.
+const S_FALSE = 0x00000001
+
+func getScheduledTasks() ([]ScheduledTask, error) {
+	var err error
+	scheduledTasks := []ScheduledTask{}
+
+	err = ole.CoInitialize(0)
```
carlpett, line 168: As `getScheduledTasks` is called on each scrape, do we need to run initialization every time, or would it suffice to do once? I'm somewhat concerned about interactions with how the WMI library uses OLE as well, and especially with calling `CoUninitialize`.

carlpett: Also, we need to consider concurrent scrapes.

collector/scheduled_task.go
```diff
@@ -0,0 +1,283 @@
...
+	res, err := oleutil.CallMethod(taskServiceObj, "GetFolder", `\`)
+	if err != nil {
+		return scheduledTasks, err
+	}
+
+	rootFolderObj := res.ToIDispatch()
+	defer rootFolderObj.Release()
+
+	var fetchTasksInFolder func(*ole.IDispatch) error
+	var fetchTasksRecursively func(*ole.IDispatch) error
```
carlpett, line 206: Is there a reason not to define these as normal free functions? Am I just missing some reuse of local vars?

collector/scheduled_task.go
```diff
@@ -0,0 +1,283 @@
...
+		code := err.(*ole.OleError).Code()
+		if code != ole.S_OK && code != S_FALSE {
+			return scheduledTasks, err
+		}
+	}
+	defer ole.CoUninitialize()
+
+	schedClassID, err := ole.ClassIDFrom(SCHEDULED_TASK_PROGRAM_ID)
+	if err != nil {
+		ole.CoUninitialize()
```
carlpett, line 179: We call this due to the `defer` anyway, any reason you call it explicitly "again"?

collector/scheduled_task.go
```diff
@@ -0,0 +1,283 @@
...
+		return scheduledTasks, err
+	}
+
+	taskSchedulerObj, err := ole.CreateInstance(schedClassID, nil)
+	if err != nil || taskSchedulerObj == nil {
+		ole.CoUninitialize()
```
carlpett, line 185: Same here, is this needed?

carlpett (commented):
Very nice! I have some questions on the OLE stuff, but generally :+1:

collector/scheduled_task.go
```diff
@@ -0,0 +1,281 @@
...
+
+func parseTask(task *ole.IDispatch) ScheduledTask {
+	scheduledTask := ScheduledTask{}
+
+	scheduledTask.Name = oleutil.MustGetProperty(task, "Name").ToString()
+	scheduledTask.Path = strings.ReplaceAll(oleutil.MustGetProperty(task, "Path").ToString(), "\\", "/")
+	scheduledTask.Enabled = oleutil.MustGetProperty(task, "Enabled").Value().(bool)
+	scheduledTask.State = TaskState(oleutil.MustGetProperty(task, "State").Val)
+	scheduledTask.MissedRunsCount = float64(oleutil.MustGetProperty(task, "NumberOfMissedRuns").Val)
+	scheduledTask.LastTaskResult = TaskResult(oleutil.MustGetProperty(task, "LastTaskResult").Val)
```
carlpett, line 261: On a final re-read, I noticed the use of `Must`-functions - this means we'll panic and crash if there's any issues here, which we really do not want to.

carlpett (changes requested):
This looks very nice to me now, except one thing I missed earlier. With that fixed we'll be good to go :)

collector/scheduled_task.go
mousavian: 💯 + ✔️

mousavian:
I have rebased this... should be ready to merge if tests are passing :)

carlpett (approved):
Nice, thank you!

breed808:
@Gerrit0284 this change hasn't yet made it into a new release yet; this was implemented after v0.18.1.
You can wait for a new release, or build the exporter from `master`.

PR #832, opened 2021-09-11
Collect IIS metrics using Perflib

daveaowen:
The IIS collector was constantly timing out on our servers, due to WMI being slow and unreliable. This change gathers all of the same metrics, but using Perflib instead of WMI, with a significant performance boost.

### DIFF
--- a/collector/iis.go
+++ b/collector/iis.go
@@ -1243,13 +1243,13 @@ type perflibW3SVC_W3WP struct {
 
 	URICacheFlushesTotal       float64 `perflib:"Total Flushed URIs"`
 	URICacheFlushesTotalKernel float64 `perflib:"Total Flushed URIs"`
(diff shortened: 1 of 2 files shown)


collector/iis.go
```diff
@@ -1,24 +1,28 @@
+//go:build windows
```
carlpett, line 1: We're not mandating Go 1.17 yet, my understanding is that the current syntax is still supported? Any other reason to have double tags?

carlpett (commented):
Impressive, thanks @daveaowen!
I don't have an IIS environment to check this on, and we don't appear to have it covered in our end-to-end tests - is all output identical?

collector/iis.go
daveaowen: Fixed. Apologies, not sure how that got in there.

daveaowen:
@carlpett Output is identical

carlpett (approved):
Excellent work, thank you!

Issue #816, opened 2021-07-05
Thermalzone not working


Ramshield:
Hi,

My exporter does not expose any metrics regarding the thermalzone.
It is enabled however:

```
windows_exporter_collector_duration_seconds{collector="thermalzone"} 0.0060052
windows_exporter_collector_success{collector="thermalzone"} 1
windows_exporter_collector_timeout{collector="thermalzone"} 0
````

```
[System.Environment]::OSVersion.Version

Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      19042  0
```

I run windows exporter with the following arguments:
`"C:\Program Files\windows_exporter\windows_exporter.exe" --log.format logger:eventlog?name=windows_exporter  --telemetry.addr :9182   --collectors.enabled cpu,cs,logical_disk,logon,memory,net,os,process,service,system,tcp,time,thermalzone,textfile` 

I'm a Linux engineer, so I have no clue how to troubleshoot something like this. Please advice, thank you!

**Comment by breed808** [comment]
Are there any relevant logs in the Event Viewer? `windows_exporter` will log to `Windows Logs -> Application`.

**Comment by luigi311** [comment]
Getting something similar with no thermal data. Looking at the event viewer and filtering for windows_exporter everything is information except 2 which are warnings 
No filters specified for process collector. This will generate a very large number of metrics!

Looks like all my things are duplicated twice which is why I have 2 warnings with the same message. Everything else seems to be working though just a thermalzone issue.

**Comment by crockk** [comment]
I am also having this issue. I suspect my hardware does not support the thermalzone collector, but I do not know how to validate this.

**Comment by breed808** [comment]
Checking if the `thermalzone` perflib metrics are present would be a good first step:

```powershell
# List Counter Sets (confirm if "Thermal Zone Information" CounterSet is present)
Get-Counter -ListSet * | Sort-Object -Property CounterSetName | Select CounterSetName
# List counters for set
Get-Counter -ListSet 'Thermal Zone Information' 
# Get a counter from the set
Get-Counter -Counter '\Thermal Zone Information(*)\Temperature'
```

I get an error ("Get-Counter: Internal performance counter API call failed. Error: 800007d1.") when running the last command, but that may be due to my VM not having access to any hardware temperature sensors.

Ramshield:
@breed808 thank you for your reply. I run those commands in PowerShell as Administrator.

```
[...]

PR #841, opened 2021-09-25
Thermalzone: return error on empty result

breed808:
See https://github.com/prometheus-community/windows_exporter/issues/816#issuecomment-898065629 for context.


carlpett (approved):
LGTM!

PR #862, opened 2021-11-05
Update dependencies

breed808:
Update project dependencies, notably go-kit/kit, which was responsible for a number of out-of-date dependencies.


jsturtevant:
There are others that are out of date too (hcsshim as example)? Should those be updated too?

carlpett:
It's a good point, but perhaps we can do that as a separate PR? This one to bump the "prometheus-shared", then others for other bits. Makes it somewhat easier to understand what broke, if anything does :)

jsturtevant:
makes sense!  I was also wondering about the go version.  The mod file specifies 1.13: https://github.com/prometheus-community/windows_exporter/blob/ce3d517cb30a629ae30d190a17158771b21abd33/go.mod#L3

breed808:
I had intended a separate PR to propose adding [Dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates) to the project to semi-automate dependency updating.

I missed the hcsshim dependencies (and likely others), I don't mind adding them here or in another PR.

Docs for the go.mod file state the version is a minimum version:

> The minimum version of Go required to compile packages in this module

So my thinking was to leave it at 1.13 rather than arbitrarily increasing the version requirement. Happy to be wrong about this though!

jsturtevant:
> So my thinking was to leave it at 1.13 rather than arbitrarily increasing the version requirement. Happy to be wrong about this though!

thanks for the explanation, I was looking for the correct golang version to build it in for #864 and noticed this

jsturtevant:
lgtm

carlpett (approved):

Issue #856, opened 2021-10-21
Replace IsAnInteractiveSession by IsWindowsService


gillg:
The function https://pkg.go.dev/golang.org/x/sys/windows/svc#IsAnInteractiveSession is deprecated because the way to detect if it's a service or not is not reliable.
As an exemple, if you wrap your service with NSSM, windows_exporter seems think it's a non interactive session.
This seems cause some stucking situations when nssm try to stop it.

Anyway, the previous function is deprecated, so it's a good option to remove it.

PR #863, opened 2021-11-05
use IsWindowsService to detect if running as service

jsturtevant:
Fixes:  #856

When working on https://github.com/prometheus-community/windows_exporter/pull/581#issuecomment-962050709 I got the error `time="2021-11-05T00:18:54Z" level=error msg="Failed to start service: The service process could not connect to the service controller." source="exporter.go:337"`

It was not running as a service so it should not fall into this code path.  I found https://pkg.go.dev/golang.org/x/sys/windows/svc#IsAnInteractiveSession was depreciated so made the switch.

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -1,3 +1,4 @@
+//go:build windows
 // +build windows
 
 package main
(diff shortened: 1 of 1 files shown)


carlpett (approved):
Thanks!

PR #868, opened 2021-11-10
some metric name typed error

xflu:
`windows_net_current_bandwidth_bytes `not exists, but `windows_net_current_bandwidth` .
`windows_net_packets_received_total_total`  not exists, but `windows_net_packets_received_total`.


carlpett:
Hi @xflu,
Re the `windows_net_current_bandwidth`, this is actually correct, but the version of the docs you are reading are for `master`, not the latest released version.

PR #871, opened 2021-11-14
Add Dependabot dependency tracking

breed808:
Bot will submit PRs when new dependency versions are detected,
preventing dependencies from becoming out-of-date.


carlpett (approved):
LGTM, although I could see daily checks possibly being a bit faster than we'll be reacting?

breed808:
That's no issue, I'll set the check to weekly and we can revise if necessary.

PR #882, opened 2021-12-01
Fix default path for textfile collector

geraudster:
This PR fixes the default path for textfile collector. The previous implementation returned the full exe path, producing lots of error message:

> "Error reading textfile collector directory "C:
Program Files\\windows_exporter\\windows_exporter.exe
textfile_inputs": open C:\Program Files\windows_exporter\windows_exporter.exe\textfile_inputs: The system cannot find the path specified."


breed808 (approved):
Thanks for this!

PR #883, opened 2021-12-02
Remove explicit LISTEN_PORT from MSI installer

### LINKED ISSUE #881: telemetry:   addr: not work in config file. (0.16.0)
Hello everyone.
I noticed that the port changes in the config file do not reflect the change.
for example:
Change the port and enable some collectors in the config file.
```yml
telemetry:
  addr: ":9192"
```
after that, the service is starting in the old predefined port. Other changes are reflected.
this is how the service was installed:

`msiexec /i C:\windows_exporter\windows_exporter-0.16.0-amd64.msi LISTEN_PORT=59182 --% EXTRA_FLAGS="--config.file=C:\windows_exporter\config.yml"`

the flag appears to have a higher priority than the config and the config cannot change it
I reinstalled the service without specifying the port and again: the default port cannot be changed from the config file.

breed808:
Explicit setting of listening port in the service definition causes port
setting in configuration file to be ignored.

Exporter already defines a default port (9182) if one is not specified,
so no impact from this change is anticipated.

Resolves #881

### DIFF
--- a/installer/windows_exporter.wxs
+++ b/installer/windows_exporter.wxs
@@ -23,8 +23,10 @@
     <SetProperty Id="ExtraFlags" After="InstallFiles" Sequence="execute" Value="[EXTRA_FLAGS]">EXTRA_FLAGS</SetProperty>
 
     <Property Id="LISTEN_ADDR" Secure="yes" />
-    <Property Id="LISTEN_PORT" Secure="yes" Value="9182" />
-    <SetProperty Id="ListenFlag" After="InstallFiles" Sequence="execute" Value="--telemetry.addr [LISTEN_ADDR]:[LISTEN_PORT]">LISTEN_ADDR OR LISTEN_PORT</SetProperty>
+    <Property Id="LISTEN_PORT" Secure="yes" />
+    <SetProperty Id="ListenFlagBoth" After="InstallFiles" Sequence="execute" Value="--telemetry.addr [LISTEN_ADDR]:[LISTEN_PORT]">LISTEN_ADDR AND LISTEN_PORT</SetProperty>
+    <SetProperty Id="ListenFlagAddr" After="InstallFiles" Sequence="execute" Value="--telemetry.addr [LISTEN_ADDR]:">LISTEN_ADDR AND (NOT LISTEN_PORT)</SetProperty>
+    <SetProperty Id="ListenFlagPort" After="InstallFiles" Sequence="execute" Value="--telemetry.addr 0.0.0.0:[LISTEN_PORT]">LISTEN_PORT AND (NOT LISTEN_ADDR)</SetProperty>
 
     <Property Id="METRICS_PATH" Secure="yes"/>
     <SetProperty Id="MetricsPathFlag" After="InstallFiles" Sequence="execute" Value="--telemetry.path [METRICS_PATH]">METRICS_PATH</SetProperty>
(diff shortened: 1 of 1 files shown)


installer/windows_exporter.wxs
```diff
@@ -23,7 +23,7 @@
     <SetProperty Id="ExtraFlags" After="InstallFiles" Sequence="execute" Value="[EXTRA_FLAGS]">EXTRA_FLAGS</SetProperty>
 
     <Property Id="LISTEN_ADDR" Secure="yes" />
-    <Property Id="LISTEN_PORT" Secure="yes" Value="9182" />
+    <Property Id="LISTEN_PORT" Secure="yes" />
     <SetProperty Id="ListenFlag" After="InstallFiles" Sequence="execute" Value="--telemetry.addr [LISTEN_ADDR]:[LISTEN_PORT]">LISTEN_ADDR OR LISTEN_PORT</SetProperty>
```
carlpett, line 27: It's been a while, but I think the reason it was defaulted is because otherwise if you *only* specify `LISTEN_ADDR`, it'll set the flag to `--telemetry.addr some-addr-specified:`, ie without the port number set. That's not a valid listening specifier, so we'd crash.
I think we'll need to do something clever/ugly with an additional property and changing the condition for `ListenFlag` in order to not set the flag unless something is specified. We have three cases:
- Neither port nor addr given -> Should not set flag
- Port given (addr given or not does not matter) -> Should set flag to addr:port (empty addr is fine)
- Addr given (but not port) -> Should set flag to addr:default_port

breed808: I had a feeling it wouldn't be that easy. I'll have a look to see if I can address those cases.

carlpett (approved):
:+1: Like the approach! Much simpler than what I saw a one-liner would be, nice idea of splitting them up instead!

breed808:
I've just managed to test MSI creation with this branh, and WIX is not happy with the duplicate symbols:

```
PS C:\Users\Ben\Desktop\windows_exporter> .\installer\build.ps1 -PathToExecutable .\windows_exporter.exe -Version 0.16.0 -Arch "amd64"
windows_exporter.wxs
C:\Users\Ben\Desktop\windows_exporter\installer\windows_exporter.wxs(27) : error LGHT0091 : Duplicate symbol 'CustomAction:SetListenFlag' found. This typically means that an Id is duplicated. Check to make sure all your identifiers of a given type (File, Component, Feature) are unique.                                                                          C:\Users\Ben\Desktop\windows_exporter\installer\windows_exporter.wxs(28) : error LGHT0092 : Location of symbol related to previous error.                                           C:\Users\Ben\Desktop\windows_exporter\installer\windows_exporter.wxs(27) : error LGHT0091 : Duplicate symbol 'CustomAction:SetListenFlag' found. This typically means that an Id is duplicated. Check to make sure all your identifiers of a given type (File, Component, Feature) are unique.                                                                          C:\Users\Ben\Desktop\windows_exporter\installer\windows_exporter.wxs(29) : error LGHT0092 : Location of symbol related to previous error.
C:\Users\Ben\Desktop\windows_exporter\installer\windows_exporter.wxs(27) : error LGHT0091 : Duplicate symbol 'WixAction:InstallExecuteSequence/SetListenFlag' found. This typically means that an Id is duplicated. Check to make sure all your identifiers of a given type (File, Component, Feature) are unique.
C:\Users\Ben\Desktop\windows_exporter\installer\windows_exporter.wxs(28) : error LGHT0092 : Location of symbol related to previous error.
C:\Users\Ben\Desktop\windows_exporter\installer\windows_exporter.wxs(27) : error LGHT0091 : Duplicate symbol 'WixAction:InstallExecuteSequence/SetListenFlag' found. This typically means that an Id is duplicated. Check to make sure all your identifiers of a given type (File, Component, Feature) are unique.
C:\Users\Ben\Desktop\windows_exporter\installer\windows_exporter.wxs(29) : error LGHT0092 : Location of symbol related to previous error.
```

I'll have to find a way to implement this on a single line, which won't be pretty.

breed808:
I believe I have it now. Using distinct Wix `Id` values worked, but it's no longer a one-liner.

carlpett (approved):
I suppose you could set a new property `ListenFlag` to the listing of all three if you want to shorten the `ServiceInstall` line, but I'm okay with this :)
Nice work!

Issue #835, opened 2021-09-13
windows_iis_worker_request_errors_total   doesn't work anymore???


DenisAlcc:

fantom732315:
![image](https://user-images.githubusercontent.com/90676446/133267362-e513fd38-a5ef-4509-abf8-69193321c715.png)

It works for me correctly. Did you specify the iis flag at startup?

DenisAlcc:
> ![image](https://user-images.githubusercontent.com/90676446/133267362-e513fd38-a5ef-4509-abf8-69193321c715.png)
> 
> It works for me correctly. Did you specify the iis flag at startup?

yes, i specify the iis flag with msiexec ENABLED_COLLECTORS

fantom732315:
Do you have this item in your metrics or is it always 0? If it is always zero, try to simulate the error

SouenMazouin:
Hi, we are having the same problem as well. 
No worries of this kind on the 0.16.0 release but only when building the master.
I suspect a problem with the perflib integration (#832).
I'm investigating and if I find any leads, I'll let you know

SouenMazouin:
After investigation I noticed that the IIS metrics of version greater than 8 were no longer instantiated,
there was also a small typo that prevented looping on the concerned perflib objects.

I opened a PR to fix this : #887

breed808:
Is this resolved with the merge of #887 or is further work required?

SouenMazouin:
Hi @breed808, the merge of #887 solved the problem for me, @DenisAlcc did that solve with you too?

PR #887, opened 2021-12-14
fix: add missing metrics for IIS version >= 8

SouenMazouin:
Allows the following metrics to be exposed again, they had disappeared after the migration to perflib :
- worker_request_errors_total
- worker_current_websocket_requests
- worker_websocket_connection_accepted_total
- worker_websocket_connection_rejected_total

Fix Issue #835

### DIFF
--- a/collector/iis.go
+++ b/collector/iis.go
@@ -4,11 +4,12 @@ package collector
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


carlpett (approved):
Thanks!

PR #893, opened 2021-12-28
Update CI token

carlpett:
Old token was using outdated format ([see Github post](https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats)), triggering email alerts.


breed808 (approved):
Looks good

PR #898, opened 2021-12-30
Migrate CI/CD to Github Actions

breed808:
Deployment example:

https://github.com/breed808/windows_exporter/actions/runs/1637157270
https://github.com/breed808/windows_exporter/releases/tag/v0.17.1-test4

Differences with current Appveyor config:

1. No `sha265sums.txt` artifact is produced for releases
2. EXE binaries are made available to repository owners but not to the public

Both of these points can be addressed if they're considered blockers.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,129 @@
...
+          go get -u github.com/prometheus/promu@v0.11.1
+          go get -u github.com/josephspurrier/goversioninfo/cmd/goversioninfo@v1.2.0
+          # GOPATH\bin dir must be added to PATH else the `promu` and `goversioninfo` commands won't be found
+          echo "$(go env GOPATH)\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+
+      - name: Build
+        run: |
+          # go mod download (or, if we don't call it, go build) will write every dependent package name to
+          # stderr, which will be interpreted as an error and abort the build if ErrorActionPreference is Stop,
+          # so we need to run it before setting the preference.
```
carlpett, line 85: Just checking - is this the case also for Github Actions?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,129 @@
...
+          foreach($Arch in "amd64","386") {
+            Write-Verbose "Building windows_exporter $MSIVersion msi for $Arch"
+            .\installer\build.ps1 -PathToExecutable .\output\windows_exporter-$BuildVersion-$Arch.exe -Version $MSIVersion -Arch "$Arch"
+            Move-Item installer\Output\windows_exporter-$MSIVersion-$Arch.msi output\
+          }
+
+          promu checksum output\
+
+      - name: Release
+        uses: softprops/action-gh-release@v1
```
carlpett, line 124: In the interest of keeping the number of external dependencies with write access to the repo low: Would it be hard to implement this ourselves as a `make` target (or just a script directly in this file)? I don't know what it conveniences it might include, but I see that `gh` (the Github CLI) is pre-installed on the image, so perhaps it's equivalent to `gh release upload <tag> output\windows_exporter-*.msi output\windows_exporter-*.exe`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,129 @@
...
+        if: ${{ failure() }}
+        run: golangci-lint run --timeout=5m -c .golangci.yaml
+
+  build:
+    runs-on: windows-2019
+    needs:
+      - test
+      - lint
+    env:
+      GO111MODULE: on
```
carlpett, line 67: I think we might actually not need this any more? It should be the default on the versions we use now

.github/workflows/ci.yml
```diff
@@ -0,0 +1,129 @@
...
+    branches:
+      - master
+  release:
+    types:
+      - published
+      - edited
+
+jobs:
+  test:
+    runs-on: windows-2019
```
carlpett, line 16: As I understand from [the runner docs](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#go), this image will default us to Go 1.15. The AppVeyor image we've been using is (currently, at least) on 1.17.
It looks like we can opt in to running a more modern version by tweaking so we use `GOROOT_1_17_X64`? Or is there more to it?

carlpett (commented):
This is amazing! :tada:
Some thoughts/questions/opinions inline :)

.github/workflows/ci.yml
breed808: I've just tested and no, this must have been an Appveyor quirk. I've fixed this in a recent push.

.github/workflows/ci.yml
breed808: That's correct, it's no longer required. I've removed it in a recent push.

.github/workflows/ci.yml
breed808: It looks like [setup-go](https://github.com/actions/setup-go) is the reccommended way to use the 1.17 installation.

I've added the action to the CI file in a recent push.

.github/workflows/ci.yml
breed808: I think I have it now, `gh release upload` has been used in the `Release` step.

breed808:
@carlpett this is ready for a re-review when you have time

carlpett:
Looks fantastic, great work! :clap:
Regarding the two differences:
- I think the checksums (aside from being "good hygiene") are depended upon by downstream packaging, eg Chocolatey. Was there some difficulty adding it? It looks like we still generate them, at least.
- Didn't understand the second part - there were .exe:s in the "demo" release?

breed808:
I've added the checksums file to the release uploads in a recent push. You were right, they're still being generated, so it was simple to add them.

> Didn't understand the second part - there were .exe:s in the "demo" release?

The EXEs are available for releases, but are not publically available for individual builds. E.G. the artifacts from a PR (see https://github.com/prometheus-community/windows_exporter/actions/runs/1640572047) aren't publically available.

From memory the Appveyor builds for PRs made the EXEs publically available.

carlpett (approved):
Excellent work, happy new ~year~ CI! :fireworks: 
And year too, of course :)

breed808:
Happy New Year to you too!

Note we'll need to remove the Appveyor integration from this repository as the config removal isn't sufficient. I've checked the repository settings but don't seem to be able to remove it. Could you have a look?

You'll also be able to revoke any Appveyor tokens.

carlpett:
Hm, no, it looks like we'll need help from an organisation owner for that.
But I'm fine with merging this and just disabling AppVeyor for now, and any PRs can be "force merged" despite failing.

carlpett:
Old token invalidated now (so I don't accidentally build the release twice)

PR #901, opened 2021-12-30
Fix some promtool warnings

mjtrangoni:
Hi @breed808 @carlpett,

This fix all `non-histogram and non-summary metrics should not have "_count" suffix` issues reported by `promtool`. See #659.

Just let me know if I should proceed with the other reported issues.


carlpett (commented):
Looks good to me as such, but to read naturally, the "subject" of the metrics should probably be in plural after the change. E.g. `handles`, `threads`, etc.

breed808:
I think a `promtool` check should be added to our CI as well, at least for the default collectors.

mjtrangoni:
@carlpett good point! I´ve made all metric plural now. PTAL.

carlpett (approved):
Thanks!

PR #902, opened 2022-01-01
Move textfile mtime metric from loop

### LINKED ISSUE #899: textfile collector is broken on 0.17.0
Hi,
after windows_exporter upgrade to 0.17.0  textfile collector is not working as expected.
Exporter is not displaying any metrics, only error message.

http://localhost:9182
```
windows_exporter
Metrics

(version=0.17.0, branch=master, revision=4d771d2bcedd4b29f0bf2bade1a45f5e0ee3198d)
```
http://localhost::9182/metrics
```

collected metric "windows_textfile_mtime_seconds" { label:<name:"file" value:"custom_metrics.prom" > gauge:<value:1.64086244e+09 > } was collected before with the same name and label values
```

PROM file:
```
#C:\Program Files\windows_exporter\textfile_inputs\custom_metrics.prom
custom_metric_1 0
custom_metric_2 1
```

breed808:
Loop was erroneously creating duplicate `windows_textfile_mtime_seconds`
metrics, causing the exporter to return a HTTP 500 error and no metrics
from any collector.

Resolves #899

### DIFF
--- a/collector/textfile.go
+++ b/collector/textfile.go
@@ -313,9 +313,10 @@ fileLoop:
 	} else {
 		for _, mf := range metricFamilies {
 			convertMetricFamily(mf, ch)
-			c.exportMTimes(mtimes, ch)
 		}
 	}
+
+	c.exportMTimes(mtimes, ch)
 
 	// Export if there were errors.
 	ch <- prometheus.MustNewConstMetric(


carlpett (approved):

breed808:
Thanks! Would it be worth cutting a new release (E.G. `v0.17.1`) to make this fix available?

carlpett:
Yes, I think so. I can do it right away.

breed808:
Sure, go for it

PR #905, opened 2022-01-02
Fix counter metrics should have "_total" suffix issue

mjtrangoni:
Hi @breed808 @carlpett,

This fix all `counter metrics should have "_total" suffix` issues reported by promtool. See #659.


breed808 (approved):
Thanks for this!

PR #907, opened 2022-01-03
Check default collector metrics with promtool

breed808:


.github/workflows/ci.yml
```diff
@@ -25,8 +27,15 @@ jobs:
...
-          go get -u github.com/prometheus/promu@v0.11.1
-          go get -u github.com/josephspurrier/goversioninfo/cmd/goversioninfo@v1.2.0
+          # promtool brings in a lot of dependencies, slowing down compilation via `go get` and `go install`, so
+          # download the binary directly and extract.
+          # go get github.com/prometheus/prometheus/cmd/promtool@main
+          Invoke-WebRequest -Uri https://github.com/prometheus/prometheus/releases/download/v$($Env:PROMTOOL_VER)/prometheus-$($Env:PROMTOOL_VER).windows-amd64.zip -OutFile prometheus-$($Env:PROMTOOL_VER).windows-amd64.zip
+          Expand-Archive -Path prometheus-$($Env:PROMTOOL_VER).windows-amd64.zip -DestinationPath .
+          Copy-Item -Path prometheus-$($Env:PROMTOOL_VER).windows-amd64\promtool.exe -Destination "$(go env GOPATH)\bin"
+
+          go install github.com/prometheus/promu@v0.11.1
```
mjtrangoni, line 37: It would be nice to make promu version a variable, and update to 0.13.0. See [here](https://github.com/prometheus/promu/releases/tag/v0.13.0)

mjtrangoni:
@breed808 
Wouldn't be more interesting for the `promtool` case to run all possible collectors and not only the default enabled ones? I was thinking about a positive list of enabled collectors, so you would have more coverage than only the default enabled collectors.
And a non-related question. Why is the memory collector disabled by default?

.github/workflows/ci.yml
breed808: Good idea, I've add that in another commit

breed808:
A number of the collectors require Windows Features to be installed and/or configured, which may require unsupported reboots of the build environment.
Examples of this would include the `ad`, `dfsr` and `container` collectors, and I'm sure there are more.

We could add more collectors to this check, which would help prevent metric naming errors, but we'll need to make sure that any enabled collectors will work on a vanilla Windows Server install.

Looking at the history, the `memory` collector has never been enabled by default. I think it would be worth adding this (perhaps in a separate PR) as most users would expect this to be enabled by default, and the collector is rather light on resources.

carlpett (approved):
Nice! I'm not sure if it might make Github parallelise runs if we run `promtool` as a separated step from the end-to-end tests?

breed808:
That's a good point, I went with the easiest method of adding this to the e2e script.
Adding another job & script for the promtool invocation would speed up CI runs from increased parallelization. I'll get that implemented before merging.

breed808:
After a number of iterations, I have it separated to a new job and script.
Text processing is rather painful with Powershell, which has resulted in some complexity in the script.

Have a re-review and let me know what you think.

carlpett (approved):
Wow, that was a lot more code than I would have expected!
But let's do it this way for now, I think.

breed808:
> Wow, that was a lot more code than I would have expected!

That's my view on this too; I tried a number of methods unsuccessfully before reaching this point.
Hopefully I've missed a simple method that someone can point out in the future...

PR #909, opened 2022-01-04
Add more ADFS metrics from `AD FS` CounterSet

akrauza:
Signed-off-by: Austin D. Krauza <krauza.austin@gmail.com>

Reformat adfsCollector struct

Signed-off-by: Austin D. Krauza <krauza.austin@gmail.com>

Add metrics to ADFS collector documentation

Signed-off-by: Austin D. Krauza <krauza.austin@gmail.com>

Update ADFS collector with useful queries and links to documentation

Signed-off-by: Austin D. Krauza <krauza.austin@gmail.com>

Remove bad table formatter

Signed-off-by: Austin D. Krauza <krauza.austin@gmail.com>

Reformat ADFS collector using gofmt

Signed-off-by: Austin D. Krauza <krauza.austin@gmail.com>

Fix ADFS Config and Artifact DB Query time metrics

Signed-off-by: Austin D. Krauza <krauza.austin@gmail.com>

### DIFF
--- a/collector/adfs.go
+++ b/collector/adfs.go
@@ -5,6 +5,7 @@ package collector
 
 import (
 	"github.com/prometheus/client_golang/prometheus"
+	"math"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


collector/adfs.go
```diff
@@ -185,5 +442,197 @@ func (c *adfsCollector) Collect(ctx *ScrapeContext, ch chan<- prometheus.Metric)
...
+		c.configDBFailures,
+		prometheus.CounterValue,
+		adfsData[0].ConfigDBFailures,
+	)
+
+	ch <- prometheus.MustNewConstMetric(
+		c.avgConfigDBQueryTime,
+		prometheus.GaugeValue,
+		adfsData[0].AvgConfigDBQueryTime*math.Pow(10, -8),
+	)
```
breed808, line 630: Metric name and description don't match the documentation. Documentation states this metric is the duration of the most recent query, while the metric in the code indicates an average of all queries.

collector/adfs.go
```diff
@@ -185,5 +442,197 @@ func (c *adfsCollector) Collect(ctx *ScrapeContext, ch chan<- prometheus.Metric)
...
+		c.artifactDBFailures,
+		prometheus.CounterValue,
+		adfsData[0].ArtifactDBFailures,
+	)
+
+	ch <- prometheus.MustNewConstMetric(
+		c.avgArtifactDBQueryTime,
+		prometheus.GaugeValue,
+		adfsData[0].AvgArtifactDBQueryTime*math.Pow(10, -8),
+	)
```
breed808, line 618: Similar situation as with the `avgConfigDBQueryTime` metric description.

breed808 (commented):
Apologies for missing this Austin! I must have erroneously marked this as "read" without responding.

I've identified a discrepancy with two of the new metrics, have a look into those two and then we should be OK to merge :+1:

collector/adfs.go
akrauza: I believe this is another one of those things where Perflib is doing something behind the scenes to the value. If you graph out that value over time (top graph below), it is ever increasing and only resets during a reboot, which would make me guess that perflib is summing the value, rather than showing an average. My intuition tells me that it's not taking 5 seconds to read from the database, and 5-10ms seems to be more realistic (when `increase` is applied).

Perhaps we should make it `_total` so that way folks know to wrap it in an `increase()` (below graph).

image

breed808: Yes, exposing the metrics as a counter with the `_total` suffix would be the correct option here.
Perflib can be "interesting" in that you can't be entirely certain if a metric is a gauge or counter until directly observing it.

akrauza: Updated metrics and description to `db_config_query_time_seconds_total` and `Accumulator of time taken for a configuration database query` respectively.

collector/adfs.go
```diff
@@ -96,21 +129,245 @@ func newADFSCollector() (Collector, error) {
...
+			"Total number of failures connecting to the configuration database",
+			nil,
+			nil,
+		),
+		avgConfigDBQueryTime: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "db_config_query_time_seconds_total"),
+			"Last duration for a configuration database query",
+			nil,
+			nil,
+		),
```
breed808, line 317: Description needs to be updated to state this is the total time of all queries.

breed808 (commented):
Thanks Austin, just one minor change to the metric descriptions and we're good to go.

collector/adfs.go
akrauza: Updated

collector/adfs.go
```diff
@@ -96,21 +129,245 @@ func newADFSCollector() (Collector, error) {
...
+			"Total number of failures connecting to the artifact database",
+			nil,
+			nil,
+		),
+		avgArtifactDBQueryTime: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "db_artifact_query_time_seconds_total"),
+			"Last duration for an artifact database query",
+			nil,
+			nil,
+		),
```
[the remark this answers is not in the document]
akrauza: Updated

breed808 (approved):
Excellent! Thanks for making the requested changes for this, it's much appreciated.

akrauza:
Anytime!!

PR #913, opened 2022-01-06
Fix memory collector promtool metric issues

mjtrangoni:
Hi @carlpett @breed808,

This fixes some issues reported by `promtool`. See #659. 
Besides of that it extends the collector documentation from [wutils.com](https://wutils.com/wmi/root/cimv2/win32_perfformatteddata_perfos_memory/). Feel free to adjust any wording issue.

NOTE2: After merging this, we could make the collector default enabled.

### DIFF
--- a/collector/memory.go
+++ b/collector/memory.go
@@ -77,7 +77,8 @@ func NewMemoryCollector() (Collector, error) {
 		),
 		CacheFaultsTotal: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "cache_faults_total"),
-			"(CacheFaultsPersec)",
+			"Number of faults which occur when a page sought in the file system cache is not found there and must be retrieved from elsewhere in memory (soft fault) "+
+				"or from disk (hard fault) (Cache Faults/sec)",
 			nil,
 			nil,
 		),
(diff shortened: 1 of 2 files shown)


carlpett:
Hi @mjtrangoni,
Thanks again for these clean-ups! In this case, I'm unsure if the right issue is being fixed - I think actually the error might be setting the metrics as gauges. Could you do a quick verification that the data doesn't seem "counter-like"?

mjtrangoni:
Hi @carlpett,

You are right, these should be `counter`,
```
windows_memory_demand_zero_faults_total
windows_memory_cache_faults_total
windows_memory_page_faults_total
windows_memory_pool_nonpaged_allocs_total
windows_memory_pool_paged_allocs_total
windows_memory_swap_page_operations_total
windows_memory_swap_page_reads_total
windows_memory_swap_page_writes_total
windows_memory_swap_pages_read_total
windows_memory_swap_pages_written_total
windows_memory_transition_faults_total
windows_memory_transition_pages_repurposed_total
windows_memory_write_copies_total
```

And this should be keep as `gauge` removing the "_total" suffix,
```
windows_memory_pool_nonpaged_bytes_total
```

I will adjust my PR. Thank you for the hint!

mjtrangoni:
@carlpett fixed. PTAL

mjtrangoni:
@carlpett @breed808 ping?

carlpett (approved):
Thanks!
@breed808 - the one of us who makes the next release should try to remember that this is a breaking change

carlpett:
Hm, looks like CI is failing. I doubt it is directly related, though, looks like an internal error in the linter? @breed808 I see you have an open branch which at least succeeded on its' last run, is that something you think would fix this?

breed808:
@carlpett there's a couple issues with the CI that will be fixed in #970. That branch has passed CI.

PR #945, opened 2022-02-18
Run CI on pushes to master branch

### LINKED ISSUE #940: Builds are not triggered on main branch so container image isn't produced
In https://github.com/prometheus-community/windows_exporter/pull/864 some logic was added to build a "bleeding edge" container:

https://github.com/prometheus-community/windows_exporter/blob/5106b829c3092ef3f1346790336ba0789ef30c01/.github/workflows/ci.yml#L180-L185

This doesn't trigger currently because the workflow only runs on PR's and tags:

https://github.com/prometheus-community/windows_exporter/blob/5106b829c3092ef3f1346790336ba0789ef30c01/.github/workflows/ci.yml#L5-L12

Adding a trigger like 

```
on:
  push:
    branches: [ master ]
``` 

would build, test and then push the bleeding edge image as well give a signal that main branch is healthy after the PR merge. @breed808 @carlpett  any thoughts on this?

breed808:
Required to keep container image (with "latest" tag) up to date. Will
also ensure the master branch is healthy after PR merges.

Resolves #940

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,11 @@
 name: windows_exporter CI/CD
 
-# Trigger on pull requests and releases
+# Trigger on pull requests, releases and pushes to master branch.
 # Deployments will only occur for releases (see `if` clauses in the build job).
 on:
+  push:
+    branches:
+      - master
   pull_request:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


jsturtevant:
lgtm

breed808:
@jsturtevant I've amended the commit to perform the GCR login on push as well, preventing the image push from failing on a `master` build.

jsturtevant:
good catch!

breed808:
Apologies for the notification spam from all the commits/pushes. I've identified a timing issue with the Promtool job, where the script queries the exporter before it is ready.

I've also set the Promtool job to run the exporter on a different port, in the event that both exporters (e2e and promtool) are running at the same time.

breed808:
The promtool job failures are affecting other PRs, such as #943, so I'll merge this one now.
@carlpett when you get to this in your notifications, let me know if this needs revisiting.

PR #948, opened 2022-02-27
Fix image build issues

### LINKED ISSUE #940: Builds are not triggered on main branch so container image isn't produced

breed808:
`master` image builds have been failing due to invalid characters in the tag name (Build step) and incorrect environment variable setting (Push Latest image step).

Resolves #940


carlpett (approved):
LGTM!

Issue #923, opened 2022-01-21
failed collecting os metrics


xll970211:
i run win11 and windows server 2008 is ok,but run in windows server 2012 has error
```
time="2022-01-21T15:48:57+08:00" level=error msg="Error reading textfile collect
or directory \"C:\\\\Users\\\\Administrator\\\\Desktop\\\\textfile_inputs\": ope
n C:\\Users\\Administrator\\Desktop\\textfile_inputs: The system cannot find the
 file specified." source="textfile.go:249"
time="2022-01-21T15:48:57+08:00" level=error msg="failed collecting os metrics:<
nil> The system cannot find the file specified." source="os.go:137"
time="2022-01-21T15:48:57+08:00" level=error msg="collector os failed after 0.00
0980s: The system cannot find the file specified." source="exporter.go:207"
time="2022-01-21T15:49:02+08:00" level=error msg="Error reading textfile collect
or directory \"C:\\\\Users\\\\Administrator\\\\Desktop\\\\textfile_inputs\": ope
n C:\\Users\\Administrator\\Desktop\\textfile_inputs: The system cannot find the
 file specified." source="textfile.go:249"
time="2022-01-21T15:49:02+08:00" level=error msg="failed collecting os metrics:<
nil> The system cannot find the file specified." source="os.go:137"
time="2022-01-21T15:49:02+08:00" level=error msg="collector os failed after 0.00
1958s: The system cannot find the file specified." source="exporter.go:207"
```

yuanpengchao:
you need run mkidr `textfile_inputs` at C:\Users\Administrator\Desktop\

Our Windows English server is running properly, Chinese server also has this error:
failed collecting os metrics:<
nil> The system cannot find the file specified." source="os.go:137

xll970211:
thank you very much.i I tried what you said mkdir textfile_inputs. The result is the same as you said
failed collecting os metrics:<
nil> The system cannot find the file specified." source="os.go:137
but we run in windows server 2008R2 Chinese，It won't happen
What should I do? @yuanpengchao

yuanpengchao:
@xll970211  I'm looking for the answer to that question, too ……囧

csrlsm:
I use release 0.171 and the OS windows server 2012 , there is the same issue. 
but in windows server 2019 it works

> i run win11 and windows server 2008 is ok,but run in windows server 2012 has error
> 
> ```
[...]

PR #954, opened 2022-03-12
Continue OS collector on absent paging file

### LINKED ISSUE #939: os collector fails when paging file is disabled (>= v0.17.0)
Having the Windows paging file disabled on all volumes results in the "os" collector failing on v0.17.0 or newer with below error.
Either running an older version (i.e. v0.16.0) or setting a paging file again (w/ system reboot for some reason) mitigates this, but would be great if this could be fixed in the exporter/collector again in an upcoming version.
Tested on both Windows 10 LTSC 2021 and Windows Server 2022.
```
failed collecting os metrics:<nil> The system cannot find the file specified.
collector os failed after 0.003015s: The system cannot find the file specified.
```

breed808:
Inactive or missing paging file should not be cause for OS collector to
fail.
Instead, log the error and continue with OS collection.

Resolves #939

### DIFF
--- a/collector/os.go
+++ b/collector/os.go
@@ -179,11 +179,7 @@ func (c *OSCollector) collect(ctx *ScrapeContext, ch chan<- prometheus.Metric) (
 	if err != nil {
 		return nil, err
 	}
-	pagingFiles, _, err := memManKey.GetStringsValue("ExistingPageFiles")
-	if err != nil {
-		return nil, err
-	}
-
+	pagingFiles, _, pagingErr := memManKey.GetStringsValue("ExistingPageFiles")
 	// Get build number and product name from registry
 	ntKey, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\Windows NT\CurrentVersion`, registry.QUERY_VALUE)
 	defer ntKey.Close()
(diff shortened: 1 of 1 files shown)


breed808:
Omitting the metric may reduce confusion as to why a paging file is 0 bytes in size. I'll amend the PR :+1:

collector/os.go
```diff
@@ -261,12 +257,21 @@ func (c *OSCollector) collect(ctx *ScrapeContext, ch chan<- prometheus.Metric) (
 		timezoneName,
 	)
 
-	ch <- prometheus.MustNewConstMetric(
-		c.PagingFreeBytes,
-		prometheus.GaugeValue,
-		pfb,
-	)
-
+	if pagingErr == nil {
```
petemounce, line 260: Could this be handled closer to the error return possibility 80 lines higher? There's been another, different registry access since; I would find it easier to follow the flow with less to read.

(Obviously, I'm also random person on the internet; I hope review comments are welcome. I'm coming from #923 where I tried to understand an error from near here and haven't yet)

collector/os.go
```diff
@@ -261,12 +257,21 @@ func (c *OSCollector) collect(ctx *ScrapeContext, ch chan<- prometheus.Metric) (
...
+			pfb,
+		)
+
+		ch <- prometheus.MustNewConstMetric(
+			c.PagingLimitBytes,
+			prometheus.GaugeValue,
+			fsipf,
+		)
+	} else {
+		log.Debugln("Could not find SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Memory Management key. windows_os_paging_free_bytes and windows_os_paging_limit_bytes will be omitted.")
```
petemounce, line 273: Could the fact that this is in the hklm hive please be reflected into the log message? There are several to choose from when verifying the system state vs code behaviour.

```suggestion
		log.Debugln("Could not find HKLM:\SYSTEM\\CurrentControlSet\\Control\\Session Manager\\Memory Management key. windows_os_paging_free_bytes and windows_os_paging_limit_bytes will be omitted.")
```

Or extract a variable for the paths as well?

petemounce (commented):
Thanks for your efforts to fix, I really appreciate it! I hope my comments are welcome; I'm not a maintainer or anything.

collector/os.go
breed808: I agree that the flow of the collector is made more difficult by the change, but the `pagingFiles` variable is required shortly afterwards to populate the `fsipf` and `pfb` variables, while the metric exports to the channel all occur at the same time to improve readability.

So I don't think it's worth trying to rearrange the `collect` function to group `pagingErr` with the paging metrics.

collector/os.go
breed808: That's a good idea, I'll add `HKLM` to the log message.

Issue #695, opened 2021-01-09
question


GhasemSystem:
Hi
Why we not implement collector for Win32_PerfRawData_Counters_* & ASP yet?
maybe add this collectors in the future?

GhasemSystem:
hello!

breed808:
Hi Ghasem,

Could you specify the PerfRawData counters that you need? If there's a corresponding Perflib source, adding the metrics should be straightforward.
The addition of new WMI collectors is something we'd like to avoid, as they bring performance and stability issues.

GhasemSystem:
Hi @breed808 

for example:

Win32_PerfRawData_Counters_HTTPServiceRequestQueues ---> CurrentQueueSize
Win32_PerfFormattedData_ASPNET*                                      ---> Requests Queued, ...

breed808:
`HTTPServiceRequestQueues` looks like it's an IIS metric and should [be available via Perflib](https://techcommunity.microsoft.com/t5/iis-support-blog/performance-counters-for-monitoring-iis/ba-p/683389). I should be able to add this metric using one of our IIS servers to gather information.

`ASPNET` appears to also have [performance metrics](https://stackify.com/asp-net-performance-counters/), but I don't have any running ASP.NET applications running that could be used to test the metric(s).

If you're able to provide performance counter information yourself, I should be able to create a collector for you. See this for details on gathering the required info: https://github.com/prometheus-community/windows_exporter/issues/666#issuecomment-732809777

Issue #641, opened 2020-10-26
Support HTTP.sys Performance Counters in IIS collector


iinuwa:
Thanks for this tool!

I noticed that some of the [HTTP.Sys performance counters](https://docs.microsoft.com/en-us/windows/win32/http/scenario-3--performance-counters) (HTTP Service, HTTP Service Request Queues, and HTTP Service Url Groups) are not collected by the IIS collector. Would it be possible to add them?

I'm not sure whether some of these have already been implemented. `windows_iis_worker_uri_cache_*` may cover the HTTP Service Counters. The specific counter I'm looking for is HTTP Service Request Queues/CurrentQueueSize.

GhasemSystem:
How resolve it?
I need to "current queue size" value and other metric from HTTP.Sys

carlpett:
Hi @GhasemSystem,
The fastest way would be to contribute the code. We'd be happy to review a pull request! Otherwise, it's on our radar, and we'll implement it when we have spare time to do so.

tweshlyallpuri:
Hi Carlpett,

Do we have any plans for this request currently?

breed808:
@tweshlyallpuri not currently. If you're able to contribute the code we'd be happy to review a pull request.

breed808:
> we are checking if we can assign a dev to support this feature.

Do you mean one of your own developers? If so then you're more than welcome to submit a Pull Request.

samisevenx00:
did we have any news about this request ? next release maybe ? we are waiting a lot for this feature.

breed808:
Looking at the linked pull request it appears work on this has stalled.

You're welcome to contribute to the pull request or submit your own. Let me know if you have any queries on contributing.

I may have some time in the next couple weeks to address some of the concerns raised by Calle in the pull request, but would not be able to properly test the collector.

PR #959, opened 2022-03-24
Add HTTP Service Request Queues

MarcDenman:
I have exposed `CurrentQueueSize` and `RejectedRequest` from the `HTTP Service Request Queues`. As always, the hardest part is naming things. I was unsure what to the name metric as I didn't want requests to be in the twice so have tried to give them a reasonable name.

The only oddity I have encountered is that it exposes an instance name which I can't tie back to an AppPool `---1` and I have no idea what it is, any ideas? Should I filter it out?

image

#695 #641

### DIFF
--- a/collector/iis.go
+++ b/collector/iis.go
@@ -14,7 +14,7 @@ import (
 )
 
 func init() {
-	registerCollector("iis", NewIISCollector, "Web Service", "APP_POOL_WAS", "Web Service Cache", "W3SVC_W3WP")
+	registerCollector("iis", NewIISCollector, "Web Service", "APP_POOL_WAS", "Web Service Cache", "W3SVC_W3WP", "HTTP Service Request Queues")
 }
 
 var (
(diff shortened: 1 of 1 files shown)


carlpett (commented):
Hi @MarcDenman,
Thanks for the PR! I think the naming looks good (aside from the type change).
Regarding the odd app name, I found the docs [here](https://docs.microsoft.com/en-us/windows/win32/http/scenario-3--performance-counters) which has this passage:
>Only one instance of the HTTP Server API counters exists per machine, as these counters represent the component-wide state. With instances of Url Group performance counters, the instance ID (for the performance counter) will match the Url Group ID. The Url Group ID can be viewed by running `netsh http show servicestate`. With instances of Request Queues performance counters, the instance name corresponds to Request Queue Name. The Request Queue Name (if one exists) can be shown by the same `netsh http show servicestate`. However, some server applications may have unnamed Request Queues that cannot be matched to a performance counter instance ID.

Could this be the case here? We'll probably want to merge any triple-dashed ones, or what do you think?

jpuskar:
I'd love to see this in!

PR #967, opened 2022-04-05
Add an option to remove .prom files after consuming them in the textfile collector.

malkia:
I wanted an option to automatically remove .prom files after they are consumed (even if they fail).

The implementation renames .prom files to .prom.stale before reading them.
In the next (and the before the first) collect cycle these would be removed.

### DIFF
--- a/collector/textfile.go
+++ b/collector/textfile.go
@@ -41,6 +41,11 @@ var (
 		"Directory to read text files with metrics from.",
 	).Default(getDefaultPath()).String()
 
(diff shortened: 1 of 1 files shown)


malkia:
I'll need to figure out git signing, before I do something about this. Sorry for the noise.

breed808:
It's OK, no need to close and re-submit. If you're using the `git` CLI tool, you can sign the current commit with `git commit --amend -s`. The `-s` flag will sign the commit.

PR #978, opened 2022-04-25
translate process name to service name

trippinnik:
when the flag collector.process.translate is used will take monitored processes and check if it is a service. if it is a service the name of service is used for the metric label instead of the processname

### DIFF
--- a/collector/process.go
+++ b/collector/process.go
@@ -28,6 +28,11 @@ var (
 		"collector.process.blacklist",
 		"Regexp of processes to exclude. Process name must both match whitelist and not match blacklist to be included.",
 	).Default("").String()
+	// Add parameter to optionally translate process names to service names ***
+	translateProcess = kingpin.Flag(
+		"collector.process.translate",
+		"If true, translate process names to service names where applicable.",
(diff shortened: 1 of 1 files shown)


carlpett:
Hi @trippinnik,
With the current implementation, I would suspect this is _really_ expensive in terms of run time of the collector, since we'd fan out with a WMI query for each process? Those can take tens or hundreds of milliseconds each, and a lot of systems would have hundreds of processes, so we'd end up with a collection time of several seconds for just this collector.

I also don't think it's unlikely that you could have many processes that map to a single service, and then you'd end up with an invalid scrape due to duplicated labels.

Perhaps it'd be good to start with a discussion of why this would be a good feature, and look at implementation strategies afterwards?

trippinnik:
> 

@carlpett The feature is centered around monitoring the services beyond up down.  The process the service runs as doesn't match what support/ops teams are looking for when focused on the service and we need a way to do that.

trippinnik:
The process that need to be looked at are not able to be mapped to which service launched the process and therefore unable to appropriately remediate

PR #985, opened 2022-05-13
Only log missing IIS instances on debug output

### LINKED ISSUE #975: Need to convert the error log to at least warning if not info - no instances found in W3SVC_W3WP or W3SVC_W3WP_IIS8 - skipping collection
The following error logs Cluttering windows event logs - Need to convert the error log to at least warning if not info
1. no instances found in W3SVC_W3WP - skipping collection.
2. no instances found in W3SVC_W3WP_IIS8 - skipping collection.
File iis.go - line numbers 1344 and 1600

breed808:
Previous output was spamming STDOUT and event logs. Default log level
is `info`, so `debug` level was chosen.

Resolves #975

### DIFF
--- a/collector/iis.go
+++ b/collector/iis.go
@@ -1341,7 +1341,7 @@ func (c *IISCollector) collectW3SVC_W3WP(ctx *ScrapeContext, ch chan<- prometheu
 		pid := workerProcessNameExtractor.ReplaceAllString(app.Name, "$1")
 		name := workerProcessNameExtractor.ReplaceAllString(app.Name, "$2")
 		if name == "" {
-			log.Error("no instances found in W3SVC_W3WP - skipping collection")
+			log.Debug("no instances found in W3SVC_W3WP - skipping collection")
 			break
 		}
 		if name == "_Total" ||
@@ -1597,7 +1597,7 @@ func (c *IISCollector) collectW3SVC_W3WP(ctx *ScrapeContext, ch chan<- prometheu
 			pid := workerProcessNameExtractor.ReplaceAllString(app.Name, "$1")
 			name := workerProcessNameExtractor.ReplaceAllString(app.Name, "$2")
 			if name == "" {
-				log.Error("no instances found in W3SVC_W3WP_IIS8 - skipping collection")
+				log.Debug("no instances found in W3SVC_W3WP_IIS8 - skipping collection")
 				break
 			}
 			if name == "_Total" ||


carlpett:
I could agree error might be a bit strong log level, but I'm not sure we should be effectively suppressing this. This is caused either by a misconfiguration of the exporter, or the OS/IIS instance misbehaving. We can't know which it is, so the prudent thing to do is to notify the user (logging the problem on each scrape is also how other major exporters in the ecosystem deal with it).
Typically, this is caused by users enabling all exporters, rather than using configuration management tools to enable the relevant ones.

breed808:
> logging the problem on each scrape is also how other major exporters in the ecosystem deal with it

That's rather helpful context for this situation. I was hopeful there was a middle-ground option available, such as logging an error once (on collector startup, or tracking an error state) rather than the log-per-collection currently in place.

It's a fair point that this is expected to be managed with a configuration management tool. I'll close the PR.

aa-matthias:
We're also experiencing this and it's cluttering the logs. @carlpett I'm not aware of of falsy configuration in our case. IIS is up and running (DefaultAppPool, Default Web Site) and we're collecting metrics for the instance but still are receiving those log entries in the windows event log.

Service config:
```cmd
"C:\Program Files\windows_exporter\windows_exporter.exe" --log.format logger:eventlog?name=windows_exporter --collectors.enabled os,iis,cpu,service,mssql,memory,logical_disk,net,cs,vmware      --collector.iis.site-whitelist "Default Web Site"
```

carlpett:
Hm, odd. Let me look through this again.

carlpett:
Okay, I probably read through this a bit too fast, sorry about that. I'm not sure that entire if-statement is even correct? Why would we bail on the entire loop if one of the entries are nameless, rather than just add it as another condition on the if below (where we skip _Total and white-/blacklisted sites)?

@aa-matthias What counter instances do you have available under `W3SVC_W3WP`? I assume empty string is one of them, but are there others?

aa-matthias:
@carlpett for me there aren't any different ones - as commented in the issue - executing the two following scripts:

https://www.activexperts.com/admin/scripts/wmiscripts/powershell/1258/
```ps1
Name: _Total

########

Name: Default Web Site

########
```

https://www.activexperts.com/admin/scripts/wmiscripts/powershell/1259/

```ps1
Name: 

########
```

chavalikesavanand:
was this checked in? any updates please? in the main repo i still see this as "error log"

PR #986, opened 2022-05-14
Add `user_input_delay` collector

### LINKED ISSUE #958: Feature Request:  Support for User Input Delay Performance Counters
Starting in Windows 2019 and higher Microsoft has released some new performance counters for Remote Desktop Session Hosts: https://docs.microsoft.com/en-us/windows-server/remote/remote-desktop-services/rds-rdsh-performance-counters

The ability to capture data for "User Input Delay per Process" and User Input Delay per Session(Max and Average)" 

The CimClasses involved are

- Win32_PerfRawData_LSM_UserInputDelayperProcess
- Win32_PerfFormattedData_LSM_UserInputDelayperProcess
- Win32_PerfRawData_LSM_UserInputDelayperSession
- Win32_PerfFormattedData_LSM_UserInputDelayperSession

breed808:
This MR adds metrics for user input delays, for both sessions and processes.

Unfortunately the documentation I've found makes no mention of the unit of time these metrics represent.

Resolves #958


carlpett (commented):
Some minor naming comments,  but overall looks good. 
I wonder if this is sufficiently stand-alone to warrant it's own collector, or if it could be a subcollector in `terminal_services` or `remote_fx` collectors that both deal with remote desktop connections?

collector/user_input_delay.go
```diff
@@ -0,0 +1,134 @@
...
+		SessionMaxInputDelay: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "session"),
+			"Maximum value for queuing delay across all user input waiting to be picked-up by any process in the session during a target time interval",
+			[]string{"session_id"},
+			nil,
+		),
+		ProcessMaxInputDelay: prometheus.NewDesc(
+			prometheus.BuildFQName(Namespace, subsystem, "process"),
+			"Maximum value for queuing delay across all user input waiting to be picked-up by the process during a target time interval",
+			[]string{"sid", "pid", "pname"},
```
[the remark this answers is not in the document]
breed808: Good spot, I've amended the label names to keep things consistent.

collector/user_input_delay.go
```diff
@@ -0,0 +1,134 @@
...
+// Process name provided by "User Input Delay per Process" ListSet is malformed and needs to be split.
+// E.G. "1:7232 <TextInputHost.exe>" could be split to:
+//  session_id: 1
+//  pid: 7232
+//  pname: TextInputHost.exe
+func splitProcessLabel(process_label string) (*processLabels, error) {
+	// Use regex in place of multiple splits
+	query := regexp.MustCompile(`(\d):(\d+)\s+<(.*)>`)
+	if !query.MatchString(process_label) {
+		return nil, errors.New("Unexpected process label structure")
```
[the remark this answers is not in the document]
breed808: Done, I've added it at the end of the error message.

collector/user_input_delay.go
```diff
@@ -0,0 +1,134 @@
...
+	MaxInputDelay float64 `perflib:"MaxInputDelay"`
+}
+
+type processLabels struct {
+	sessionID    string
+	pid          string
+	process_name string
+}
+
+// Process name provided by "User Input Delay per Process" ListSet is malformed and needs to be split.
```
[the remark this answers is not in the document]
breed808: Poor choice of words. The name provided by Perflib isn't particularly useful, as the pid, sessionid and process name combination ensures most would be unique.

I've modified the comment wording slightly, let me know if it's clear enough.

breed808:
> I wonder if this is sufficiently stand-alone to warrant it's own collector, or if it could be a subcollector in terminal_services or remote_fx collectors that both deal with remote desktop connections?

I was considering this too. This would be the smallest collector in the exporter, once merged, with only two metrics.

If users of this collector will always be running the `terminal_services` or `remote_fx` collectors, I'd say it'd be worth moving these metrics to one of those collectors.

collector/user_input_delay_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+func BenchmarkUserInputDelayCollector(b *testing.B) {
+	benchmarkCollector(b, "user_input_delay", NewUserInputCollector)
+}
+
+func TestSplitProcessLabel(t *testing.T) {
+	data := "1:1268 <svchost.exe>"
+	expectedOutput := processLabels{
+		sessionID:   "1",
+		pid:         "1268",
```
carlpett, line 16: Quite nice annotation from the test failure here!

collector/user_input_delay.go
```diff
@@ -0,0 +1,135 @@
...
+		if err != nil {
+			return nil, err
+		}
+		ch <- prometheus.MustNewConstMetric(
+			c.ProcessMaxInputDelay,
+			prometheus.GaugeValue,
+			process.MaxInputDelay,
+			labels.sessionID,
+			labels.processID,
+			labels.processName,
```
carlpett, line 131: What sort of cardinality does this result in, btw? Something we need to warn the user about?

collector/user_input_delay.go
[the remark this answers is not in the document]
carlpett: :+1:

carlpett:
> If users of this collector will always be running the `terminal_services` or `remote_fx` collectors, I'd say it'd be worth moving these metrics to one of those collectors.

Good question. I _think_ this would be something one would be interested in if running `terminal_services`, but should probably be checked with users.

collector/user_input_delay.go
breed808: Once of the labels is a process ID, so the cardinality will be quite high. I'll add a warning similar to the `process` and `service` collectors.

collector/user_input_delay_test.go
breed808: Very nice! I'll fix this in the next push.

collector/user_input_delay.go
breed808: Similar warning has been added, as well as the whitelist/blacklist flags from the `process` collector.

carlpett:
Bit of compilation errors, but looks good! Would be nice to figure out which other collector, if any, to merge it into before merging the PR, though. Otherwise we'll be stuck with it due to backwards compatibility, right?

PR #997, opened 2022-05-29
Update gopkg.in/yaml_v2 to yaml_v3

breed808:
Previous version was vulnerable to CVE-2022-28948

Signed-off-by: Ben Reedy <breed808@breed808.com>


carlpett (approved):
LGTM, was a bit hard to find a list of breaking changes between v2 and v3 but it _appears_ we aren't affected by anything.

PR #1036, opened 2022-07-29
Fix Error "Service Access is denied" in service collector useApi 

alvarocabanas:
Fix error preventing to collect certain restricted services when using the service collector in api mode "useApi".
Since the handler used sets by default SC_MANAGER_ALL_ACCESS when creating a service handler, in certain windows distributions and for certain services, it is not possible to open a connection to those services that are restricted.
Opening the service with GENERIC_READ fixes the issue.

This addresses https://github.com/prometheus-community/windows_exporter/issues/1037

Signed-off-by: Alvaro Cabanas <albanas@gmail.com>

### DIFF
--- a/collector/service.go
+++ b/collector/service.go
@@ -6,6 +6,7 @@ package collector
 import (
 	"fmt"
 	"strings"
+	"syscall"
 
 	"github.com/StackExchange/wmi"
 	"github.com/prometheus-community/windows_exporter/log"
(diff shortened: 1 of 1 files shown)


breed808:
Thanks for submitting this! It's always good to reduce required permissions of the exporter where possible.

I take it that this is following up on the work previously submitted by @carlossscastro in #812? I'm not familiar with the syscalls/APIs in use so I'm not sure if I can provide a meaningful review.

carlossscastro:
Hi @breed808,

That's exactly a follow up from the implementation I did in https://github.com/prometheus-community/windows_exporter/pull/812. After a couple of reports from users that not all the services were being monitored, we found out that GO standard library to handle windows services requires `SC_MANAGER_ALL_ACCESS` permissions which is not ideal.
So @alvarocabanas implemented an alternative way, with direct syscalls to overcome that.
From the tests we both did (we work together in the same team) there is no loss in performance (which was the main reason for https://github.com/prometheus-community/windows_exporter/pull/812) and it solves the permissions issue.
From my side, FWIW, this is reviewed and approved

breed808:
Thanks @carlossscastro, that context helps a lot here :+1:

breed808 (approved):

carlossscastro:
Thanks @breed808

PR #1045, opened 2022-08-23
fix Windows Service timeout during high CPU (eg. post Windows Update)

jammiemil:
This PR attempts to resolve a long standing [issue](https://github.com/prometheus-community/windows_exporter/issues/551) with the windows service timing out when CPU is constrained on the host machine.

The underlying cause of this is due to how late the service initialization happens in the code due to it existing in main() which is the last part of code to be run during startup as per:

image

What we attempt to do here is call the windows service as early as possible by placing it in a separate package's init() at the top of the main packages import list causing it to occur as early in the process as possible

Originally I had also moved the initiation of logging into this same package however i ran into issues with double logging in event viewer so ill revisit this at a later date. 

The benefit of having the logging called as early as possible is that you can properly debug issues such as this as until logging is initiated as part of main() in the main package none of the log commands can actually execute

### DIFF
--- a/collector/service.go
+++ b/collector/service.go
@@ -6,6 +6,7 @@ package collector
 import (
 	"fmt"
 	"strings"
+	"syscall"
 
 	"github.com/StackExchange/wmi"
 	"github.com/prometheus-community/windows_exporter/log"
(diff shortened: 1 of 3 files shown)


jammiemil:
As it stands this code will initiate a service super fast but i have subsequently broken the 'stop' of a service and am struggling to work out how best to make this work. The cause of this is the channel used to inform on service stop requests is not acted on in main(), only in the new initiate package, and im getting out of my depth :)

PR #1047, opened 2022-08-23
fix Windows Service timeout during high CPU (eg. post Windows Update)

jammiemil:
This PR attempts to resolve a long standing [issue](https://github.com/prometheus-community/windows_exporter/issues/551) with the windows service timing out when CPU is constrained on the host machine.
The underlying cause of this is due to how late the service initialization happens in the code due to it existing in main() which is the last part of code to be run during startup as per:

image
 
What we attempt to do here is call the windows service as early as possible by placing it in a separate package's init() at the top of the main packages import list causing it to occur as early in the process as possible
Originally I had also moved the initiation of logging into this same package however i ran into issues with double logging in event viewer so ill revisit this at a later date.
The benefit of having the logging called as early as possible is that you can properly debug issues such as this as until logging is initiated as part of main() in the main package none of the log commands can actually execute

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -52,7 +52,6 @@ type prometheusVersion struct {
 const (
 	defaultCollectors            = "cpu,cs,logical_disk,net,os,service,system,textfile"
 	defaultCollectorsPlaceholder = "[defaults]"
-	serviceName                  = "windows_exporter"
 )
 
 var (
--- a/exporter.go
+++ b/exporter.go
@@ -4,7 +4,7 @@
 package main
 
 import (
-	//We keep this at the top to ensure it happens first
+	//Its important that we do these first so that we can register with the windows service control ASAP to avoid timeouts
 	"github.com/prometheus-community/windows_exporter/initiate"
 	"github.com/prometheus-community/windows_exporter/log"
 
--- a/exporter.go
+++ b/exporter.go
@@ -4,6 +4,10 @@
 package main
 
 import (
+	//We keep this at the top to ensure it happens first
+	"github.com/prometheus-community/windows_exporter/initiate"
+	"github.com/prometheus-community/windows_exporter/log"
+
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


jammiemil:
As it stands this code will initiate a service super fast but i have subsequently broken the 'stop' of a service and am struggling to work out how best to make this work. The cause of this is the channel used to inform on service stop requests is not acted on in main(), only in the new initiate package, and im getting out of my depth :)

initiate/initiate.go
```diff
@@ -0,0 +1,60 @@
...
+	changes <- svc.Status{State: svc.Running, Accepts: cmdsAccepted}
+loop:
+	for {
+		select {
+		case c := <-r:
+			switch c.Cmd {
+			case svc.Interrogate:
+				changes <- c.CurrentStatus
+			case svc.Stop, svc.Shutdown:
+				log.Debug("Service Stop Received")
```
jammiemil, line 30: For anyone looking to help get this over the line, Currently when someone hits 'stop' on the windows service we execute up to here fine, but the overall process is not stopping, presumably because this happens out in exporter.go. If you then hit 'Stop' a second time this code does not fire, presumably because the service component here is closed down whilst the process component stays open.

 I know im doing something wrong here but im not sure what.

jammiemil: Came down to the definition of my channel being wrong

exporter.go
```diff
@@ -4,6 +4,10 @@
 package main
 
 import (
+	//We keep this at the top to ensure it happens first
```
breed808, line 7: It'd be worth commenting _why_ the ordering is important, rather than stating what's being done here.

initiate/initiate.go
```diff
@@ -0,0 +1,59 @@
+package initiate
```
breed808, line 1: A comment here describing the purpose of the package would be nice.

breed808 (commented):
This is looking nice! The change is mostly moving code out of the main exporter package, so the only improvement I can think of is additional documentation in the form of some comments.

Thanks for putting in the effort to resolve this issue :+1:

breed808 (approved):
Looking good :+1: 
I'm happy to merge if you're comfortable with your testing.

jammiemil:
I've tested this a fair bit by replacing the binary in a VM and the service certainly starts much quicker under load + service stop/start/restart functionality does not appear to be hampered.

The only thing i haven't tested i guess is installing it from scratch using the msi since i wasn't able to build this but i see no reason why that would be affected since that code hasn't been touched.

breed808:
I agree, the MSI installation shouldn't be affected.

PR #1051, opened 2022-08-30
Export extra version information

benridley:
Adds major version, minor version, and build number as independent
labels when exporting OS info.


breed808 (approved):
Looks good! Thanks for submitting this.

PR #1078, opened 2022-10-18
Try to bump /x/sys to the current stable version to fix IsWindowsService()

gillg:
Quick try to bump the lib /x/sys and using the latest "stable" version 0.1.0 instead of 0.0.0-snapshot


breed808 (approved):
Strange that Dependabot missed this one. Thanks for submitting this!

Issue #787, opened 2021-05-20
Properly understanding CPU utilization metrics


higels:
We noticed recently that on a reasonably loaded system (AMD 7702P, 128 hyper threads, 3.35Ghz at low load, will come down to 2.6Ghz as things heat up), there was a large discrepancy in overall and per-hyperthread utilization when comparing taskmgr vs. our windows_exporter metrics + the stuff we pull from WMI independently.

Specifically, taskmgr reported that the average utilization of the system was between 65 and 70%, with many individual cores being completely utilized, with their hyper thread siblings being between 10 and 30% utilized.

Our data from windows_exporter showed that `1 - avg(rate(windows_cpu_time_total{mode='idle'}[2m]))` was between 50 and 55%. Our data from the WMI gauge `\\Processor(_Total)\\% Processor Time` agreed with windows_exporter.

The data for individual hyperthreads showed two bands, one between 70 and 85% utilization and the other between 20 and 50%.

We updated our independently gathered metrics to use `\\Processor Information(_Total)\\% Processor Utility` and this seemed to line up with what taskmgr was providing for overall utilization. `% Processor Time` is seemingly very old and doesn’t handle variance in CPU frequency.

If we trust taskmgr, with windows_exporter / perflib, we have a 10-15% difference in overall system utilization at higher load as our CPU clocks down, and the per core utilization is being misallocated across each pair of hyperthreads.

This brings me to my questions:

1. Has anyone else observed this? It seems to happen on all of the various CPU types we have.
1. Even though it’s a gauge, would a PR which exposed ProcessorUtilityRate be accepted, since it seems to offer us something that IdleTimeSeconds does not?
1. What does processor_performance actually represent? I had hoped that it would help me scale idle time to show some kind of “performance lost to frequency or IPC decrease”, but all I can say for sure is that its rate is close to 0 when a system is idle, and close to 80M $somethings per second on a very busy hyperthread. The maximum seems to be around this number regardless of base CPU frequency.

carlpett:
There's been a number of reports of the data not lining up, but noone has done such a thorough investigation. Thanks for that!
So, first off, on Win2008R2+, we actually use `\\Processor Information(*)\*` counters, rather than `\\Processor(*)`, but I recall the overall data being similar.
Interestingly, we seem to added ProcessorUtilityRate so we extract the data, but don't expose an actual metric:
[...]

PR #1088, opened 2022-11-02
Add cpu metrics based on newer and more accurate perflib sources

higels:
This change adds 4 new CPU related metrics:

 * process_mperf_total
 * processor_rtc_total
 * processor_utility_total
 * processor_privileged_utility_total

and renames the existing process_performance to
processor_performance_total, since it was previously misunderstood and was unlikely to be have been useful without the above new metrics

The data sources for these are not particularly well understood, and the examples show that in some cases, arbitrary scaling factors are required to actually make them useful, but in my testing on hundreds of systems with a broad range of CPUs and operating systems from 2012r2 through to 2019 has proved out that we can use them to accurately display actual CPU frequencies and CPU utilisation as it is represented in taskmgr.

Things I don't particularly like and would like input on:

 * I would have preferred to do the scaling of processor_mperf_total in the code, but there isn't an elegant way of doing this right now.
 * Maybe processor_mperf_total should be called processor_mperformance_total.
 
See #787 for discussion.

### DIFF
--- a/collector/cpu.go
+++ b/collector/cpu.go
@@ -37,6 +37,10 @@ type cpuCollectorFull struct {
 	ProcessorFrequencyMHz    *prometheus.Desc
 	ProcessorMaxFrequencyMHz *prometheus.Desc
 	ProcessorPerformance     *prometheus.Desc
+	ProcessorMPerf           *prometheus.Desc
+	ProcessorRTC             *prometheus.Desc
+	ProcessorUtility         *prometheus.Desc
+	ProcessorPrivUtility     *prometheus.Desc
 }
 
 // newCPUCollector constructs a new cpuCollector, appropriate for the running OS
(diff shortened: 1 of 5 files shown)


breed808 (commented):
Huge apologies for my absence these past few weeks; work and personal matters have taken up much of my spare time lately.

The PR looks good. I'm impressed with the code changes (documentation and unit tests are great :+1: ), though we'll need to update the e2e test output with the new and updated metrics.

Would you be comfortable doing that, or would you like me to push a commit with the changes to your feature branch?

higels:
Thanks Ben - no worries on the delay; I did leave my issue related to this sitting for over a year after all!

I'll see if I can get the e2e tests fixed up.

higels:
e2e tests looks good now - that was a bit less work than anticipated and I shouldn't have procrastinated for as long as I did.

breed808 (approved):
Not a problem, we all procrastinate from time to time. Thanks again for putting the work into this change :+1:

PR #1096, opened 2022-11-19
Compatibily with new flag web.listen-address

bck01215:
Signed-off-by: Brandon Kauffman <bck01215@gmail.com>

### DIFF
--- a/installer/windows_exporter.wxs
+++ b/installer/windows_exporter.wxs
@@ -24,9 +24,9 @@
 
     <Property Id="LISTEN_ADDR" Secure="yes" />
     <Property Id="LISTEN_PORT" Secure="yes" />
-    <SetProperty Id="ListenFlagBoth" After="InstallFiles" Sequence="execute" Value="--telemetry.addr [LISTEN_ADDR]:[LISTEN_PORT]">LISTEN_ADDR AND LISTEN_PORT</SetProperty>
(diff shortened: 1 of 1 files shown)


JDA88:
I agree this change is necessary but as it’s a breaking change, can we have both parameter working at least until the `0.22.*` to ease the transition for organization using it?

breed808:
There's no easy way to have both options working correctly, as the exporter toolkit we rely on requires the new flag.

The `v0.21.0` release could be pinned to 63fb570, but would miss out on most of the recent changes & fixes.

There is the option of continuing to run v0.20.0 until you are ready for v0.21.0 or v0.22.0.

JDA88:
Ok, I understand.
I didn't have the opportunity to test is but what would append if you set both flags in 0.20 and 0.21? If the additional flag is ignored that would be perfect.
The issue for us is to find a way to ease the migration and it's currently a case our deployment scripts do not handle.

breed808:
Technically both flags could be included, but from memory the only the new flag would be properly used by the exporter, due to the change in behavior for the exporter toolkit.

JDA88:
I'll try to test this. It should ease our migration path.

JDA88:
OK after some test you can't use both flag at the same time, you get the error message:
`windows_exporter-0.21.0-amd64.exe: error: unknown long flag '--telemetry.addr', try --help`

We will adjust our deployement script to check for the version of the binary and change the command line in consequence

PR #1098, opened 2022-11-27
Include OpenHardwareMonitor collector

b-tuma:
Collects OpenHardwareMonitor sensors (if the application is running).

Example of metrics:
`windows_openHardwareMonitor_sensor_value{index="0",name="GPU Core",parent="/nvidiagpu/0",sensor_type="Temperature"} 50`
`windows_openHardwareMonitor_sensor_value{index="4",name="CPU Core #4",parent="/intelcpu/0",sensor_type="Clock"} 3914.177490234375`

Implemented it due to Thermalzone not getting temperatures consistently on different hardwares.


collector/openhardwaremonitor.go
```diff
@@ -0,0 +1,82 @@
...
+func init() {
+	registerCollector("openhardwaremonitor", NewOpenHardwareMonitorCollector)
+}
+
+// A openHardwareMonitorCollector is a Prometheus collector for WMI Sensor metrics
+type SensorCollector struct {
+	SensorType *prometheus.Desc
+	Identifier *prometheus.Desc
+	Parent     *prometheus.Desc
+	Name       *prometheus.Desc
```
breed808, line 20: Are the sensor names dynamic? If the names are static and known before execution, it may be worth moving the `name` from the labels to the metric name.

breed808 (commented):
Thanks for putting the time into this PR!
I've left a couple minor comments, and you'll need to rebase this feature branch on `master` to resolve the merge conflict on the Go module files.

Let me know if further info is needed.

collector/openhardwaremonitor.go
b-tuma: The sensor name is dynamic... Talking about this, I wanted to move SensorType from the label to the metric name, as it has [8 predefined types with known units](http://openhardwaremonitor.org/wordpress/wp-content/uploads/2011/04/OpenHardwareMonitor-WMI.pdf), but couldn't find a way to make the metric name change accordingly. I'm new to Go (and this codebase), so maybe it's just ignorance, but is there any way to resolve the metric name during execution?

collector/openhardwaremonitor.go
```diff
@@ -0,0 +1,82 @@
...
+	Value      *prometheus.Desc
+	Max        *prometheus.Desc
+	Min        *prometheus.Desc
+	Index      *prometheus.Desc
+}
+
+// NewOpenHardwareMonitorCollector ...
+func NewOpenHardwareMonitorCollector() (Collector, error) {
+	const subsystem = "openHardwareMonitor"
```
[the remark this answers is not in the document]
b-tuma: My bad, didn't notice the existing convention there, fixed!

b-tuma:
Sorry for the delay, December has been kind of hectic here.
Thanks! I'll work on the issues mentioned

collector/openhardwaremonitor.go
breed808: The `NewOpenHardwareMonitorCollector` function is called on exporter initialisation, so we can't dynamically create metric names at runtime.

As the types are static, they could be included in the metric names. Here's two as an example:

```go
	return &SensorCollector{
		VoltageValue: prometheus.NewDesc(
			prometheus.BuildFQName(Namespace, subsystem, "sensor_value_volts"),
			"Value from an OpenHardwareMonitor sensor in volts",
			[]string{"parent", "index", "name"},
			nil,
		),
	}, nil
	return &SensorCollector{
		TemperatureValue: prometheus.NewDesc(
			prometheus.BuildFQName(Namespace, subsystem, "sensor_value_degrees"),
			"Calue from an OpenHardwareMonitor sensor in degrees Celsius",
			[]string{"parent", "index", "name"},
			nil,
		),
	}, nil
```

duxet:
I've merged this PR with master and added separate metric for each sensor type in my branch: https://github.com/duxet/windows_exporter/tree/openhardwaremonitor

The only problem I have with this change is that `/metrics` endpoint requires ~17s to complete with only openhardwaremonitor collector enabled, but maybe it's just some problem with my machine or openhardwaremonitor is just a bit sluggy 🤔

PR #1100, opened 2022-12-04
Cleaning up listen addresses flags before parsing config file

knabben:
Parsing the flags twice duplicates the value for slice flags, it seems other binaries (alertmanager, exporters) are using both flags and treat it differently, besides this dirty fix solving this issue in particular, a bigger refactoring is required.

The following scenarios were tested:

1) Setting the bind flag directly: `--config.file=config.yml --web.listen-address=":9099"` bind on `:9099`

2) No bind flag and config set + passed via flag:
```
config.yml:
web:
  listen-address: :9183

--config.file=config.yml
```
bind on :9183

3) No bind flags and config flag = `--config.file=config.yml` - bind on default :9182

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -303,6 +303,12 @@ func main() {
 		if err != nil {
 			log.Fatalf("%v\n", err)
 		}
+
+		// NOTE: This is temporary fix for issue #1092, calling kingpin.Parse
+		// twice makes slices flags duplicate its value, this clean up
+		// the first parse before the second call.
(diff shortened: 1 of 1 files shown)


knabben:
@breed808 this is a block to run the project as Kubernetes HPC using > 0.20.0

breed808 (approved):
Thanks for submitting and investigating this @knabben!

I acknowledge this is a not-so-ideal hack to the current not-so-ideal situation (same options exposed via flags and config file requiring double parsing).
A discussion on the method of exposing options needs to be had, but that can be done in the issue.

PR #1108, opened 2022-12-16
fix: Continue os collection on missing page file

### LINKED ISSUE #1069: OS collector fails with message "failed collecting os metrics:<nil> CreateFile D:\pagefile.sys: The system cannot find the file specified."
We are using 0.19.0 version with Prometheus 2.34.0.
windows_exporter_collector_success shows 0 for os collector. All other collectors succeeded.

we checked that pagefile is indeed enabled only on d: drive with "system managed" size. It is disabled on c: drive

we already tried executing lodctr /R many times and restarting windows exporter windows service many times.
Windows exporter windows service runs in local system account.

Windows exporter config:

collectors:
  enabled: cpu,cpu_info,cs,iis,logical_disk,logon,memory,net,os,process,service,system,time

image

breed808:
Resolves #1069

### DIFF
--- a/collector/os.go
+++ b/collector/os.go
@@ -202,10 +202,12 @@ func (c *OSCollector) collect(ctx *ScrapeContext, ch chan<- prometheus.Metric) (
 	for _, pagingFile := range pagingFiles {
 		fileString := strings.ReplaceAll(pagingFile, `\??\`, "")
 		file, err := os.Stat(fileString)
+		// For unknown reasons, Windows doesn't always create a page file. Continue collection rather than aborting.
 		if err != nil {
-			return nil, err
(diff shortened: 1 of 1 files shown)


ReneZeidler (changes requested):
`file` is nil on error, see https://github.com/prometheus-community/windows_exporter/issues/1069#issuecomment-1354761514

```diff
		if err != nil {
			log.Debugf("Failed to read page file (reason: %s): %s\n", err, fileString)
-		}
-		fsipf += float64(file.Size())
+		} else {
+			fsipf += float64(file.Size())
+		}
```

breed808:
Thanks for picking that up. Perhaps I shouldn't submit PRs before coffee...

Let me know if the recent push has any issues

ReneZeidler:
Thanks for the fix, it works as expected

PR #1112, opened 2022-12-21
Remove init functions from collectors

breed808:
Behaviour of init functions has been centralised in `collector/init.go`,
and can be called during exporter startup. This allows the exporter to
control the timing of collector initialisation, rather than relying on
the import & `init()` method.

This should reduce unexpected behaviour arising from the use of
`init()`, such as #551.


SuperQ:
This needs a DCO sign-off. You can use `git commit -s --amend` to add it.

PR #1137, opened 2023-02-15
Added quote information for absolute paths

DennisGaida:
This had me running in circles because `ioutil.ReadFile` did not read the local file for me when running as a service (i.e. `config.yml` in the same folder as the `windows_exporter.exe`). I had to use absolute paths and additionally quote them.


breed808 (approved):
:thumbsup:

PR #1161, opened 2023-03-20
feat: Allow default metrics to be disabled

### LINKED ISSUE #1159: [question] collectors.enabled working not as expected
Hi team,
Just trying to start the exporter by command `windows_exporter-0.21.0-386.exe --config.file="C:\prom_exporter\config.yml" --collectors.enabled textfile` and still see some go_memstat*  and windows_exporter_collector_* metrics, self metrics in another words. Can I disable it somehow? In node exporter there is parameter --web.disable_exporter-metrics. Will it work here?
The second question is can I collect metrics from multiple files or only one? By default I received error in the output.
Thanks.

breed808:
Introduces a new `--disable-exporter-metrics` to allow users to disable metrics included by the Prometheus client library (promhttp_*, process_*, go_*)

Resolves #1159

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -269,6 +269,10 @@ func main() {
 			"telemetry.path",
 			"URL path for surfacing collected metrics.",
 		).Default("/metrics").String()
+		disableExporterMetrics = kingpin.Flag(
+			"web.disable-exporter-metrics",
+			"Exclude metrics about the exporter itself (promhttp_*, process_*, go_*).",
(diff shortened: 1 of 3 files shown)


JDA88:
Ii it `--web.disable-exporter-metrics` or `--disable-exporter-metrics` as described in the first comment?

breed808:
Thanks for spotting that! I've updated the comment which will come in handy for the next release.

Issue #679, opened 2020-12-09
Change blacklist / whitelist wording


roidelapluie:
Some collectors, like process, use blacklist / whitelist.

https://github.com/prometheus-community/windows_exporter/blob/master/docs/collector.process.md

We should use other, non race-charged terms.

breed808:
Resolved by #1337

PR #1170, opened 2023-04-01
feat: Deprecate whitelist/blacklist flags

breed808:
Flags have been deprecated in favour of include/exclude terminology.

This addresses #679. Resolution of the issue will occur when the flags are removed.


breed808:
@jkroepke note that I've changed the flag constant names in this PR, they'll need to be updated for any embedded windows_exporter instances.
This should be the only time I break things by changing the flag names.

jkroepke:
@breed808 I was ware of this PR.

Do it. Do it now, before next release 👍 

It should break nothing at least on my site, because there is not release which includes the current yet. I consider anything as unstable until tis released.

JDA88:
How long do you plan to keep the olds flags working in // of the new ones?

breed808:
@JDA88 I like to leave the old flags for at least another two releases (I.E. removing the old flags in `v0.26.0`), but this will depend on the future release cadence.

If there's enough work on the exporter to warrant another two releases in a short period of time, I'd expect the old flags to remain for more than two releases

JDA88:
> If there's enough work on the exporter to warrant another two releases in a short period of time, I'd expect the old flags to remain for more than two releases

Thanks for the info, this one will require a little refactoring on our side

PR #1179, opened 2023-04-08
fix Terminal Service duplicate session names under load.

peekjef72:
On prod env (30 hosts with more than 40 cnx each), connection's ,names are reused  and caused  duplicated label names.
Add sequential number to name so name is unique.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/sirupsen/logrus v1.9.0
 	github.com/yusufpapurcu/wmi v1.2.2
 	go.opencensus.io v0.23.0 // indirect
-	golang.org/x/sys v0.6.0
+	golang.org/x/sys v0.7.0
 	gopkg.in/yaml.v3 v3.0.1
 )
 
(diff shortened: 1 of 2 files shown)


breed808:
This appears to be solving a similar issue to #1133 (duplicate perflib entries).

Are suffixes added to the connection names by Perflib, causing duplicates in the collector?
E.G. `connection_name` and `connection_name#2`.

In the above example, would these be two unique connections or are they unrelated and Perflib has simply re-used the name?

peekjef72:
I have experimened this week with the first fix( add a sequential to session names).
The duplicates have exactly the same name : e.g.: RDP-tcp 3, or Disk-X... 
I have no idea where they are from, and we don't use these metrics at all. We only use **"LocalSessionCount**" sum all over the farm.
I have published a new fixe that remove duplicated names, because it is a better solution (to my opinion). 
The first fix increase metrics cardinality, because sessions are not delivered in the same order each time; so "RDP-Tcp-3_X", should be "RDP-TCP-3_y" at next poll.
=> see PR#1184

PR #1180, opened 2023-04-08
Service custom labels

peekjef72:
this PR adds some feature to service collector:
 - add config parameter **collector.service.services-list** with a comma separated list of service names:
 ```yml
collector:
  service:
    services-list: windows_exporter, winRM, pushprox_client, Dhcp
```
the param will only be checked if services-where is not set.
the list will be used to build a service-where query based on Name and is equivalent to:
```yml
collector:
  service:
    services-where: "Name = 'elmt_1' or Name = "elmt_X' or ..."
```
 - add a new parameter that can only be set in config file: 
It allows to set any number of custom labels value for each service:
e.g.:
```yml
collector:
  service:
    services:
      windows_exporter:
        application: prometheus
        custom1: val1
      pushprox_client:
        application: prometheus
        custom1: val1
      winRM:
        application: windows
        custom1: val2
      Dhcp:
        application: windows
        custom1: val3
```
Use case: allow to build a generic Prometheus alert on not running service and to use the specified associated labels to drive a specific behavior. For me they are used to route for a specific documentation based on context.<br>
Label's names must be identical for each service. Not identical labels names are removed !
This parameter as a lower priority than service-where and service-list.
It accepts a dict (see above example) or a list; in this case it behaves like services-list.
e.g.:
```yml
collector:
  service:
    services:
      - dhcp
      - pushprox_client
      - windows_exporter
      - winRM
```

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/leoluk/perflib_exporter/perflib"
+	"github.com/prometheus-community/windows_exporter/config"
 	"github.com/prometheus-community/windows_exporter/log"
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/sys/windows/registry"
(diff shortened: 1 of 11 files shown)


peekjef72:
replace by PR [PR#1185](https://github.com/prometheus-community/windows_exporter/pull/1185)

PR #1185, opened 2023-04-15
Add new parameters for collector.service: services-list and custom labels

peekjef72:
replace previous PR [PR#1180](https://github.com/prometheus-community/windows_exporter/pull/1180)

This PR adds some feature to service collector:
 - add config parameter **collector.service.services-list** with a comma separated list of service names:
 ```yml
collector:
  service:
    services-list: windows_exporter, winRM, pushprox_client, Dhcp
```
the param will only be checked if services-where is not set.
the list will be used to build a service-where query based on Name and is equivalent to:
```yml
collector:
  service:
    services-where: "Name = 'elmt_1' or Name = "elmt_X' or ..."
```
 - add a new parameter that can only be set in config file: 
It allows to set any number of custom labels value for each service:
e.g.:
```yml
collector:
  service:
    services:
      windows_exporter:
        application: prometheus
        custom1: val1
      pushprox_client:
        application: prometheus
        custom1: val1
      winRM:
        application: windows
        custom1: val2
      Dhcp:
        application: windows
        custom1: val3
```
Use case: allow to build a generic Prometheus alert on not running service and to use the specified associated labels to drive a specific behavior. For me they are used to route for a specific documentation based on context.<br>
Label's names must be identical for each service. Not identical labels names are removed !
This parameter as a lower priority than service-where and service-list.
It accepts a dict (see above example) or a list; in this case it behaves like services-list.
e.g.:
```yml
collector:
  service:
    services:
      - dhcp
      - pushprox_client
      - windows_exporter
      - winRM
```

Then generic alert services:

```yml
groups:
- name: Window Server Alerts
  rules:

  # Sends an alert when the 'sqlserveragent' service is not in the running state for 3 minutes.
  - alert: WindowServiceNotRunning
    expr: windows_service_state{state="running"} == 0
    for: 3m
    labels:
      severity: high
    annotations:
      summary: "Service {{ $labels.name }} for {{ $labels.application }} down for 3 min."
      description: "Service {{ $labels.name }} for Application {{ $labels.application }} on instance {{ $labels.instance }} has been down for more than 3 minutes."
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Name | Description
 `LISTEN_PORT` | The port to bind to. Defaults to 9182.
 `METRICS_PATH` | The path at which to serve metrics. Defaults to `/metrics`
 `TEXTFILE_DIR` | As the `--collector.textfile.directory` flag, provide a directory to read text files with metrics from
-`REMOTE_ADDR` | Allows setting comma separated remote IP addresses for the Windows Firewall exception (whitelist). Defaults to an empty string (any remote address).
+`REMOTE_ADDR` | Allows setting comma separated remote IP addresses for the Windows Firewall exception (allow list). Defaults to an empty string (any remote address).
 `EXTRA_FLAGS` | Allows passing full CLI flags. Defaults to an empty string.
 
 Parameters are sent to the installer via `msiexec`. Example invocations:
@@ -150,7 +150,7 @@ The prometheus metrics will be exposed on [localhost:9182](http://localhost:9182
 
 ### Enable only process collector and specify a custom query
 
-    .\windows_exporter.exe --collectors.enabled "process" --collector.process.whitelist="firefox.+"
+    .\windows_exporter.exe --collectors.enabled "process" --collector.process.include="firefox.+"
 
 When there are multiple processes with the same name, WMI represents those after the first instance as `process-name#index`. So to get them all, rather than just the first one, the [regular expression](https://en.wikipedia.org/wiki/Regular_expression) must use `.+`. See [process](docs/collector.process.md) for more information.
 
--- a/collector/iis.go
+++ b/collector/iis.go
@@ -4,6 +4,7 @@
 package collector
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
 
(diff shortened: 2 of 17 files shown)


collector/textfile.go
```diff
@@ -308,7 +307,12 @@ fileLoop:
 
 		// Only set this once it has been parsed and validated, so that
 		// a failure does not appear fresh.
-		mtimes[f.Name()] = f.ModTime()
+		info, err := f.Info()
+		if err != nil {
+			error = 1.0
```
breed808, line 312: It's be worth adding a log entry here, to prevent user confusion in the event of a failure.

breed808 (commented):
I've been reviewing each commit individually, and it appears they may be out of order?
9d8270f15a8a31926b62756508372c9731ba1932 fails to build:

```
$ git status
HEAD detached at 9d8270f
...
$ go build
# github.com/prometheus-community/windows_exporter/collector
collector/init.go:17:33: undefined: config.ConfigHook
collector/init.go:253:21: undefined: ServiceBuildHook
collector/collector.go:59:51: undefined: config.ConfigHook
collector/collector.go:62:87: undefined: config.ConfigHook
collector/collector.go:85:38: undefined: config.ConfigHook
```

peekjef72:
> I've been reviewing each commit individually, and it appears they may be out of order? [9d8270f](https://github.com/prometheus-community/windows_exporter/commit/9d8270f15a8a31926b62756508372c9731ba1932) fails to build:
> 
> ```
> $ git status
> HEAD detached at 9d8270f
> ...
> $ go build
> # github.com/prometheus-community/windows_exporter/collector
> collector/init.go:17:33: undefined: config.ConfigHook
> collector/init.go:253:21: undefined: ServiceBuildHook
> collector/collector.go:59:51: undefined: config.ConfigHook
> collector/collector.go:62:87: undefined: config.ConfigHook
> collector/collector.go:85:38: undefined: config.ConfigHook
> ```

I'm a little bit confused with what you have tried to do.
I have committed files by group of changes ( what they do), not by independant function.
The program globally compiles. That is all I can say.
Do you prefer I post a single global commit ?

breed808:
You're on the right track; each commit should be a self-contained change.
However `9d8270f` wasn't self-contained as it referenced objects that didn't exist (config.ConfigHook, ServiceBuildHook, config.ConfigHook).

These were likely introduced in a later commit. Interactively rebasing the feature branch such that `9d8270f` occurs _after_ the commit introducing these objects should have worked.

A single, global commit does work but can become difficult to review once it reaches a certain size.

PR #1189, opened 2023-04-17
Add yaml for WS2022 nodes

vrapolinario:
Adding a yaml spec with an image supported on Windows Server 2022 nodes.

### DIFF
--- a/kubernetes/windows-exporter-daemonset-ws2022.yaml
+++ b/kubernetes/windows-exporter-daemonset-ws2022.yaml
@@ -1,0 +1,61 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  labels:
+    app: windows-exporter
+  name: windows-exporter
+  namespace: monitoring
+spec:
+  selector:
+    matchLabels:
+      app: windows-exporter
+  template:
+    metadata:
+      labels:
+        app: windows-exporter
+    spec:
+      securityContext:
+        windowsOptions:
+          hostProcess: true
+          runAsUserName: "NT AUTHORITY\\system"
+      hostNetwork: true
+      initContainers:
+        - name: configure-firewall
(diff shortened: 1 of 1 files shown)


breed808 (commented):
Thanks for submitting this. The 2022 manifest appears to be identical to the existing manifest, except for the image tag.

I don't think the level of duplication here is ideal, but the complexity of the manifests isn't enough to warrant a helm chart (not that I could support one anyway).

Do you think it would be worth updating the tag in the existing manifest?

vrapolinario:
That would be a solution for WS2022 nodes - hopefully that's what people are using these days, but then it becomes a problem for people running WS2019, hence my proposal for a separate file.

breed808:
Just so I understand correctly, for windows containers the image and node must match? I.E. a Windows 2019 node cannot run 2022 container images?

vrapolinario:
That is correct. And the images are version specific.

vrapolinario:
That is correct. And the images are version specific. More info here: https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility?tabs=windows-server-2022%2Cwindows-11

breed808:
Thanks for enlightening me on this, I wasn't aware of the Windows container/host restrictions. If you sign the commit (to pass the DCO check) I'll get this merged.

vrapolinario:
#sign-off

breed808:
@vrapolinario you'll need to sign the commit with `git commit --amend --no-edit -s`. You can open the DCO details link for more info on the signing.

vrapolinario:
Sorry for the delay - JubeCon and all. I have just signed the commit following the instructions. Apologies as my Git abilities are very limited. :)

breed808 (approved):
Not a problem, thanks for updating it.

PR #1192, opened 2023-04-22
switch to go-kit logger

### LINKED ISSUE #979: How to log to a file (rather than stdout / stderr / eventlog)

jkroepke:
Fixes #979
Replaces #987 

See 

* https://github.com/prometheus/node_exporter/pull/1575 
* https://github.com/prometheus/blackbox_exporter/pull/218
* https://github.com/prometheus/mysqld_exporter/pull/433
 
for similar changes

depends on

- https://github.com/prometheus/common/pull/474
- https://github.com/go-kit/log/pull/31

---

This is a breaking change for the `log.format` CLI flags.

* `--log.format logger:stderr` -> `--log.file stderr --log.format logfmt`
* `--log.format logger:stdout?json=true` -> `--log.file stdout --log.format json`
* `--log.format logger:eventlog?name=windows_exporter` -> `--log.file eventlog`

The breaking change happens on node_exporter before version 1, too. They dont add any backwards compability.

windows_exporter is still < version 1. In terms of semver, a breaking change is fine here.


collector/iis.go
```diff
@@ -1027,8 +1033,8 @@ func dedupIISNames[V hasGetIISName](services []V) map[string]V {
 }
 
 func (c *IISCollector) collectWebService(ctx *ScrapeContext, ch chan<- prometheus.Metric) (*prometheus.Desc, error) {
-	var webService []perflibWebService
-	if err := unmarshalObject(ctx.perfObjects["Web Service"], &webService); err != nil {
+	var WebService []perflibWebService
+	if err := unmarshalObject(ctx.perfObjects["Web Service"], &WebService, c.logger); err != nil {
```
jkroepke, line 1037:
```suggestion
	var webService []perflibWebService
	if err := unmarshalObject(ctx.perfObjects["Web Service"], &webService, c.logger); err != nil {
```

jkroepke:
@breed808 can you do take a look here?

jkroepke:
@breed808 I would appreciate a review here.

jkroepke:
@breed808 I would appreciate a review here, please let me know, if I can assist here.

collector/container.go
```diff
@@ -255,7 +260,7 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
 		)
 
 		if len(cstats.Network) == 0 {
-			log.Info("No Network Stats for container: ", containerDetails.ID)
+			_ = level.Info(c.logger).Log("msg", "No Network Stats for container", "containetId", containerDetails.ID)
```
breed808, line 263: Spelling error? `containetId` -> `containerId`

collector/time.go
```diff
@@ -20,14 +23,16 @@ type TimeCollector struct {
...
 }
 
-func newTimeCollector() (Collector, error) {
-	if getWindowsVersion() <= 6.1 {
-		return nil, errors.New("Windows version older than Server 2016 detected. The time collector will not run and should be disabled via CLI flags or configuration file")
+func newTimeCollector(logger log.Logger) (Collector, error) {
+	const subsystem = "time"
+	logger = log.With(logger, "collector", subsystem)
 
+	if getWindowsVersion(logger) < 100 {
```
breed808, line 30: Looking at the documentation for `getWindowsVersion`, a value of `100` is correct here. However it might be best to move this fix into a separate commit for increased visibility if we need to revisit the value in the future.

breed808 (commented):
Apologies for my absence in reviewing this. Thanks very much for the time and effort you've put into this change, it's another really good PR :smile:

I've left a few minor comments, but aside from those I have no issues with change proposed by this PR.

log/logger.go
```diff
@@ -0,0 +1,90 @@
...
+func (f *AllowedFile) Set(s string) error {
+	switch s {
+	case "stdout":
+		f.w = os.Stdout
+	case "stderr":
+		f.w = os.Stderr
+	case "eventlog":
+		f.w = nil
+	default:
+		file, err := os.OpenFile(s, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0640)
```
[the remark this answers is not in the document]
jkroepke: I have to re-test this. It seems like it does not have any effect here.

Reading https://pkg.go.dev/os#Chmod, it says only 0200 is supported for "owner write". I test 0600, 0400, 0640 and each combination gives the admin group full permission on it.

Using 0000 here sets the file to read-only.

collector/container.go
jkroepke: thanks!

collector/time.go
jkroepke: You are correct.

jkroepke:
I addressed all your changes. I hope thats fine now.

breed808 (approved):
Looks good, thanks for integrating those changes :+1:

PR #1195, opened 2023-04-23
Move prometheus collector from main to collectors

jkroepke:
This is useful, if windows_exporter is embedded als library. Right now the collector is part of the main package and can not start separately.

Tested on a windows 10 client machine

### DIFF
--- a/collector/prometheus.go
+++ b/collector/prometheus.go
@@ -1,0 +1,190 @@
+//go:build windows
+// +build windows
+
+package collector
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/prometheus-community/windows_exporter/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// Base metrics returned by Prometheus
+
+var (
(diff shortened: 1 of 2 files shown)


breed808:
Code looks good, I'm trying to understand the use-case. Would this be desired to omit the "base" metrics and collection when using the exporter as a library?

jkroepke:
I on the track to bump the windows_exporter dependency on the grafana agent. This patch is included the outdated fork of [grafana](https://github.com/grafana/windows_exporter). As I understood, if windows_exporter is used as library, its currently not possible to re-use the Prometheus Registry from windows_exporter. The go process base metrics are not included. but before the prometheus registry from main package could not be used anyways.

breed808 (approved):
Thanks for the context, I'm happy to merge it :+1:

jkroepke:
Thanks @breed808 for merging it!

The last part of my PR series is #1192 that an external log instance can be used, but https://github.com/prometheus/common/pull/474 is really slow here.

PR #1201, opened 2023-05-04
fix: add missing vswitch packets_sent_total

ilyam8:
`windows_hyperv_vswitch_packets_sent_total ` is in the [docs](https://github.com/prometheus-community/windows_exporter/blob/master/docs/collector.hyperv.md#metrics) but not being collected. This PR fixes the problem.


breed808 (approved):
Thanks for submitting this!

Issue #1129, opened 2023-01-28
Missing metrics from 'container' collector


bragi92:
The following [document](https://github.com/prometheus-community/windows_exporter/blob/54f86001ad296b9b2334f8e0a7af6bb0c4f95bf3/docs/collector.container.md) states that windows_container_network_receive_bytes_total is supposed to be present when one has enabled the container collector in windows exporter.

I have an AKS cluster where I deployed windows exporter as a [daemonset ](https://github.com/Azure/prometheus-collector/blob/main/internal/windowsExporterInstaller/windows-exporter-daemonset.yaml) with the container collector enabled.

I see that the following two metrics : `windows_container_network_received_bytes_total` and `windows_container_network_transmitted_bytes_total` are not present when I query for them using `http://{NODE_IP}:9182/metrics`.

I do see similar metrics present from the [net](https://github.com/prometheus-community/windows_exporter/blob/master/docs/collector.net.md) collector  called  `windows_net_bytes_received_total` and `windows_net_bytes_sent_total`.

Have the former metrics been deprecated in lieu of the latter?

breed808:
Might this be an incorrect metric name? `windows_container_network_receive_bytes_total` is in the documentation and code, but not `windows_container_network_received_bytes_total`.
It appears to be a similar situation with transmitted -> transmit.

These particular metrics haven't been modified since introduction in 8841091f9c7c9e88e6a5948a656af1bc29cfcd58.

bragi92:
@breed808 Sorry, I made a typo when creating the issue. I have been using the proper name present in the documenation and code i.e. `windows_container_network_receive_bytes_total` and `windows_container_network_transmit_bytes_total`. and I do not see the metrics at all.

When I query `http://{NODE_IP}:9182/metrics.` I do not see any metrics that start with the format `windows_container_network_*` too. I'm not sure if this was ever working but its possible the underlying methods to fetch the container network metrics are now silently failing?

breed808:
Are any container metrics being exposed by the collector? Is the `container` collector returning a successful result? I.E. `windows_exporter_collector_success{collector="container"} 1`

It'd be worth enabling debug logging and viewing the exporter output (if possible in a container).

bragi92:
Yes, I have verified that the success metrics returns 1:
![image](https://user-images.githubusercontent.com/28612268/221984818-bf86d94b-4c6a-4819-927a-36b58229463d.png)

[...]

Issue #806, opened 2021-06-15
Lots of warnings: No Network Stats for container


bip7686:
I'm running the windows exporter through wins for my kubernetes cluster.
The windows exporter is printing a lot of messages like:
```
time="2021-06-15T18:28:29+02:00" level=info msg="No Network Stats for container: 52026722c95ef2d36d3bc894dfe4ee1e910277c315e4878fa43b3dddc533e5e2" source="container.go:231"
```

WMI Setup:
```time="2021-06-15T14:26:06+02:00" level=warning msg="No where-clause specified for service collector. This will generate a very large number of metrics!" source="service.go:41"
time="2021-06-15T14:26:06+02:00" level=info msg="Enabled collectors: cache, system, os, cpu, memory, logical_disk, cs, container, service, net" source="exporter.go:347"
time="2021-06-15T14:26:06+02:00" level=info msg="Starting windows_exporter (version=0.16.0, branch=master, revision=f316d81d50738eb0410b0748c5dcdc6874afe95a)" source="exporter.go:399"
time="2021-06-15T14:26:06+02:00" level=info msg="Build context (go=go1.15.6, user=appveyor-vm\\appveyor@appveyor-vm, date=20210225-10:52:19)" source="exporter.go:400"
time="2021-06-15T14:26:06+02:00" level=info msg="Starting server on 0.0.0.0:9100" source="exporter.go:403"
time="2021-06-15T14:26:06+02:00" level=info msg="TLS is disabled." source="gokit_adapter.go:38"
time="2021-06-15T14:26:06+02:00" level=error msg="Failed to start service: The service process could not connect to the service controller." source="exporter.go:337"
```

Windows Node:
BuildNumber: 17763
Version: 10.0.17763

carlpett:
Just to be clear, is this about missing data that should be there, or about excessive logging? Does the containers actually have network interfaces attached to them?

Luis-TT:
The data seems to be there. How would I check if the containers have network interfaces attached to them?

jsturtevant:
checkout how this can be done with [hostprocess containers](https://github.com/prometheus-community/windows_exporter/blob/master/kubernetes/kubernetes.md#kubernetes-quick-start). There are also examples on wiring all this up with the rest of the Prometheus stack in https://github.com/prometheus-operator/kube-prometheus/blob/main/docs/windows.md

PR #1218, opened 2023-06-03
Resolve issue preventing networking container stats from being sent to collector

tzifudzi:
#### What type of PR is this?
- bugfix
 
- This fix ensures that network stats for containerd on Windows are successfully collected. Before this change, other container stats such as CPU and memory are successfully collected, but network stats are not being collected successfully when using containerd.
- The root cause is that the code for collecting network stats was originally written to work with dockershim which relies on HCS v1 schema. After dockershim removal as Kubernetes's container runtime, containerd adoption has increased and this error is more frequently encountered when using containerd as the runtime. containerd uses HCS v2 schema whereby the network stats need to be queried from the HNS component.

#### Which issue(s) this PR fixes:
- https://github.com/prometheus-community/windows_exporter/issues/1129
- https://github.com/prometheus-community/windows_exporter/issues/806

#### Notes for reviewers
- This is my first pull request in this [windows_exporter](https://github.com/prometheus-community/windows_exporter) project. Any feedback and guidance is appreciated.

#### Testing/Verification
Manually tested by compiling new binary with changes and subsequently installing and running the new binary. Tested on both Docker and containerd. 

###### When testing with containerd on Kubernetes
- used containerd 1.6.6
- used Windows Server 2022 Datacenter 10.0.20348.1726
- used AWS EKS with the [AWS VPC CNI plugin](https://github.com/aws/amazon-vpc-cni-plugins/tree/master/plugins/vpc-bridge)
 
```
# HELP windows_container_network_transmit_bytes_total Bytes Sent on Interface
# TYPE windows_container_network_transmit_bytes_total counter
windows_container_network_transmit_bytes_total{container_id="containerd://1ef7eb8f8c96538c8395c3b195029e1e674951b0105d14fe31f955b276b54008",interface="467D7000-DEA3-4BD8-B3CE-424CEEB959D0"} 301289
```

###### When testing with Docker on Kubernetes
- used Docker 20.10.21
- used Windows Server 2022 Datacenter 10.0.20348.1726
- used Windows Server 2019 Datacenter 10.0.17763.4377
- used AWS EKS with the [AWS VPC CNI plugin](https://github.com/aws/amazon-vpc-cni-plugins/tree/master/plugins/vpc-bridge)

```
# HELP windows_container_network_transmit_bytes_total Bytes Sent on Interface
# TYPE windows_container_network_transmit_bytes_total counter
windows_container_network_transmit_bytes_total{container_id="docker://04c0c86b64342c4bf9c8390555bc254e5f07f2cc85e046fe8b5fc8d2f9d81106",interface="52A2F900-32C7-4CB1-832D-8108ED966FB4"} 241375
```

###### When testing with Docker on plain Windows host (without Kubernetes)
- used Docker 24.0.2
- used Windows Server 2022 Datacenter 10.0.20348.1787

```
# HELP windows_container_network_transmit_bytes_total Bytes Sent on Interface
# TYPE windows_container_network_transmit_bytes_total counter
windows_container_network_transmit_bytes_total{container_id="docker://062ac7f1925dd3a7054089b05bf9b2b8bdd4920d89fe3b944584e1dbf7e98065",interface="87AD7194-4445-4D5B-869F-5B6AF713C121"} 15745
```

### DIFF
--- a/collector/container.go
+++ b/collector/container.go
@@ -4,6 +4,9 @@
 package collector
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/Microsoft/hcsshim"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -199,6 +202,8 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
 		return nil, nil
 	}
 
+	containerPrefixes := make(map[string]string)
+
 	for _, containerDetails := range containers {
 		container, err := hcsshim.OpenContainer(containerDetails.ID)
 		if container != nil {
@@ -214,7 +219,9 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
 			_ = level.Error(c.logger).Log("msg", "err in fetching container Statistics", "containerId", containerDetails.ID, "err", err)
 			continue
 		}
+
 		containerIdWithPrefix := getContainerIdWithPrefix(containerDetails)
+		containerPrefixes[containerDetails.ID] = containerIdWithPrefix
 
 		ch <- prometheus.MustNewConstMetric(
 			c.ContainerAvailable,
@@ -258,54 +265,6 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
 			float64(cstats.Processor.RuntimeKernel100ns)*ticksToSecondsScaleFactor,
 			containerIdWithPrefix,
 		)
-
-		if len(cstats.Network) == 0 {
-			_ = level.Info(c.logger).Log("msg", "No Network Stats for container", "containerId", containerDetails.ID)
-			continue
-		}
-
-		networkStats := cstats.Network
-
-		for _, networkInterface := range networkStats {
-			ch <- prometheus.MustNewConstMetric(
-				c.BytesReceived,
-				prometheus.CounterValue,
-				float64(networkInterface.BytesReceived),
-				containerIdWithPrefix, networkInterface.EndpointId,
-			)
-			ch <- prometheus.MustNewConstMetric(
-				c.BytesSent,
-				prometheus.CounterValue,
-				float64(networkInterface.BytesSent),
-				containerIdWithPrefix, networkInterface.EndpointId,
-			)
-			ch <- prometheus.MustNewConstMetric(
-				c.PacketsReceived,
-				prometheus.CounterValue,
-				float64(networkInterface.PacketsReceived),
-				containerIdWithPrefix, networkInterface.EndpointId,
-			)
-			ch <- prometheus.MustNewConstMetric(
-				c.PacketsSent,
-				prometheus.CounterValue,
-				float64(networkInterface.PacketsSent),
-				containerIdWithPrefix, networkInterface.EndpointId,
-			)
-			ch <- prometheus.MustNewConstMetric(
-				c.DroppedPacketsIncoming,
-				prometheus.CounterValue,
-				float64(networkInterface.DroppedPacketsIncoming),
-				containerIdWithPrefix, networkInterface.EndpointId,
-			)
-			ch <- prometheus.MustNewConstMetric(
-				c.DroppedPacketsOutgoing,
-				prometheus.CounterValue,
-				float64(networkInterface.DroppedPacketsOutgoing),
-				containerIdWithPrefix, networkInterface.EndpointId,
-			)
-			break
-		}
-
 		ch <- prometheus.MustNewConstMetric(
 			c.ReadCountNormalized,
 			prometheus.CounterValue,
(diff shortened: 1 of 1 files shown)


collector/container.go
```diff
@@ -95,38 +96,38 @@ func newContainerMetricsCollector() (Collector, error) {
...
+			"Bytes Received on HNS endpoint",
+			[]string{"shared_containers", "endpoint_id"},
 			nil,
 		),
 		BytesSent: prometheus.NewDesc(
 			prometheus.BuildFQName(Namespace, subsystem, "network_transmit_bytes_total"),
-			"Bytes Sent on Interface",
-			[]string{"container_id", "interface"},
+			"Bytes Sent on HNS endpoint",
+			[]string{"shared_containers", "endpoint_id"},
```
tzifudzi, line 106: @jsturtevant Network stats are going to be per HNS endpoint. Each endpoint can be associated with zero or more container ids. 

We can associate each endpoint network metric with the label 
- `shared_containers`
- `container_ids`
- or something else

What would you recommend? It made sense to name the label `shared_containers` because that is the name the the underlying HNS component uses.

collector/container.go
```diff
@@ -254,53 +255,6 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
 			containerIdWithPrefix,
 		)
 
-		if len(cstats.Network) == 0 {
```
tzifudzi, line 257: Do we need backward compatibility with docker runtime using this manner of querying for network stats? Not sure if HNS query is going to work on docker runtime and I didn't get a chance to test it as of yet.

collector/container.go
```diff
@@ -327,6 +281,66 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
...
+	}
+
+	for _, endpoint := range hnsEndpoints {
+		endpointStats, err := hcsshim.GetHNSEndpointStats(endpoint.Id)
+		if err != nil {
+			log.Warnf("Failed to fetch stats for HNS endpoint %s", endpoint.Id)
+			continue
+		}
+
+		var sharedContainers = strings.Join(endpoint.SharedContainers, ",")
```
tzifudzi, line 303: Other metric labels contain a prefix of the runtime i.e. `containerd://...`  e.g. `windows_container_storage_read_size_bytes_total{container_id="containerd://85adcd..."} 2.569728e`

It makes sense to continue use this prefix, but the current change as is does not include a prefix and looks like
```
windows_container_network_receive_bytes_total{endpoint_id="cc1ce900-...",shared_containers="e5ce36c..."} 3.48512446e
```

Do we want to continue including the prefix for the container ids? This would mean the network stats labelling will depend on the HCS calls previously made to get the container info. If the HCS call to get container info fails then network stats labelling will also fail, and no metrics will be sent to the collector hence why in the initial change I have not included the prefix.

Shall we add the prefix based on the previous HCS calls or its fine to not include it? I can create a map to keep track of the container runtimes for the containers.

tzifudzi:
@jsturtevant Directly mentioning you here as I would appreciate your review.

collector/container.go
jsturtevant: This should stay the same otherwise downstream consumers will break.  Since we are putting the endpoint ID into the metric, it will be up to tooling to stitched it together properly.

This is the same thing that happens today on the docker v1 schema. If you were to look at a dockershim implementation each container would get the same value but would have the shared "interface". The tooling putting it together would need to know about that the containers share the interface and should group the together based on interface.

collector/container.go
jsturtevant: Yes we need backwards compat for awhile.  Docker is still using v1 api's so for folks that are using windows exporter without Kubernetes (aka just docker installed) will need to be supported.

Since the network in Docker uses v1, I am not sure if the stats will work from v2 api.  In kubernetes, the CNI's had all transitioned to using v2 api's so this wasn't a concern. 

We will need to test if the calls work, if not come up with a way make the distinction.

collector/container.go
jsturtevant: As mentioned above, this would be a breaking change, IMO we should continue to use the `contianerd_id` and `interface` for the labels. The tooling using this information should be able to aggregate this information together in a coherent way.

jsturtevant (commented):
Thanks for picking this up!

collector/container.go
tzifudzi: Got it, will make the change so we don't introduce breaking changes.

collector/container.go
tzifudzi: Got it, will make the change test on both container runtimes and make sure the logic supports both.

collector/container.go
tzifudzi: Got it, will make the suggested change as described so we don't introduce breaking changes.

collector/container.go
tzifudzi: Despite this code removal, should be good here for both Docker and containerd. 

If necessary I am willling to make the change to support the v1 api calls removed here, but I don't think its necessary as I have tested on both Docker and containerd and metrics pull through successfully. See PR description for more information regarding the environment I used during testing.

collector/container.go
tzifudzi: I am resolving this conversation. I have since reverted to the initial metric naming and labelling to avoid making a breaking change.

collector/container.go
tzifudzi: I am resolving this conversation. I have since reverted to the initial metric naming and labelling to avoid making a breaking change.

collector/container.go
jsturtevant: ah interesting!  I had a suspicion that this would work for both but wasn't sure.  I think the v2 api's have been released for long enough that we can drop the support 👍

collector/container.go
```diff
@@ -327,6 +286,71 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
...
+					containerPrefix, endpointId,
+				)
+				ch <- prometheus.MustNewConstMetric(
+					c.DroppedPacketsOutgoing,
+					prometheus.CounterValue,
+					float64(endpointStats.DroppedPacketsOutgoing),
+					containerPrefix, endpointId,
+				)
+			} else {
+				log.Warn("Failed to collect network stats for container %s", containerId)
```
jsturtevant, line 349: For readability, I might reverse this check to be something like:

```
if !ok {
log.Warn("Failed to collect network stats for container %s", containerId)
continue
}

do all metric stuff....
```

Also something to consider is did this failed to collect or did we get a network stat that we might not have a contianer for (due to a timing issue?) this message is probably ok but I think its a slight distinction between the two

collector/container.go
jsturtevant: just had a thought, did you test this with docker outside of k8s?  I think in K8s the cni's are all using v2 apis

jsturtevant:
@breed808 I think this is close, @tzifudzi is looking into one more thing https://github.com/prometheus-community/windows_exporter/pull/1218#discussion_r1237847087 but could we run CI against it?

collector/container.go
tzifudzi: @jsturtevant I have just concluded tests on a bare metal Windows host (not via Kubernetes) and metrics are showing as expected with no further changes. We should be good. 

See snippet of output of Prometheus metrics endpoint
```
# HELP windows_container_network_transmit_bytes_total Bytes Sent on Interface
# TYPE windows_container_network_transmit_bytes_total counter
windows_container_network_transmit_bytes_total{container_id="docker://062ac7f1925dd3a7054089b05bf9b2b8bdd4920d89fe3b944584e1dbf7e98065",interface="87AD7194-4445-4D5B-869F-5B6AF713C121"} 15745
```

Environment used in testing
- Docker Engine: 24.0.2
- OS Version: Windows Server 2022 Datacenter

breed808:
I suspect the merge conflict is preventing the CI checks from running, might be worth rebasing the feature branch on `master` and resolving the conflict.

Aside from that, great work on this PR so far! I must profess my lack of k8s+Windows experience, but the conversation in this PR is reassuring: there's been a strong focus on stability and testing :+1:

collector/container.go
tzifudzi: ### Suggestion 1
> For readability, I might reverse this check to be something like...

Refactored as suggested for better readability.

### Suggestion 2
> ...this message is probably ok but I think its a slight distinction between the two

This particular error would only occur if as you mentioned, we get a network stat that we do not have a container for. Ideally we could just fetch network endpoints and not have to match with a container id and avoid this potential scenario, but then this would not be backwards compatible with the old manner of specifying the container id and container runtime as labels for the metric. Either way, this scenario for failure should not really occur but it could due to a timing issue or something else. Am happy to change the error message if you have a better suggestion?

tzifudzi:
@breed808 Thanks for the great feedback. Conflicts have been resolved and I also re-tested the changes manually in nodes with different configurations as listed in the description. Let me know if that resolved the CI checks. Awaiting @jsturtevant's response for one last comment and  after that we should be good to go.

collector/container.go
jsturtevant: The changes lgtm, easier to read!

jsturtevant: as for the error message, with the change it seems to reads better to me and I think we are in good shape

collector/container.go
```diff
@@ -258,78 +265,97 @@ func (c *ContainerMetricsCollector) collect(ch chan<- prometheus.Metric) (*prome
...
+			}
 
-		for _, networkInterface := range networkStats {
 			ch <- prometheus.MustNewConstMetric(
 				c.BytesReceived,
 				prometheus.CounterValue,
-				float64(networkInterface.BytesReceived),
-				containerIdWithPrefix, networkInterface.EndpointId,
+				float64(endpointStats.BytesReceived),
+				containerPrefix, endpointId,
```
jsturtevant, line 325: nit: this looks to be the `containerIdWithPrefix` not just the containerPrefix which could be confusing.  If it was just the prefix this wouldn't work but looking above it is the the containerIdwithPrefix.

jsturtevant:
LGTM

breed808 (approved):
Looks good! Great work with this one.

Issue #1226, opened 2023-06-12
Allow multiple instances of the windows_exporter when using a library. 


mattdurham:
Allow the windows_exporter to be used as a library and allow multiple instances of the exporter running at time. The Grafana Agent embeds the windows_exporter and for the longest time we used it as a singleton but now needing the ability to run it multiple times.

breed808:
I believe this has been resolved from @jkroepke's recent work on the exporter. Let us know if this is still an issue.

PR #1227, opened 2023-06-12
Allow multiple instances of windows_exporter to run by removing globals.

mattdurham:
This is an attempt to solve https://github.com/prometheus-community/windows_exporter/issues/1226. This works by saving the output of `flags` func into an `interface{}` and then injecting that into the `builder` func. Removing the need for global variables, also allowing things to interject their own settings object before the build.

Entirely open to feedback and other ways to do it. Could also use a psuedo dependency injection and avoid the `interface{}` conversion. Something like
```
type Settings interface {
  GetMSSQLSettings() *MSSQLSettings
  ....
}

```


collector/init.go
```diff
@@ -2,406 +2,418 @@ package collector
 
 import "github.com/alecthomas/kingpin/v2"
 
-// collectorInit represents the required initialisation config for a collector.
-type collectorInit struct {
+// Initializer represents the required initialisation config for a collector.
```
mattdurham, line 5: This file had the most functional change, also not very happy with the name, can change to CollectorInit though linting complains about that.

breed808:
Apologies for the delay in reviewing this. I think it'd be worthwhile rebasing the feature branch on `master` to remove the duplicate commits and resolve the merge conflicts (unfortunately there's been some churn in the code with recent merges).

Once that's done I'll be happy to have a closer look at the change. I'm open to supporting multiple instances :+1:

jkroepke:
Fixed in #1324

PR #1241, opened 2023-07-10
Integrate perflib

### LINKED ISSUE #1240: Integrate perflib_exporter into windows_exporter
Hi,

writing this issue at this time seems like [perflib_exporter](https://github.com/leoluk/perflib_exporter) has some lacks of maintenance. It includes the maintenance of outdated dependencies and other bugs seen as issue or PR.

The author of the library confirms that he is not longer using windows in production.
https://github.com/leoluk/perflib_exporter/issues/25

I would like to ask if it make sense to merge the perflib_exporter into the windows_exporter to reduce the amount of unmaintained libraries. Since windows_exporter mostly depends on WMI metrics, it could be benefit of this.

jkroepke:
Fixes #1240 

What I do here:

1. Copy https://github.com/leoluk/perflib_exporter/tree/master/perflib to `/perflib`
2. Copy needed consts from https://github.com/leoluk/perflib_exporter/blob/master/collector/mapper.go into  `/perflib/consts.go`
3. Remove unused helper
4. Remove `HelpName` lookups, since `windows_exporter` do not use them
5. Introduce lazy-loading for `QueryNameTable` to reduce startup times.
6. Remove the additional loop at the `QueryNameTable` initialization.


jkroepke:
Implementations since from point of view, we may need some additional tests.

jkroepke:
I run some test on Windows Server, I go no issues with default settings.

perflib/perflib.go
```diff
@@ -209,8 +209,6 @@ func queryRawData(query string) ([]byte, error) {
 		return nil, fmt.Errorf("failed to encode query string: %v", err)
 	}
 
-	defer syscall.RegCloseKey(syscall.HKEY_PERFORMANCE_DATA)
```
breed808: Why were the calls to close `HKEY_PERFORMANCE_DATA` removed? Are they not necessary?

jkroepke: I reading the [Microsoft docs](https://learn.microsoft.com/en-us/windows/win32/perfctrs/using-the-registry-functions-to-consume-counter-data), `RegCloseKey` does not close the Registry call, but frees the DLLs. Since we constantly fetch from registry, I'm not sure if is an good idea to clear the cache and frees the DLLs on each loop. 

Looking at 

* Datadog agent, they do not [call](https://github.com/DataDog/datadog-agent/blob/599c6147dafafbb64772a238c679e1a56b1e0234/pkg/flare/archive_win.go#L43) `syscall.RegCloseKey`, too.
* [Munin](https://github.com/munin-monitoring/munin-node-win32/blob/791d8cd8d54f0d1177d456396f1719761f164930/src/plugins/PerfCounterMuninNodePlugin.cpp#L40), they do also not closing the Handle.
* [pywin32](https://github.com/mhammond/pywin32/blob/e99541164bb65cc76b08804c7a7bbdc52afa47d4/win32/Lib/win32pdhutil.py#L39) they do also not closing the Handle.


There is also a opened PR to remove this RegCloseKeys - https://github.com/leoluk/perflib_exporter/pull/43

breed808: Thanks, that's good context on the removal

breed808:
Looking at the commits the feature branch could do with a rebase (we're currently proposing to merge a merge commit); I have no objections to the change itself, it's looking good.

Once rebased, are you happy for me to merge this?

jkroepke:
> Once rebased, are you happy for me to merge this?

Let me take an other test, in case I do the rebase incorrectly.

jkroepke:
Everything seems working fine in default configuration.

image

jkroepke:
@leoluk this PR contains only a subset of perflib_exporter. Only the functions and vars that windows_exporter requires are part of this PR

I guess to maintain the `perflib_exporter`, it requires an more deeper knowledge about the WMI metrics itself which I do not have.

breed808 (approved):
@jkroepke are you happy for this to be merged? I have no further concerns or queries for this change.

jkroepke:
I'm happy, yes

PR #1245, opened 2023-07-12
fix: Update init container in windows exporter so that it works on both WS2019 and WS2022

bragi92:
A cluster can container multiple nodepools with both WS 2019 and WS 2022.

Deploying both the YAML's on one's cluster will leave some containers always in the init creation phase with errors. I have switched the init container to use the powershell ltsc nanoserver image which works on both WS 2019 and WS 2022 so that one daemonset deployment is enough.


mloskot:
@bragi92 
> A cluster can container multiple nodepools with both WS 2019 and WS 2022.

FYI, my AKS clusters that I mention in https://github.com/MicrosoftDocs/azure-docs/issues/112066  are only based on WS 2022

breed808 (approved):
Thanks for submitting this! You'll need to sign the commit (see the DCO link in the PR checks) in order to fix the failing DCO job.

bragi92:
Thanks @breed808 I have signed off on the commimt. Please take a look when you get a chance.

Issue #1236, opened 2023-07-06
Feature Request: Allow Textfile collector to collect file from multiple paths


LaurenceChau:
Currently textfile collector can only collect file from one directory. Can add a new feature to allow it to collect file from multiple directories thanks. 

suggestion for the config.yaml
```
collectors:
  enabled: textfile
collector:
  textfile:
    directories: 
    - C:\Program Files\windows_exporter\textfile_inputs
    - C:\Program Files\windows_exporter\textfile_inputs2
```

suggestion for the cli
```
--collector.textfile.directories 'C:\Program Files\windows_exporter\textfile_inputs,C:\Program Files\windows_exporter\textfile_inputs2'
```

DiniFarb:
I'd like to implement this. @breed808 would that be ok or is already someone working on it?

breed808:
I'm not working on it at the moment. You're welcome to submit a PR for it.

JDA88:
Will you then replace `--collector.textfile.directory` with `--collector.textfile.directories` for the CLI?
If they are both available it could be confusing and one of them should erase the other

DiniFarb:
Yeah, I was more thinking of expanding the `--collector.textfile.directory`command so that it can handle either one or multible directories. I have already started and for the CLI it would also be possible to just multiply the command like `--collector.textfile.directory="D:/test/t1" --collector.textfile.directory="D:/test/t2"` or would that be total confusing ?

JDA88:
Expanding the `--collector.textfile.directory` is probably the best option.
If some people find it confusing adding `--collector.textfile.directories` as an alias for a couple of version then make the singular one deprecated

I am personally not a big fan of the "multiply the command" approach

PR #1262, opened 2023-07-29
Textfile collector: collect files from multiple paths

DiniFarb:
Regarding #1236 and more

Added Features:
- collect files from multiple paths
- walkDir func for all paths to collect files from subdirs
- new flag command `collector.textfile.directories`

About the `collector.textfile.directories` commmand:
For now it is possible to use `collector.textfile.directories`, `collector.textfile.directory` or both. They have the same functionality and in case both commands would be used, the dirs are concatenated. I am not sure if that is ok so lets discuss 😉

As soon all is settled, I'll update the docs - `collector.textfile.md` 😊

### DIFF
--- a/collector/textfile.go
+++ b/collector/textfile.go
@@ -64,7 +64,7 @@ type textFileCollector struct {
 func newTextFileCollectorFlags(app *kingpin.Application) {
 	textFileDirectory = app.Flag(
 		FlagTextFileDirectory,
-		"Directory or Directories to read text files with metrics from.",
+		"DEPRECATED: Use --collector.textfile.directories",
 	).Default("").String()
 	textFileDirectories = app.Flag(
 		FlagTextFileDirectories,
--- a/collector/textfile.go
+++ b/collector/textfile.go
@@ -19,7 +19,6 @@ package collector
 import (
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -37,11 +36,13 @@ import (
 )
 
 const (
-	FlagTextFileDirectory = "collector.textfile.directory"
+	FlagTextFileDirectory   = "collector.textfile.directory"
+	FlagTextFileDirectories = "collector.textfile.directories"
 )
 
 var (
-	textFileDirectory *string
+	textFileDirectory   *string
+	textFileDirectories *string
 
 	mtimeDesc = prometheus.NewDesc(
 		prometheus.BuildFQName(Namespace, "textfile", "mtime_seconds"),
@@ -54,7 +55,7 @@ var (
 type textFileCollector struct {
 	logger log.Logger
 
-	path string
+	directories string
 	// Only set for testing to get predictable output.
 	mtime *float64
 }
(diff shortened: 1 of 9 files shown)


collector/textfile.go
```diff
@@ -250,109 +261,116 @@ func (cr carriageReturnFilteringReader) Read(p []byte) (int, error) {
...
+				return nil
 			}
-			if mf.Help == nil {
-				help := fmt.Sprintf("Metric read from %s", path)
-				mf.Help = &help
+			if !dirEntry.IsDir() && strings.HasSuffix(dirEntry.Name(), ".prom") {
+				_ = level.Debug(c.logger).Log("msg", fmt.Sprintf("Processing file: %s", path))
+				families_array, err := scrapeFile(path, c.logger)
+				if err != nil {
+					_ = level.Error(c.logger).Log("msg", fmt.Sprintf("Error screaping file: %q. Skip File.", path), "err", err)
```
breed808, line 283: `screaping` -> `scraping`

breed808 (commented):
Thanks for submitting this!

Is the plan to eventually deprecate the `collector.textfile.directory` flag in favour of the `collector.textfile.directories` flag? If so, it'd be worth adding a deprecation warning to the old flag (see the [process collector](https://github.com/prometheus-community/windows_exporter/blob/b0a1a38829fc5c8052ea41f9e414fe0dea7c9e47/collector/process.go#L80) for an example).

DiniFarb:
I wasn't quite sure about deprecating the `collector.textfile.directory` flag. Technically both flags could stay but it could lead to confusion. Maybe @JDA88 or @LaurenceChau want to give also their input about what flag would be best?  

I think the best way to go is as mentioned from @breed808 - adding a deprecation warning to `collector.textfile.directory` flag and then remove it a few versions later.

JDA88:
> I think the best way to go is as mentioned from @breed808 - adding a deprecation warning to `collector.textfile.directory` flag and then remove it a few versions later.

I agree, as long as the current `--collector.textfile.directory` continue to work like today for at least a few versions

collector/textfile.go
```diff
@@ -63,17 +64,27 @@ type textFileCollector struct {
 func newTextFileCollectorFlags(app *kingpin.Application) {
 	textFileDirectory = app.Flag(
 		FlagTextFileDirectory,
-		"Directory to read text files with metrics from.",
-	).Default(getDefaultPath()).String()
+		"DEPRECATED: Use --collector.textfile.directories",
+	).Default("").String()
```
breed808, line 68: I feel we should add `.Hidden()` here to discourage use of the flag for new users, similar to the deprecated flags in the `process` and other exporters.

breed808 (commented):
Looking good! I have one minor concern for the deprecated flag, but everything else is fine.
Thanks for updating this.

EDIT: It'd also be worth rebasing the feature branch on `master` to remove the merge commits.

DiniFarb:
I think this is ready to merge or does it need some further checks/reviews ?

breed808 (approved):
Looks good! Thanks for your efforts with this one.

Issue #1263, opened 2023-07-29
BUG: Breaking config file CLI argument changes 0.22 to 0.23


jadanah:
Looks like the changes from https://github.com/prometheus-community/windows_exporter/pull/1243 results in being unable to load a config file using a relative path e.g. `.\windows_exporter.exe --config.file=config.yml`

**0.22.0 - WORKING**
```
$ .\windows_exporter-0.22.0-amd64.exe --config.file=config.yml
time="2023-07-29T22:43:30+01:00" level=info msg="Loading configuration file: config.yml" source="config.go:37"
time="2023-07-29T22:43:30+01:00" level=warning msg="No where-clause specified for service collector. This will generate a very large number of metrics!" source="service.go:49"
time="2023-07-29T22:43:30+01:00" level=warning msg="No filters specified for process collector. This will generate a very large number of metrics!" source="process.go:59"
```

**0.23.1 - BROKEN**
```
$ .\windows_exporter-0.23.1-amd64.exe --config.file=config.yml
ts=2023-07-29T21:43:43.063Z caller=exporter.go:161 level=error msg="could not load config file" err="parse \"config.yml\": invalid URI for request"
```

DiniFarb:
Yeah my bad, I am gonne fix that. Thanks for raising this😊

PR #1265, opened 2023-07-31
fix: config file CLI argument for relative paths -> v0.23.1 

DiniFarb:
Fix for #1263, there was an unnecessary check with `url.ParseRequestURI`. If you throw in gibberish it'll fail anyways.


breed808 (approved):
Looks good, thanks for submitting this!

PR #1266, opened 2023-08-01
fix: Create a monitoring namespace if it does not exist in windows exporter daemonset

bragi92:


bragi92:
@breed808 Please take a look at this PR when you get a chance. The current YAML would fail to deploy if the monitoring namespace did not exist. I have updated it to create one if its doesn't exist and it is a no-op if it already does.

rashmichandrashekar (approved):

breed808 (approved):
Looks good :+1:

Issue #1258, opened 2023-07-22
v0.23.0/v0.23.1 - Windows service can't stop without an error


kago-dk:
When I try to stop the windows_exporter service (v0.23.0/v0.23.1), it always returns an error:
<img width="401" alt="image" src="https://github.com/prometheus-community/windows_exporter/assets/49529296/a210e083-3b51-4dcf-98bc-c6a1821b5836">

The event viewer does not provide a lot of details
_"The windows_exporter service terminated unexpectedly.  It has done this 1 time(s).  The following corrective action will be taken in 60000 milliseconds: Restart the service."_

* I have tried windows_exporter-0.23.1-amd64.msi (to rule out a configuration issue on my part) and windows_exporter-0.23.1-amd64.exe, and both approaches produce the same problem.
* I have tried with config.yml (my existing setup using windows_exporter-0.23.1-amd64.exe) and without config.yml (as windows_exporter-0.23.1-amd64.msi does), and both approaches produce the same problem.

windows_exporter-0.22.0-amd64.msi does not have a similar issue.

kago-dk:
Am I the only one seeing this issue?

DiniFarb:
This might be cos the svc is handled in a separate routine with the ini() function. Let me explain what I think:

The communication between the exe and windows happens in the `initiate.go` file strictly speaking in this function: 

https://github.com/prometheus-community/windows_exporter/blob/8d615956074d9329eb33ece5ec19fc00a8d2af48/initiate/initiate.go#L22-L43

So if the windows service is stopped the case `case svc.Stop, svc.Shutdown:` will take place. Send a stop signal to the main thread with `s.stopCh <- true` then break the loop and send a signal back to windows that it is about to stop, with `changes <- svc.Status{State: svc.StopPending}`. But this might never happen, if the main process ends very quickly after `s.stopCh <- true` the `changes <- svc.Status{State: svc.StopPending}` will never be reached cos the main process does not wait for finishing any subroutines. This is a matter of µs cos in the main thread after the stop is received only one message is logged:
https://github.com/prometheus-community/windows_exporter/blob/8d615956074d9329eb33ece5ec19fc00a8d2af48/exporter.go#L307-L312

[...]

PR #1276, opened 2023-08-22
fix: Windows service can't stop without an error

DiniFarb:
ok I was not entirely right with my theory in #1258. The real reason for the stop error is that this line is a nil pointer:
https://github.com/prometheus-community/windows_exporter/blob/8d615956074d9329eb33ece5ec19fc00a8d2af48/initiate/initiate.go#L34

This logline uses the global `logger` variable which is never set, it is only set inside of the `init()` function. Therefore the log attempt on line 34 will cause a nil pointer, the exe dies and no signal will be send and so on. 
This global variable was added with v23: https://github.com/prometheus-community/windows_exporter/commit/8509bc69a60bb6b608d390fabdf6e4d64b939b2e

Nevertheless, what I wrote in #1258 is also possible. I have tested with:
```go
loop:
	for {
		select {
		case c := <-r:
			switch c.Cmd {
			case svc.Interrogate:
				changes <- c.CurrentStatus
			case svc.Stop, svc.Shutdown:
				_ = logger.Info(100, "Service Stop Received")
				s.stopCh <- true
				time.Sleep(1 * time.Millisecond)
				changes <- svc.Status{State: svc.StopPending}
				break loop
			default:
				_ = logger.Error(102, fmt.Sprintf("unexpected control request #%d", c))
			}
		}
	}
	return
```
Only one milisecond was enough to cause the same error as in #1258.

So this PR includes:
 - fix: for global `logger` variable
 - prevention of possible error in the stop process

### DIFF
--- a/initiate/initiate.go
+++ b/initiate/initiate.go
@@ -32,14 +32,14 @@ loop:
 				changes <- c.CurrentStatus
 			case svc.Stop, svc.Shutdown:
 				_ = logger.Info(100, "Service Stop Received")
-				s.stopCh <- true
+				changes <- svc.Status{State: svc.StopPending}
 				break loop
 			default:
 				_ = logger.Error(102, fmt.Sprintf("unexpected control request #%d", c))
 			}
 		}
 	}
-	changes <- svc.Status{State: svc.StopPending}
+	s.stopCh <- true
 	return
 }
 
@@ -57,7 +57,7 @@ func init() {
 	}
 
 	if isService {
-		logger, err := eventlog.Open("windows_exporter")
+		logger, err = eventlog.Open("windows_exporter")
 		if err != nil {
 			os.Exit(2)
 		}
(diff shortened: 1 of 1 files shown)


breed808:
Thanks very much for looking into this! Would the same fix need to be applied on line 51 so the `logger.Error()` call in line 55 so the exporter doesn't crash in the event `svc.IsWindowsService()` returns an error?

E.G.

```go
	isService, err := svc.IsWindowsService()
	if err != nil {
		logger, err = eventlog.Open("windows_exporter")
		if err != nil {
			os.Exit(2)
		}
		_ = logger.Error(102, fmt.Sprintf("Failed to detect service: %v", err))
		os.Exit(1)
	}

```

DiniFarb:
Yeah, I was thinking first it does not really matter actually, cos if it comes to the error block of `svc.IsWindowsService()` a new instance of the eventlog logger is created and only used in this error block and since the app will exit the process in any way at this block it never comes to use the global var. But now I think it would actually be better and less confusing if the eventlog is always tied to the global variable. I'll update it as you suggested. 👍

breed808 (approved):
Thanks!

Issue #1300, opened 2023-09-29
Add revision to windows_os_info


Masterwow3:
Is: Currently, no Windows revision number is included in windows_os_info.

Would it be possible to output the Windows revision number in the windows_os_info?
The revision changes with every Windows update, so it is possible to determine how up-to-date a machine is.

Example:
````
# HELP windows_os_info OperatingSystem.Caption, OperatingSystem.Version
# TYPE windows_os_info gauge
windows_os_info{revision="4737",build_number="17763",major_version="10",minor_version="0",product="Microsoft Windows Server 2019 Standard",version="10.0.17763"} 1
````

Example in CMD:
![image](https://github.com/prometheus-community/windows_exporter/assets/14217001/41e2d2e3-77b9-4463-a502-1504d84eabbe)

DiniFarb:
Is someone working on that? If not, I'd like to implement it.

@breed808 ok if I work on this ?

PR #1317, opened 2023-10-28
feat(#1300): add revision to windows_os_info

DiniFarb:
Regarding: #1300 

Added revision to windows_os_info. It was already there in the registry `SOFTWARE\Microsoft\Windows NT\CurrentVersion` which is already opened, just needed to grab it.


breed808 (approved):
Looks good, thanks for this :+1:

Issue #1323, opened 2023-11-02
Add Windows host/instance aggregate interface counters for PacketsReceivedTotal and PacketsSentTotal


patricia-dibenedetto:
Currently, 'packets_received_total' and 'packets_sent_total' metrics are available per interface/nic, this is a request for these metrics to be made available per host/instance across all interfaces effectively aggregating them. Believe this is a small change to "collector/net.go" with minimal performance impact, if any. Open to feedback as to the desired name for these two (2) new metrics. Thank you for your consideration.

PR #1329, opened 2023-11-08
Add Windows host/instance aggregate interface counters for PacketsReceivedTotal and PacketsSentTotal

patricia-dibenedetto:
Add Windows host/instance aggregate interface counters for PacketsReceivedTotal and PacketsSentTotal. 
These host metrics are available in the Linux Node Exporter, so requesting these be added to Windows exporter for parity. 
See https://github.com/prometheus-community/windows_exporter/issues/1323

### DIFF
--- a/collector/net.go
+++ b/collector/net.go
@@ -53,6 +53,10 @@ type NetworkCollector struct {
 	PacketsSentTotal         *prometheus.Desc
 	CurrentBandwidth         *prometheus.Desc
 
+	// The total packets received/sent for the instance/node across all interfaces
+	NodePacketsReceivedTotal *prometheus.Desc
+	NodePacketsSentTotal     *prometheus.Desc
+
 	nicIncludePattern *regexp.Regexp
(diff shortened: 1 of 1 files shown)


breed808:
Are these proposed metrics distinct from the existing `windows_net_packets_received_total` and `windows_net_packets_sent_total` metrics, when disregarding the `nic` flag?

I.E. would the following two PromQL queries return identical results?

```
windows_net_node_packets_received_total
```
```
sum without (nic) (windows_net_packets_received_total)
```

If that's the case I don't believe we need to add the additional metrics, users can drop the `nic` label and aggregate by instance.

patricia-dibenedetto:
@breed808 In our use case of windows_exporter, we are not using PromQL to process this information but using an OTEL Prometheus receiver and we are not able to specify 'SUM without (nic)' type queries in the OTEL Prometheus receiver, therefore asking to introduce this minor change in the windows_exporter.

PR #1335, opened 2023-11-14
Check for nil strings and include wrappers around each.

mattdurham:
When trying to remove the fork grafana agent uses ran into an issue with some strings being nil via the config. Kingpin never allows this so it hasnt been an issue yet.

Created two wrapper methods to make it easier, I could have went with one new func and ! the call  but I think its cleaner to have two. I checked for any *string and *bool.

Tested the windows_exporter and the Grafana Agent using this without problem.

### DIFF
--- a/pkg/collector/exchange/exchange.go
+++ b/pkg/collector/exchange/exchange.go
@@ -13,6 +13,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/prometheus-community/windows_exporter/pkg/perflib"
 	"github.com/prometheus-community/windows_exporter/pkg/types"
+	"github.com/prometheus-community/windows_exporter/pkg/utils"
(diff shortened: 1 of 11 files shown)


mattdurham:
For example `oldSiteExclude` is only set from `NewFromFlag` so it will be always nil when dealing with using a config. Its very likely a few of these are never nil, but it was easier to wrap them all instead of figuring out the usage of each.

mattdurham:
The where clause fields and a few others look fine without the check but then I have to think on whether its safe or not.

mattdurham:
If you would prefer I not touch those strings, I can go back and check them.

jkroepke (approved):
LGTM 👍

jkroepke:
> If you would prefer I not touch those strings, I can go back and check them.

I just want to know, where are the nil values come from. But almost all "old" settings are affected here.

PR #1336, opened 2023-11-14
Do not abort if we do not know which user is running the process

kilokang:
There are a few cases when we want the exporter to run without knowing which user is running it that may fail because of the issue golang/go#37348

This PR prints a warning instead of aborting.


jkroepke (approved):
LGTM 👍

Thanks for your contribution.

PR #1337, opened 2023-11-15
Remove deprecated flags: --collector.*.*-blacklist / --collector.*.*-whitelist

jkroepke:
This PR removing deprecated flags (blacklist, whitelist) + deprecated flag on textfile collector.

Since our major version is still <1, the breaking change is acceptable. If not, we can postpone this.

### DIFF
--- a/pkg/collector/iis/iis.go
+++ b/pkg/collector/iis/iis.go
@@ -3,7 +3,6 @@
 package iis
 
 import (
-	"errors"
 	"fmt"
 	"regexp"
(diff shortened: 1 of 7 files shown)


breed808 (approved):
Looks good. I'd previously committed to removing these in the `v0.26.0` release of the exporter, though I feel enough time has passed to warrant the removal of the deprecated flags.

Feel free to merge this now or wait until after `v0.25.0` is released.

PR #1339, opened 2023-11-16
add sites to iss config constructor

mattdurham:
This resolves an issue where sites were not included, I spot checked all the other configs and they look solid.


jkroepke (approved):
Thanks.

PR #1342, opened 2023-11-17
CI: Add stale action

jkroepke:
windows_exporter has a lot of opened issue with may not longer interest. I would like to cleanup some of the approx 300 issues and staled PRs.

This action uses https://github.com/actions/stale . After 90 days of inactivity, a post will be written from gh actions. After another 30 days of interactivity, the issue will be closed.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,20 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+permissions:
+  issues: write
+  pull-requests: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v8
+        with:
+          stale-issue-message: 'This issue has been marked as stale because it has been open for 90 days with no activity. This thread will be automatically closed in 30 days if no further activity occurs.'
+          exempt-issue-labels: 'lifecycle/frozen'
+          days-before-stale: 90
+          days-before-close: 30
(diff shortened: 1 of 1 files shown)


breed808:
I had considered adding this action some time back, but couldn't decide on appropriate timings. Too low and it adds a burden on users raising the issues, which might discourage users from submitting/updating issues. I also want to avoid this tool resulting in issue duplicates (user creates issue with information, bot closes issue, new user creates new issue missing information from the first issue).

I'd be in favour of introducing this with larger timings, such as 6 months before marking as stale, and a 2-3 month duration before closing issues after that. Does that seem reasonable?

jkroepke:
Them times that I choiced here based on large projects like https://github.com/helm/helm

90/30 that are 3 months + 1 months. I guess there is an inactivity for 3 months, its already stale from my point of view.

Mention that after 120 days of interactivity, users still can post to closed issues and ask for reopen, because the issues are not locked.

special issues can have a lifecycle/frozen label. That excludes tickets from state action.

I feel 180/90 days a bit to long. If the bot posts the stale warning, its unliky that users will posts after 2-3 weeks. Mostly, they react in the first 2 weeks.

breed808:
Just to clarify, would issues closed by this proposed action be locked?

jkroepke:
Example: https://github.com/helm/helm/issues/9293

No locking, just closing.

breed808 (approved):
OK, as we're not preventing users from re-opening tickets I'm happy to proceed with this.

We can always tweak the timings later depending on user feedback.

PR #1345, opened 2023-11-20
fix(docs): Add textfile to enabled_collectors

### LINKED ISSUE #1344: README.md example implies `TEXTFILE_DIR` flag enables `textfile` collector
The example for service installation with the `.msi` package has the `TEXTFILE_DIR` argument but doesn't actually enable the `textfile` collector in the `ENABLED_COLLECTORS` argument, which may lead some to believe that providing `--collector.textfile.directories` will automatically enable this collector. 

image

This lead to some confusion for me personally, so I assume it may lead to further confusion if this example is not edited.

breed808:
Omission of the collector implied that setting `TEXTFILE_DIR` was sufficient to enable the collector, which is not true.

Resolves #1344


jkroepke (approved):
LGTM 👍

PR #1351, opened 2023-11-27
Add check for enable worker process so it is properly set.

mattdurham:
When using process with config this was not being set.


jkroepke:
Makes sense. Arrgh there are still issues from rework. 

Thanks!

jkroepke (approved):

mattdurham:
I think its all, I ran all the exporters with config that worked on my system and none break with defaults.

Issue #1349, opened 2023-11-25
Feature Request: Enhance textfile collector to trigger a script


DiniFarb:
### Introduction
I would like to propose an enhancement to the `textfile` collector. I am bringing this proposal here to gather feedback. If the community and maintainers are receptive to this idea, I would be willing to implement this feature by myself.

### Proposed Feature
The enhancement involves adding an option to the `textfile` collector to trigger a PowerShell script. The proposed option could look like this:

`--collector.textfile.trigger_ps_script=C:/My_textfile_generator.ps1`

This script would run each time the `/metrics` path is hit, generating one or multiple files for the normal `textfile` collector to use. The return of the script could be used to set the `error` metric of the collector and log some information.

### Rationale
I came up with this idea because I currently have several instances of `windows_exporter` where I use the `textfile` collector to gather additional information each time Prometheus collects metrics. To achieve this, I've been using the Windows Task Scheduler to trigger a script that generates these files in a time interval nearly synchronized with Prometheus's collection intervals. Incorporating this feature directly into the collector would improve:

- **Real-time Data:** Ensuring that metrics in Prometheus reflect the most recent state, as the text files would be updated at each collection cycle.
- **Simplified Workflow:** Eliminating the need for separate scheduling or manual intervention to update the text files, thus streamlining the metric collection process.

### Seeking Feedback
I am looking forward to feedback from the community and maintainers on this proposal. I am open to suggestions and willing to discuss further to refine this idea.

### Thoughts
In considering various approaches for this enhancement, another idea crossed my mind: the implementation of a separate collector named `scripts`. Instead of generating text files, this new collector could directly execute scripts, with the scripts output being the additional metrics themselves.

PR #1354, opened 2023-12-04
Textfile script trigger

DiniFarb:
PR explained in #1349


jkroepke:
Hi @DiniFarb 

sorry, I did not saw your proposal before. 

I'm not really happy to execute arbitrary scripts from windows_exporter. windows_exporter run often as high privileged user and if wrongly configured, it can lead the privileged escalation.

For script execution, I would recommend something like this https://github.com/ricoberger/script_exporter

jkroepke:
>  It would just have been so practical if you only had to install the windows_exporter for third party metrics on a Windows server and didn't have to start additional processes

Thats the use-case for https://github.com/grafana/agent. It embedded various exporters into one binary.

If you agree with this, I would like to close this.

PR #1357, opened 2023-12-05
Remove old event source registry key

### LINKED ISSUE #1318: The description for Event ID 3299 from source windows_exporter cannot be found
I have recently rolled out v0.24.0 to our servers (combination of Server 2016-2022) and have noticed that the even is filled with the following errors:

"The description for Event ID 3299 from source windows_exporter cannot be found. Either the component that raises this event is not installed on your local computer or the installation is corrupted. You can install or repair the component on the local computer.

If the event originated on another computer, the display information had to be saved with the event.

The following information was included with the event: 

ts=2023-10-30T07:26:44.404Z caller=tls_config.go:277 level=info msg="TLS is disabled." http2=false address=[::]:9182

The message resource is present but the message was not found in the message table"

As a test, I rolled back to v0.22.0 and the 3299 events no longer existed and ran with the usual event id 101/102.
I can see that if I have v0.22.0 installed the "Computer\HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\windows_exporter" has many more fields than v0.24.0

jkroepke:
fixes #1318

### DIFF
--- a/installer/windows_exporter.wxs
+++ b/installer/windows_exporter.wxs
@@ -29,7 +29,7 @@
     <SetProperty
             Id="WixQuietExecCmdLine"
             Value="&quot;[%ComSpec]&quot; /c reg delete HKLM\SYSTEM\CurrentControlSet\Services\EventLog\Application\windows_exporter /f"
-            Before="ConvolutedCoreCountCustomAction"
+            Before="RemoveEventSource"
             Sequence="execute"
     />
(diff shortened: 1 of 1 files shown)


breed808 (approved):
Looks good! I've tested this locally and the event descriptions no longer have the message described in #1318 :+1:

PR #1365, opened 2023-12-19
metric name & useful query

exgolden:
The naming convention for these metrics was different from that of the other collectors, so I renamed them following the same standard.
I also added a useful query to show the percentage of a specific volume in the instance
Signed-off-by: Isai Alejandro isai_alejandro@outlook.com


jkroepke (approved):
LGTM 👍

Issue #1352, opened 2023-11-30
Windows service can't stop without an error.


Techni-P:
With windows_exporter 0.24.0 we have again the same Issue like:
https://github.com/prometheus-community/windows_exporter/issues/1258

When I try to stop the windows_exporter service (v0.24.0), it always returns an error:
![image](https://github.com/prometheus-community/windows_exporter/assets/152503042/a12a6edd-42fe-4aef-a1c5-6d0ec6ef53de)
and the Process is still running in the Background.

jkroepke:
@DiniFarb this issue can be close, right?

PR #1372, opened 2024-01-06
refactor: svc.Run() as main routine blocker

DiniFarb:
### Regarding Issue #1352

I have made some tests with `svc.Run()` in the main Go routine here: https://github.com/DiniFarb/windows_exporter/commit/16888c2c0fdd8f8f954281a5e6b3b11de6128b04#diff-88d237fa0e6dd92a26d0dd0a14ad18246fb0924b68520062e9c9150e2004d7fe 
With that change, I could no longer reproduce the stop error. 
In that commit, the Windows service interface is initialized after the whole windows_exporter has already started. But the interface should actually be initialized as early as possible, so that the `svc.Status{State: svc.StartPending}` can be sent back before most of the service is loaded. Thus, I have refactored the startup process while I tried the following order:

1. Initialize Kingpin flags.
2. Load args/config file.
3. Initialize logger.
4. Initialize Windows service (if not CLI).
5. Build collectors.
6. Build HTTP server and start.

I know that this is not a small change, so I am very open to changing things to keep a better understanding of the startup process for everyone. :)


jkroepke:
I have the feeling that it reverts https://github.com/prometheus-community/windows_exporter/pull/1047 and re-introduce #551

DiniFarb:
Yes and in #1047 was also mentioned that this fix could affect the stop process (https://github.com/prometheus-community/windows_exporter/pull/1047#issuecomment-1224117744)

All in all it is not so easy to keep that in a good way. The solution from grafana/agant looks promising though. But this would be an even bigger change then I did.

I just realized how this stop issue can be fixed even easier without refactoring the main start process. With just moving the stop signal after the `svc.Run()` it is assured that the window interface can handle the stop process and then send the final stop to the main routine. Here would be the according changes: https://github.com/prometheus-community/windows_exporter/commit/9921ba31d179d814a9fef4de36a49b1ddcca5347

For now I could close this PR and add a new one with the small fix in 9921ba3. 
But yeah I agree, in the end it would be better to split it up like it is done in the grafana agent.

PR #1376, opened 2024-01-08
add call to set perf counter query

mattdurham:
I believe this has a good chance of fixing https://github.com/prometheus-community/windows_exporter/issues/1373 and https://github.com/prometheus-community/windows_exporter/issues/1374 and https://github.com/prometheus-community/windows_exporter/issues/1371.

PR #1377, opened 2024-01-08
CI: Provide binary builds from GH actions

jkroepke:
This PR uploads the windows_exporter.exe file from job lint as job artifact.

It give us the opportunity the ask different users, if PR like #1376 solves issues. Before this, users have build own exporter binaries which could be a blocker for inexperienced users.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -77,6 +77,15 @@ jobs:
       - name: Promtool
         run: make promtool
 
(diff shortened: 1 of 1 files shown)


jkroepke:
Example of CI artefact: https://github.com/prometheus-community/windows_exporter/actions/runs/7451612418?pr=1377

breed808 (approved):
Looks good!

PR #1378, opened 2024-01-08
Remove i386 builds

jkroepke:
This PR removes x86 binary builds.

The x86 binary builds are broken, because Windows provides only int32 values for PerfCounter which leads into user error, if the value is higher than 32-bit integer.

I would like to remove build to reduce the maintenance efforts.


breed808 (approved):
The change is good and I agree with the reasoning. While it might be technically possible to support this, I think maintainer time is better spent elsewhere.

Issue #1382, opened 2024-01-11
textfile_dir not inst


irnmn:
use command msiexec /i windows_exporter-0.25.0-amd64.msi ENABLED_COLLECTORS="cpu,cs,iis,logical_disk,memory,net,os,process,scheduled_task,service,system,tcp,textfile" LISTEN_PORT="9182" TEXTFILE_DIR="C:\custom_metrics\" freezing process in progress
But if you run without the last command, then everything is successful

DiniFarb:
This my fault, because as I was implementing #1262 (v0.24.0) I forgot to update the wxs https://github.com/prometheus-community/windows_exporter/blob/6ede10e29aeea4e5aa250a20ade395ce14058fdc/installer/windows_exporter.wxs#L47-L48

With  #1262 the flag `--collector.textfile.directory` was marked deprecated and then in the next release removed by #1337 (v0.25.0)

@breed808 should I bump in a PR to fix? If so my question would be, should I also update the ID like `Id="TEXTFILE_DIRS"` instead of `Id="TEXTFILE_DIR"`

breed808:
Yes please, a PR is welcome. I'd go with `Id="TEXTFILE_DIRS"` as that clearly indicates that multiple directories can be supplied. The `TEXTFILE_DIR` reference in the main README will need to be updated too.

irnmn:
will there be a new version 0.25.1 or what?

jkroepke:
https://github.com/prometheus-community/windows_exporter/releases/tag/v0.25.1 - please let us know, if you issue is resolved

eklunko:
> The `TEXTFILE_DIR` reference in the main README will need to be updated too.

Possibly the `Description` cell for TEXTFILE_DIRS in the main README needs to be updated too.

PR #1384, opened 2024-01-12
fix:(#1382): rename TEXTFILE_DIR to TEXTFILE_DIRS

DiniFarb:
As announced in #1382 the according fix.

And I was wondering if deprecation warning here should be removed or changes since it argument is gone now?
https://github.com/prometheus-community/windows_exporter/blob/6ede10e29aeea4e5aa250a20ade395ce14058fdc/docs/collector.textfile.md?plain=1#L12-L16


jkroepke:
Thank you.

The old flag was deprecated long time, but only visible at start-up logs.

jkroepke (approved):

Issue #1389, opened 2024-01-16
Version 0.25.1 Ignores the Collect Parameter


mwallace78:
When using the collect parameter in the Prometheus config to filter collectors, all enabled collectors are still being returned.

On a system that has 0.25.1 installed with all collectors enabled at installation time, we can see them all attempting to return even though we are only passing "cpu" as a collect param:
```
curl 'http://<server>:9182/metrics?collect%5B%5D=cpu' | grep windows_exporter_collector_success
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
# HELP windows_exporter_collector_success windows_exporter: Whether the collector was successful.
# TYPE windows_exporter_collector_success gauge
windows_exporter_collector_success{collector="ad"} 0
windows_exporter_collector_success{collector="adcs"} 0
windows_exporter_collector_success{collector="adfs"} 0
windows_exporter_collector_success{collector="cache"} 1
windows_exporter_collector_success{collector="container"} 0
windows_exporter_collector_success{collector="cpu"} 1
windows_exporter_collector_success{collector="cpu_info"} 1
windows_exporter_collector_success{collector="cs"} 1
windows_exporter_collector_success{collector="dfsr"} 1
windows_exporter_collector_success{collector="dhcp"} 0
windows_exporter_collector_success{collector="dns"} 0
windows_exporter_collector_success{collector="exchange"} 0
windows_exporter_collector_success{collector="fsrmquota"} 0
windows_exporter_collector_success{collector="hyperv"} 0
windows_exporter_collector_success{collector="iis"} 0
windows_exporter_collector_success{collector="logical_disk"} 1
windows_exporter_collector_success{collector="logon"} 1
windows_exporter_collector_success{collector="memory"} 1
windows_exporter_collector_success{collector="mscluster_cluster"} 0
windows_exporter_collector_success{collector="mscluster_network"} 0
windows_exporter_collector_success{collector="mscluster_node"} 0
windows_exporter_collector_success{collector="mscluster_resource"} 0
windows_exporter_collector_success{collector="mscluster_resourcegroup"} 0
windows_exporter_collector_success{collector="msmq"} 0
windows_exporter_collector_success{collector="mssql"} 0
windows_exporter_collector_success{collector="net"} 1
[...]

PR #1392, opened 2024-01-21
fix: Apply query filter to metrics

breed808:
Refactor in 0711268d3c3025b5b6ed49b3862264b7e7600bec introduced #1389, where the `collect[]=` parameter is being ignored, preventing users from filtering metrics.

This change ensures the collector filter is applied.


breed808:
@jkroepke can you have a look at this one when you have time?

jkroepke:
Sorry, I will looked into this - this week.

jkroepke (approved):
Tested. And works fine.

Only the output is filtered, not sure if this is expected.

jkroepke:
Sorry for been so long unresponsive here.

PR #1396, opened 2024-01-26
docs!: Remove support for EOL Windows versions

breed808:
Microsoft currently support Windows Server 2016 or newer, and Windows 10 and Windows 11 (21HR or later). Dropping support for end-of-life Windows Server versions will reduce maintenance overhead for project maintainers.

This removal of support is intended for the next major exporter release (v0.26.0).

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,8 @@ See detailed steps to install on Windows Kubernetes [here](./kubernetes/kubernet
 
 `windows_exporter` supports Windows Server versions 2016 and later, and desktop Windows version 10 and 11 (21H2 or later).
 
+Windows Server 2012 and 2012R2 are supported as best-effort only, but not guaranteed to work.
+
 ## Usage
 
     go get -u github.com/prometheus/promu
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ See detailed steps to install on Windows Kubernetes [here](./kubernetes/kubernet
 
 ## Supported versions
 
-windows_exporter supports Windows Server versions 2008R2 and later, and desktop Windows version 7 and later.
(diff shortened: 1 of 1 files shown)


jkroepke:
JFYI:

Technically, Windows Server 2012+ are still supported as part of Extended Security Update (https://learn.microsoft.com/en-us/lifecycle/announcements/windows-server-2012-r2-end-of-support)

Any Windows Server on Azure gets them on free and they are tons of it.

breed808:
That's a good point.
Perhaps we could clarify that Windows Server 2012 with ESU is supported? Or just support 2012 for the duration of the ESU period?

jkroepke:
Not sure if its an difference for us. APIs are the same. And we can't proof if a report has ESU or not.

I would go with best effort/not garantueed suppor for 2012, since go itself drops 2012 already

README.md
```diff
@@ -140,7 +140,7 @@ See detailed steps to install on Windows Kubernetes [here](./kubernetes/kubernet
 
 ## Supported versions
 
-windows_exporter supports Windows Server versions 2008R2 and later, and desktop Windows version 7 and later.
+`windows_exporter` supports Windows Server versions 2016 and later, and desktop Windows version 10 and 11 (21H2 or later).
```
jkroepke, line 143:
```suggestion
`windows_exporter` supports Windows Server versions 2016 and later, and desktop Windows version 10 and 11 (21H2 or later).

Windows Server 2012 and 2012R2 are supported as best-effort only, but not guaranteed to work.
```

jkroepke (approved):
At least there are tons of 2012R2 machines which getting support from Microsoft for free (ESU), however [go](https://tip.golang.org/doc/go1.20#windows) it does not support 2012R2 anymore.

I would say, we officially support 2016. But If some contribute a fix for 2012R2, I would like to accept it.

PR #1442, opened 2024-04-16
Add label volume_name to logical_disk collector

aburtasov:
Added label volume_name using WIN32_LogicalDisk to logical_disk collector

Example metric with new label:

HELP windows_logical_disk_free_bytes Free space in bytes, updates every 10-15 min (LogicalDisk.PercentFreeSpace)
TYPE windows_logical_disk_free_bytes gauge
windows_logical_disk_free_bytes{volume="C:",**volume_name**="Local Disk"} 1.9264438272e+10
windows_logical_disk_free_bytes{volume="E:",**volume_name**="EXCH9-E"} 3.5105275904e+10

### DIFF
--- a/pkg/collector/logical_disk/logical_disk.go
+++ b/pkg/collector/logical_disk/logical_disk.go
@@ -423,8 +423,8 @@ func (c *collector) collect(ctx *types.ScrapeContext, ch chan<- prometheus.Metri
 					logicalDisk.VolumeName,
 				)
 
-			} else {
-				continue
+				break
+
 			}
 		}
 
--- a/pkg/collector/logical_disk/logical_disk.go
+++ b/pkg/collector/logical_disk/logical_disk.go
@@ -3,6 +3,7 @@
 package logical_disk
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
 
@@ -11,6 +12,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/prometheus-community/windows_exporter/pkg/perflib"
 	"github.com/prometheus-community/windows_exporter/pkg/types"
+	"github.com/prometheus-community/windows_exporter/pkg/wmi"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/collector/logical_disk/logical_disk.go
```diff
@@ -271,118 +290,144 @@ func (c *collector) collect(ctx *types.ScrapeContext, ch chan<- prometheus.Metri
...
+				ch <- prometheus.MustNewConstMetric(
+					c.ReadWriteLatency,
+					prometheus.CounterValue,
+					volume.AvgDiskSecPerTransfer*perflib.TicksToSecondScaleFactor,
+					volume.Name,
+					logicalDisk.VolumeName,
+				)
+
+			} else {
+				continue
```
jkroepke, line 427: this this an error path? If yes, we should throw any error here instead silent continue

jkroepke (changes requested):
Work, tested with Emojis.

image

pkg/collector/logical_disk/logical_disk.go
aburtasov: Hello! 
No, this is just processing the loop condition. In this loop we are looking for a match of volume(for example, C: == C: ). And if we find a match, then we write down the metrics, if a match is not found at the current iteration, we continue to search

pkg/collector/logical_disk/logical_disk.go
```diff
@@ -271,118 +290,144 @@ func (c *collector) collect(ctx *types.ScrapeContext, ch chan<- prometheus.Metri
...
+				)
+
+				ch <- prometheus.MustNewConstMetric(
+					c.ReadWriteLatency,
+					prometheus.CounterValue,
+					volume.AvgDiskSecPerTransfer*perflib.TicksToSecondScaleFactor,
+					volume.Name,
+					logicalDisk.VolumeName,
+				)
+
```
jkroepke, line 425:
```suggestion
				break
```

In case the letter was found, can stop the loop here.

aburtasov: Looks more optimal. Fixed cycle

jkroepke (approved):
Thanks

jacbo0112:
Looking forward to the official version

truck0321:
I noticed after building and testing from master that the logical disks collector stopped showing user profile disk metrics.  I'm not sure if it was from this change yet but thought I'd give a heads up here first.  

v25.1
image

master
image

truck0321:
I submitted issue #1493

PR #1445, opened 2024-04-20
Implement a Windows application manifest

jkroepke:
I'm using github.com/tc-hib/go-winres to generate an application manifest to archive a version number on the exe file.

Unbenannt

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ output/
 .vscode
 .idea
 *.syso
+app.syso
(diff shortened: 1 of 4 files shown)


breed808 (approved):
Looks like it'll need a rebase on `master` (apologies for the delayed review!), but other than that it looks good :+1:

jantari:
Thank you a lot for this! Binaries without version metadata are the worst :laughing:

PR #1447, opened 2024-04-20
installer: Disable Windows Firewall Exception by default. Can be re-enabled with ADD_FIREWALL_EXCEPTION=yes

### LINKED ISSUE #513: Allow to disable firewall configuration
Please add ability to skip firewall configuration. Every time after installation I'm deleting that record. We have our own firewall rules.

Also it's not secure

jkroepke:
Fixes #513 

Signed-off-by: Jan-Otto Kröpke <mail@jkroepke.de>

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[*.wxs]
+
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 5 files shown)


breed808 (approved):
Looks good! We'll need to mention this in the next release's change notes, I suspect there are quite a number of users relying on the default firewall rule, though the rule created from an earlier installation should persist, if I recall correctly.

PR #1450, opened 2024-04-21
iis: Add IIS info metric

### LINKED ISSUE #1449: [requirements]   IIS version number
There is no version number in the returned data of IIS. I see that the code has obtained the version, but there is no data to return the version number. Can this be added.

jkroepke:
Fixes #1449

### DIFF
--- a/pkg/collector/iis/iis.go
+++ b/pkg/collector/iis/iis.go
@@ -1532,15 +1532,6 @@ func (c *collector) collectW3SVC_W3WP(ctx *types.ScrapeContext, ch chan<- promet
 		}
 
 		ch <- prometheus.MustNewConstMetric(
-			c.Info,
-			prometheus.GaugeValue,
-			app.Threads,
-			name,
-			pid,
-			"busy",
-		)
-
-		ch <- prometheus.MustNewConstMetric(
 			c.Threads,
 			prometheus.GaugeValue,
 			app.Threads,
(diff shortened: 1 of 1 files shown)


jkroepke:
Tested by https://github.com/prometheus-community/windows_exporter/issues/1449#issuecomment-2071210233

pkg/collector/iis/iis.go
```diff
@@ -1514,6 +1531,15 @@ func (c *collector) collectW3SVC_W3WP(ctx *types.ScrapeContext, ch chan<- promet
 			continue
 		}
 
+		ch <- prometheus.MustNewConstMetric(
```
breed808, line 1534: Is this section necessary? We appear to be adding to the channel twice for the same metric here.

pkg/collector/iis/iis.go
```diff
@@ -1514,6 +1531,15 @@ func (c *collector) collectW3SVC_W3WP(ctx *types.ScrapeContext, ch chan<- promet
...
+			c.Info,
+			prometheus.GaugeValue,
+			app.Threads,
+			name,
+			pid,
+			"busy",
+		)
+
```
jkroepke, line 1542:
```suggestion
```

pkg/collector/iis/iis.go
jkroepke: Seems like there was an mistake/leftover from copy/paste

breed808 (approved):
:+1:

PR #1451, opened 2024-04-21
scheduled_task: Move OLE connection to collect function

### LINKED ISSUE #1425: Issue with scheduled_task
The scheduled_task returns `CO_E_NOTINITIALIZED`  when I try to use it. I think it is getting initialized in build but then uninited.  if I remove this line https://github.com/prometheus-community/windows_exporter/blob/master/pkg/collector/scheduled_task/scheduled_task.go#L130 it works as expected.

jkroepke:
Based on: https://github.com/influxdata/telegraf/blob/2de80263d5875531b5c39eb63c85b20109a35799/plugins/inputs/win_wmi/query.go#L77

Fixes #1425

### DIFF
--- a/pkg/collector/scheduled_task/scheduled_task.go
+++ b/pkg/collector/scheduled_task/scheduled_task.go
@@ -3,6 +3,7 @@
 package scheduled_task
 
 import (
+	"errors"
 	"fmt"
 	"regexp"
 	"runtime"
@@ -63,7 +64,6 @@ const (
 	TASK_RESULT_SUCCESS TaskResult = 0x0
 )
 
-// RegisteredTask ...
 type ScheduledTask struct {
 	Name            string
 	Path            string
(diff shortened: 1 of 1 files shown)


jkroepke:
@mattdurham does fix the issue on your side?

Connecting to OLE during collect should also increase the support for multiple instances and reduce the locking to an minimum.

On my local system, I do not longer had issues.

mattdurham:
I think it should but will verify

breed808:
@mattdurham did you manage to verify the change?

jkroepke:
@breed808 otherwise, if you agree - take a review and merge it.

It was wrong anyways to construct the OLE init on builder func. If the issue is not resolved, feedback can be applied in https://github.com/prometheus-community/windows_exporter/issues/1425

breed808 (approved):
Looks good to me :+1:

PR #1469, opened 2024-05-06
Add: push docker images to dockerhub and quay.io

jkroepke:
Fixes #1476 

This PR introduce the image push to dockerhub and quay.io. 

~~There is a new image flavor (hostprocess image) which is designed for hostprocesses. It works on any OS.~~ - It's only possible to build that image with buildkit which is not available under Windows

Additionally, the release pipelines now includes pushes from master branch. The release process will now continuously tested and we are able to gain faster feedback from the community.

I'm expected some push error after merge, since I'm not a makefile pro.


SuperQ (approved):
Nice

breed808 (approved):
:+1:

PR #1471, opened 2024-05-07
Remove landing page

### LINKED ISSUE #1465: Introduce a flag to disable the default home page
Multiple security audits got triggered on our company because they found the windows exporter service responded with an html page at the root and exposed the endpoint from there (`/metrics`, `/health`, `/version`)

Even though I’m not a big fan of security by obfuscation maybe introducing a flag to disable the home page (`/`) could make sense and make the exporter less visible in enterprise environment.

jkroepke:
Fixes #1465

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -211,34 +211,6 @@ func main() {
 		mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
 	}
 
-	if *metricsPath != "/" && *metricsPath != "" {
-		landingConfig := web.LandingConfig{
-			Name:        "Windows Exporter",
-			Description: "Prometheus Exporter for Windows servers",
-			Version:     version.Info(),
-			Links: []web.LandingLinks{
-				{
-					Address: *metricsPath,
-					Text:    "Metrics",
-				},
-				{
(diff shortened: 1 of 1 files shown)


exporter.go
```diff
@@ -64,6 +64,10 @@ func main() {
 			"web.disable-exporter-metrics",
 			"Exclude metrics about the exporter itself (promhttp_*, process_*, go_*).",
 		).Bool()
+		disableExporterLadingPage = app.Flag(
```
breed808, line 67:
```suggestion
		disableExporterLandingPage = app.Flag(
```

breed808 (commented):
Given `node_exporter` exposes a similar landing page by default, my preference would be to have this flag default to `true`. Thoughts?

jkroepke:
> my preference would be to have this flag default to `true`.

I guess then, we can remove the entire feature. Who will set this flag to true?

breed808:
I was thinking that too. Let's remove it and we can always revert if there's negative feedback regarding the removal.

breed808 (approved):
:+1:

PR #1475, opened 2024-05-11
Remove unused `*prometheus.Desc` return value from collectors `collect()` function

breed808:
Added relevant CI checks to prevent this from being accidentally re-introduced.

Uneven logging pairs have also been fixed.


jkroepke (approved):
Looks good to me

PR #1476, opened 2024-05-11
Add docker-repo-name make target

SuperQ:
Add the `docker-repo-name` Makefile target that is used by the container description sync action.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
-export GOOS=windows
-export DOCKER_IMAGE_NAME ?= windows-exporter
-export DOCKER_REPO ?= ghcr.io/prometheus-community
+GOOS=windows
+DOCKER_IMAGE_NAME ?= windows-exporter
+DOCKER_REPO ?= prometheuscommunity
 
 VERSION?=$(shell cat VERSION)
 DOCKER?=docker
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -66,3 +66,8 @@ push:
 	$(DOCKER) manifest push --purge $(DOCKER_REPO)/$(DOCKER_IMAGE_NAME):$(VERSION)
 
 push-all: build-all push
+
+.PHONY: docker-repo-name
+docker-repo-name:
+	@echo "$(DOCKER_REPO)/$(DOCKER_IMAGE_NAME)"
```
jkroepke, line 72:
```suggestion
	@echo "prometheuscommunity/$(DOCKER_IMAGE_NAME)"
```

Otherwise, the output would result into: `ghcr.io/prometheus-community/windows-exporter`

SuperQ: This is due to the Makefile including `export DOCKER_REPO ?= ghcr.io/prometheus-community`.

We should probably fix the publishing to default `DOCKER_REPO` to a DockerHub compatible name.

jkroepke: I have no clue who decide to push windows_exporter to ghcr only and not following the standards. 

If you agree, I will include the change into the PR #1469 (https://github.com/prometheus-community/windows_exporter/pull/1469/commits/8b6c36bd892f61a727e55238976d839d7f2cac2e)

PR #1487, opened 2024-05-14
Additional CI linters

breed808:
Update `golangci-lint-action` to the latest version.

Add `prealloc` to check for occasions where slices can be pre-allocated.

Add `perfsprint` to check for unneeded uses of `fmt.Sprintf` or uses that can be replaced with faster alternatives.


jkroepke:
I really like it that we enable more linters.

For my projects, I have inverted the defaults.

Enable all linters, except some. What did you think about this?

https://github.com/jkroepke/openvpn-auth-oauth2/blob/main/.golangci.yaml#L1-L36

breed808:
I've done the same for any private work projects, so I don't mind reversing the logic. I've made this change and enabled a couple other linters.

I'll continue work enabling more linters in future PR(s).

.github/workflows/lint.yml
```diff
@@ -98,9 +98,9 @@ jobs:
           go-version-file: 'go.mod'
 
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v3
+        uses: golangci/golangci-lint-action@v6
         with:
-          version: v1.55.2
+          version: v1.58
```
jkroepke, line 103:
```suggestion
          args: --out-format "github-actions,colored-line-number"
          version: v1.58
```

With that, we could gain useful output in just one run and we could avoid a second run.

jkroepke (approved):
I love more strictness! Great PR!

.github/workflows/lint.yml
breed808: Done!

PR #1488, opened 2024-05-14
feat: allow setting of exporter process priority

### LINKED ISSUE #724: No data collected when CPU is overloaded
Hi team!
Can you please give me an advice how to collect metrics when host's CPU is overloaded?

As you can see in the screenshot below, no data has been collected between the red lines.
The query is 
`100 - (avg by (instance) (irate(windows_cpu_time_total{mode="idle", instance=~"my_host"}[1m])) * 100)`
image

Is this the exporter issue or a problem with my local configuration?

My Prometheus config:

```
  - job_name: 'windows_exporter' 

    scrape_interval: 25s
    scrape_timeout: 20s
```
 
And windows_exporter config:
```
collectors:
  enabled: cpu,cs,logical_disk,net,os,service,system,process,iis
log:
  level: warn
scrape:
  timeout-margin: 0.1
collector:
  service:
    services-where: (NAME = 'windows_exporter') OR (Name = 'W3SVC') OR (Name = 'Redis')
  process:
    whitelist: windows_exporter|w3wp
```

breed808:
Hopefully this will mitigate or resolve #724

### DIFF
--- a/exporter.go
+++ b/exporter.go
@@ -31,7 +31,11 @@ import (
 	"github.com/prometheus/common/version"
 	"github.com/prometheus/exporter-toolkit/web"
 	webflag "github.com/prometheus/exporter-toolkit/web/kingpinflag"
+	"golang.org/x/sys/windows"
 )
+
+// https://learn.microsoft.com/en-us/windows/win32/procthread/process-security-and-access-rights
+const PROCESS_ALL_ACCESS = windows.STANDARD_RIGHTS_REQUIRED | windows.SYNCHRONIZE | windows.SPECIFIC_RIGHTS_ALL
 
 // Same struct prometheus uses for their /version endpoint.
 // Separate copy to avoid pulling all of prometheus as a dependency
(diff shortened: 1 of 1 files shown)


exporter.go
```diff
@@ -145,6 +179,16 @@ func main() {
 		return
 	}
 
+	// Only set process priority if a non-default and valid value has been set
+	if *processPriority != "normal" && priorityStringToInt[*processPriority] != 0 {
```
jkroepke, line 183: Does `priorityStringToInt[*processPriority]` results into a panic, if `--process.priority=invalid` is given?

exporter.go
```diff
@@ -31,8 +31,12 @@ import (
 	"github.com/prometheus/common/version"
 	"github.com/prometheus/exporter-toolkit/web"
 	webflag "github.com/prometheus/exporter-toolkit/web/kingpinflag"
+	"golang.org/x/sys/windows"
 )
 
+// https://learn.microsoft.com/en-us/windows/win32/procthread/process-security-and-access-rights
+const PROCESS_ALL_ACCESS = windows.STANDARD_RIGHTS_REQUIRED | windows.SYNCHRONIZE | 0xffff
```
jkroepke, line 38: There is a const for 0xffff

https://github.com/golang/sys/blob/6943ab67c485d110829e8c77c3a46235a07d73b9/windows/security_windows.go#L1002

exporter.go
breed808: If `processPriority` is ever `nil`, it'd result in a nil pointer de-reference. However a default value or user-supplied value is always set so that's not an issue.

If an invalid value such as `*processPriority="invalid"` is provided, it will result in the default `uint32` value of 0 and be ignored.

exporter.go
breed808: Good spot, I've added that in

jkroepke (approved):
Tested locally

PR #1490, opened 2024-05-14
Remove push to quay.io

jkroepke:
We don't know the robots account name if quay,io.

I asked in Slack https://cloud-native.slack.com/archives/C01AUBA4PFE/p1715422584716549 but this block us to publish and release a new RC.


breed808 (approved):
Looks fine to me. You'll want to sign the commit to pass that DCO check

Issue #1458, opened 2024-04-25
Windows service collector takes over 500s to scrape data


paologallinaharbur:
We have been noticing the exporter taking more time to scrape data than expected both if executed with `use-api` or with the old `WMI` while filtering services
```
$ windows_exporter.exe --collectors.enabled service --log.level debug --collector.service.services-where "Name='***'" 
or
$ windows_exporter.exe --collectors.enabled service --log.level debug --collector.service.use-api
```

We have a "normal" number of services that are scraped, but the exporter takes a lot to answer, in particular I find weird the following metrics
```
windows_exporter_collector_duration_seconds{collector="service"} 0.434567
windows_exporter_perflib_snapshot_duration_seconds 528.34678
```

I have the feeling that scraping the services is "quick" but crating a perflib snapshot wastes a lot of time, [even thought the service has an empty list of perfCounters
](https://github.com/prometheus-community/windows_exporter/blob/f6b91e5cb0f710550777c7ff261c21e6296c27a9/pkg/collector/service/service.go#L87-L88)
 - Do you have any idea why this could be happening?
 - Is the perflib snapshot creation something that be avoided somehow? As far as I know it reports the time taken by [this function](https://github.com/prometheus-community/windows_exporter/blob/f6b91e5cb0f710550777c7ff261c21e6296c27a9/pkg/perflib/utils.go#L10) to be executed, but I was unable to understand more 😕

jkroepke:
That feels strange, because service collector doesn't interact with perflib

jkroepke:
@paologallinaharbur by any chance, I would like to know, if #1497 help you.

> Do you have any idea why this could be happening?

WMI is slow, and the API based approach is asking each single service about status and query.

#1497 is using a different approach by asking the Windows API once. On my local system, it's quite fast but It wont provide all information, like run_as, pid or start mode

jkroepke:
It could be possible that query nothing takes up to 300 seconds on your system.

jkroepke:
Maybe you are hitting a bug in windows_exporter.

With 

> windows_exporter.exe --collectors.enabled service

you only have enabled a collectors which does not register and request any perflib counters.

This leads to an situation where is `where` clause gets empty. (This is what I mean with query nothing). I guess thats unknown behavior here. An empty where clause cloud also leads to return all data. But I'm unable to verify that.

[...]

PR #1497, opened 2024-05-17
Service V2 collector

jkroepke:
Ref: #1458 (@paologallinaharbur needs to confirm this as fix)

The old service collector via WMI query doesn't perform well. 
The alternative approach (using the API direct) is an better improvement, however it needs time, because is asks each single service about state and configuration. That doesn't scale well. Additionally, admin permissions is mandatory. 

I introduce a V2 collector here, because it's an breaking change in terms of metrics. (labels are different and information are omit)

The code based on https://github.com/DataDog/datadog-agent/blob/afbd8b6c87939c92610c654cb07fdfd439e4fb27/pkg/util/winutil/scmmonitor.go#L59

# Performace

WMI:
windows_exporter_collector_duration_seconds{collector="service"} 0.4285152

API:
windows_exporter_collector_duration_seconds{collector="service"} 0.3175442

V2:
windows_exporter_collector_duration_seconds{collector="service"} 0.0046183

---

### DIFF
--- a/docs/collector.service.md
+++ b/docs/collector.service.md
@@ -2,11 +2,9 @@
 
 The service collector exposes metrics about Windows Services
 
-|||
--|-
-Metric name prefix  | `service`
-Classes             | [`Win32_Service`](https://msdn.microsoft.com/en-us/library/aa394418(v=vs.85).aspx)
-Enabled by default? | Yes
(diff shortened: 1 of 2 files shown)


breed808 (approved):
:+1:

PR #1499, opened 2024-05-18
Fix timezone caching issues

### LINKED ISSUE #1486: windows_os_timezone not updating
The metric windows_os_timezone seems to only update with a restart of the windows exporter service, which is kind of useless when you want to monitor/alert if the timezone ever changes to something it shouldn't 😅

jkroepke:
Fixes #1486

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/prometheus-community/windows_exporter
 
-go 1.21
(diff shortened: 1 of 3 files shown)


breed808 (approved):
A good solution to the timezone issue, thanks for this one :+1:

PR #1500, opened 2024-05-18
printer collector: Use ENUM pattern for printer status

jkroepke:
Something that I missed on my review. 

For consistency reasons, we should use the same metric pattern was we are using for services.

Tested locally


jkroepke (commented):
LGTM 👍

PR #1498, opened 2024-05-17
logical disk: Fix metrics for non drive letter disks

jkroepke:
Fixes #1493
Fixes #1491

I introduce a dedicated `windows_logical_disk_info` metric here is avoid duplicate label info

```
# HELP windows_logical_disk_idle_seconds_total Seconds that the disk was idle (LogicalDisk.PercentIdleTime)
# TYPE windows_logical_disk_idle_seconds_total counter
windows_logical_disk_idle_seconds_total{volume="C:"} 3958.073043
windows_logical_disk_idle_seconds_total{volume="D:"} 3193.0903421
windows_logical_disk_idle_seconds_total{volume="D:\\Test"} 153.78342139999998
windows_logical_disk_idle_seconds_total{volume="HarddiskVolume2"} 2885.0179252
windows_logical_disk_idle_seconds_total{volume="HarddiskVolume3"} 2934.6000697
# HELP windows_logical_disk_info A metric with a constant '1' value labeled with logical disk information
# TYPE windows_logical_disk_info gauge
windows_logical_disk_info{disk="-1",filesystem="",partition="-1",serial_number="",volume="D:\\Test",volume_name=""} 1
windows_logical_disk_info{disk="-1",filesystem="",partition="-1",serial_number="",volume="HarddiskVolume2",volume_name=""} 1
windows_logical_disk_info{disk="-1",filesystem="",partition="-1",serial_number="",volume="HarddiskVolume3",volume_name=""} 1
windows_logical_disk_info{disk="0",filesystem="NTFS",partition="2",serial_number="668EEC37",volume="C:",volume_name="Windows"} 1
windows_logical_disk_info{disk="1",filesystem="NTFS",partition="0",serial_number="50AE953B",volume="D:",volume_name="Temporary Storage"} 1
```

PR #1511, opened 2024-06-03
Fix omission of RemoteFX Graphics metrics

benash:


jkroepke:
Thanks you, cloud you please include the change from #1508 ?

jkroepke (approved):

jkroepke:
@benash Could please add an empty commit here to trigger the DCO app again?

benash:
It seems this also might need #1498 to avoid timing out.

PR #1533, opened 2024-07-09
fix(README.md): Corrected image links

FRFlo:
The ghcr.io and the docker.io image are inverted with their description


jkroepke:
thanks!

jkroepke (approved):

PR #1637, opened 2024-09-23
cpu: add workaround for counter resets related to `% Processor Utility` metric

### LINKED ISSUE #1299: Possible aberation with the new CPU metrics on rtc counter reset (usage % above 100%)
windows_exporter version: **0.23.0**

I am trying to migrate our CPU usage percent metric to the new ones introduce in [v0.21.0](https://github.com/prometheus-community/windows_exporter/releases/tag/v0.21.0)

Query 1 (**Old**) per core usage (limited to 1 core for the example):
`((1-(avg without (mode) (rate(windows_cpu_time_total{mode="idle", core="0,0"}[1m])))) * 100)`

Query 2 (**New**) per core usage (limited to 1 core for the example):
`rate(windows_cpu_processor_utility_total{core="0,0"}[1m]) / rate(windows_cpu_processor_rtc_total{}[1m])`

The issue is that on sometimes I can get values way above 100% (120%+) with the new query. I fixed it with a `clamp`, but it's far from ideal. (I actually just reverted the cahnge in production until I have a proper way to fix the issue)

Example with rate Query1 in yellow and Query2 in green
image

Looking at the detail of  each new metrics I might have found the root cause:

- `windows_cpu_processor_utility_total` is fine (It continue to increment until the computer reboots)
- `windows_cpu_processor_rtc_total` on the other hand is reseted every time it reach 4300000000, and the calculation artefacts append exatly at that time.

Example bellow with `windows_cpu_processor_rtc_total` graphed for 2 different servers
image

@higels is this reset expected in between reboots? Am I doing something wrong?
From my understanding, with the lack of `xrate` we should avoid frequent counter resets

PS: Our Windows servers are mostly VMware VMs

jkroepke:
Fixes #1299 

## Problem

The `% Processor Utility` performance counter hold 2 values: Amount of busy time (V), amount of CPU tick (D).

To calculate the CPU percentage, the follow formula is needed: `(V1 - V0) / (D1 - D0)`. In prometheus, it can calculates via `rate(V[1m]) / rate(D[1m])`.

The CPU tick counter is an unsigned int32, with overflows approx. each 2 hours and the rollover happens on 4294967295 (2^32).

Program, that fetching the raw values, can handle the counter reset, because interger substractor works on overflows, too.  (go ref: [go.dev/play/p/57GXzfT-m5G](https://go.dev/play/p/57GXzfT-m5G))

For example:

C0 = ((2^32)-10)
C1 = 10
C1 - C0 = 20

The problem here is that all Prometheus values are float64. By converting the values, the bit based mechanic is lost.

This image explain how Prometheus handle the overflow which results into gaps using the rate function

image

## Solution

The solution will introduce a dedicated counter which do some interger calcuation on each scrape. and add the value to a float64 counter


pkg/collector/cpu/cpu.go
```diff
@@ -251,6 +262,30 @@ func (c *Collector) collectFull(ctx *types.ScrapeContext, logger *slog.Logger, c
...
+
+		if val, ok := c.processorMPerfValues[core]; ok {
+			c.processorMPerfValues[core] = cpuCounter{
+				uint32(cpu.ProcessorMPerf),
+				val.totalValue + float64(uint32(cpu.ProcessorMPerf)-val.lastValue),
+			}
+		} else {
+			c.processorMPerfValues[core] = cpuCounter{
+				uint32(cpu.ProcessorMPerf),
+				cpu.ProcessorMPerf,
```
jkroepke, line 285:
```suggestion
				0,
```

pkg/collector/cpu/cpu.go
```diff
@@ -251,6 +262,30 @@ func (c *Collector) collectFull(ctx *types.ScrapeContext, logger *slog.Logger, c
...
 
+		if val, ok := c.processorRTCValues[core]; ok {
+			c.processorRTCValues[core] = cpuCounter{
+				uint32(cpu.ProcessorRTC),
+				val.totalValue + float64(uint32(cpu.ProcessorRTC)-val.lastValue),
+			}
+		} else {
+			c.processorRTCValues[core] = cpuCounter{
+				uint32(cpu.ProcessorRTC),
+				cpu.ProcessorRTC,
```
jkroepke, line 273:
```suggestion
				0,
```

PR #1638, opened 2024-09-24
tcp: use GetExtendedTcpTable to display states Closed, Listening, SynSent, SynRcvd, CloseWait, TimeWait ...

astigmata:
Windows API Integration

The program interacts with the Windows API using syscalls to access low-level TCP information:

It loads the iphlpapi.dll library and uses the GetExtendedTcpTable function to retrieve TCP connection information.

The MIB_TCPROW_OWNER_PID struct is defined to represent a row in the TCP connection table.

```
# HELP windows_tcp_connections_state_count Number of TCP connections by state and address family
# TYPE windows_tcp_connections_state_count gauge
windows_tcp_connections_state_count{af="ipv4",state="CLOSE_WAIT"} 2
windows_tcp_connections_state_count{af="ipv4",state="ESTABLISHED"} 9
windows_tcp_connections_state_count{af="ipv4",state="LISTENING"} 14
windows_tcp_connections_state_count{af="ipv6",state="ESTABLISHED"} 4
windows_tcp_connections_state_count{af="ipv6",state="LISTENING"} 11
windows_tcp_connections_state_count{af="ipv6",state="TIME_WAIT"} 2
```


jkroepke:
Hi @astigmata 

great PR so far! I liked that WMI is not used. But it needs to be restructured a little bit. Do you have an example output of the metrics?

In another PR (#1635), I'm also start to use `iphlpapi.dll` to gain informations about routes.

Since the multiple collection may use `iphlpapi.dll`, the logic to Windows API should be separated to an dedicated packages, like I did it in https://github.com/prometheus-community/windows_exporter/pull/1635/files

pkg/collector/tcp/tcp.go
```diff
@@ -144,7 +193,72 @@ func (c *Collector) Build(_ *slog.Logger, _ *wmi.Client) error {
 		[]string{"af"},
 		nil,
 	)
-
+	c.connectionsCLOSED = prometheus.NewDesc(
```
jkroepke, line 196: Does it make sense to have an dedicated metric for each state?

What about having one metric (connection_state) and a label "state"

astigmata: Great idea

astigmata: state label added

pkg/collector/tcp/tcp.go
```diff
@@ -35,6 +37,7 @@ type Collector struct {
 	segmentsReceivedTotal      *prometheus.Desc
 	segmentsRetransmittedTotal *prometheus.Desc
 	segmentsSentTotal          *prometheus.Desc
+	TCPState                   *prometheus.Desc
```
jkroepke, line 40:
```suggestion
	tcpState                   *prometheus.Desc
```

alle files should be keep private

pkg/collector/tcp/tcp.go
```diff
@@ -144,7 +147,11 @@ func (c *Collector) Build(_ *slog.Logger, _ *wmi.Client) error {
 		[]string{"af"},
 		nil,
 	)
-
+	c.TCPState = prometheus.NewDesc(
```
jkroepke, line 150:
```suggestion
	c.tcpState = prometheus.NewDesc(
```

pkg/headers/iphlpapi/types.go
```diff
@@ -0,0 +1,11 @@
+package iphlpapi
+
+// MIB_TCPROW_OWNER_PID represents a row in the TCP connection table
```
jkroepke, line 3: I have the feeling that MIB_TCPROW_OWNER_PID just works, if af =IPv4.

If af is IPv6, the struct seems different. https://learn.microsoft.com/de-de/windows/win32/api/tcpmib/ns-tcpmib-mib_tcp6row_owner_pid

For example, dwLocalAddr is an [16]byte (128bit) value, which is 4 times larger than uint32. This could be the reasons, why IPv6 has some unknown values.

```
windows_tcp_connection_state_count{af="ipv6",state="UNKNOWN_0"} 16
```

pkg/headers/iphlpapi/iphlpapi.go
jkroepke, line 1: Taking a look at https://github.com/pytimer/win-netstat/blob/efa1aff6aafc7b37ca3ffb136b6b27a11178ffd6/tcp.go#L12  - it also separate between IPv4 and IPv6. 

For me, its also fine to just copy the code, since win-netstat has MIT license, too. A reference to the original code should be left.

jkroepke (changes requested):

pkg/collector/tcp/tcp.go
astigmata: ok replaced TCPState with connectionsStateCount

pkg/collector/tcp/tcp.go
astigmata: ok replaced TCPState with connectionsStateCount

pkg/headers/iphlpapi/iphlpapi.go
astigmata: ok thanks for the link

pkg/headers/iphlpapi/types.go
astigmata: now it works correctly

jkroepke:
@astigmata After a review, I can take the lint issues, thats fine!

astigmata:
> @astigmata After a review, I can take the lint issues, thats fine!

@jkroepke Thank you for the help, it's my first contrib

jkroepke:
Oh, you did it very well for the first contribution! :-) 

I will leave it open for now, because I would like to deliver some bug-fix only releases in context of 0.29.0, but this PR wil be included in the next 0.30.0 release!

Thank you!

jkroepke (approved):
LGTM 👍

PR #1679, opened 2024-10-10
*: [0.27] don't exclude resources, if exclude option is empty. 

jkroepke:
Ref: https://github.com/grafana/alloy/issues/1845#issuecomment-2402522821

### DIFF
--- a/pkg/types/regexp.go
+++ b/pkg/types/regexp.go
@@ -3,6 +3,6 @@ package types
 import "regexp"
 
 var (
-	RegExpAny   = regexp.MustCompile(".+")
-	RegExpEmpty = regexp.MustCompile("")
+	RegExpAny   = regexp.MustCompile("^.+$")
+	RegExpEmpty = regexp.MustCompile("^$")
 )


ptodev:
The regex for `include` should probably also be ignored if it's empty, since it'll match everything if it's empty.

pkg/types/regexp.go
jkroepke, line 1: @ptodev by changing the defaults here, the default values at alloy should not longer be "empty" by default which should be sufficient for that case? That would be the most clean approach.

jkroepke:
@ptodev Not sure who takes the responsibility here. If an empty pattern from type `*regex.Regex` is given as parameter which matches everything and it passed as exclude option, it works as expected. 

Instead check against an empty pattern, I would like to check against `nil`.

pkg/types/regexp.go
ptodev: Isn't it backwards incompatible though, since if you set an empty string previous versions of the exporter would treat this differently than new versions?

ptodev (commented):
> Instead check against an empty pattern, I would like to check against `nil`.

Do you mean changing the `string` config argument to `*string`? It could be a legitimate thing to do, but wouldn't it also break users who have already explicitly set it to `""`?

jkroepke:
> Do you mean changing the `string` config argument to `*string`?

At the moment, the config argument has the type `*regexp.Regexp`. windows_exporter expects a `*regexp.Regexp`, not an `string`.

pkg/types/regexp.go
jkroepke: Correct. Until reaching 1.0, we do some cleanups on the collector interfaces, including breaking changes.

We plan to reach 1.0 in mid 2025. Once reach 1.0 stable, we are not allowed such breaking changes.

The bug where I agree, is that the default value of exclude is excluding everything by default. Changing the default should solve that kind of issue.

From Alloy point of view, that should be fine as well. Alloy is not using a empty string by default, it's using the mention default values here.

https://github.com/grafana/alloy/blob/ee9034b5daaab953e2368fd5bf45f9ead47b3771/internal/component/prometheus/exporter/windows/config_windows.go#L44-L49

ptodev:
Ok, I see! Checking for nil does make sense.

PR #1682, opened 2024-10-10
chore(docs): fix link to update collector

ehrenfeu:
Probably the collector got renamed from `updates` to `update` at some point but the link in the README was forgotten.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Name     | Description | Enabled by default
 [textfile](docs/collector.textfile.md) | Read prometheus metrics from a text file |
 [thermalzone](docs/collector.thermalzone.md) | Thermal information |
 [time](docs/collector.time.md) | Windows Time Service |
-[updates](docs/collector.updates.md) | Windows Update Service |
+[update](docs/collector.update.md) | Windows Update Service |
 [vmware_blast](docs/collector.vmware_blast.md) | VMware Blast session metrics |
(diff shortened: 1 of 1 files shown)


jkroepke:
Thanks!

jkroepke (approved):

ehrenfeu:
Thanks for merging @jkroepke ! 💚 

Since I'm reading this this in the title now, I'd like to ask one thing that confuses me since a while when looking at e.g. the changelogs of a new release: what does `chore` mean? I was searching for a component of that name but couldn't find any...
Should it actually just read `core` instead, or is this a misinterpretation / lack of understanding on my side?

jkroepke:
Glad you ask, I havn't that style documented here.

The base for that kind of term are conventional commits. I used this for all changes not related specific code package.

long: https://www.conventionalcommits.org/en/v1.0.0/
short: https://gist.github.com/Zekfad/f51cb06ac76e2457f11c80ed705c95a3

I used the style for PR titles, because PR title are used for changelog generation. But if someone add an fix for the update collector, `update:` may used here.

In this case, `docs:` would be the correct prefix, but my [hand-written Ci check](https://github.com/prometheus-community/windows_exporter/blob/90ac0b269efda2c615961bd80ad5e12e0359f80c/.github/workflows/pr-check.yaml#L40) doesn't support `docs:` right now.

PR #1690, opened 2024-10-15
docs: add actual Windows Update service name

ehrenfeu:
To check if the _Windows Update_ service is running it is useful to run the `Get-Service` cmdlet, therefore it's good to know the actual service _name_ (as opposed to the _display name_ that is used e.g. by the Services snap-in).

For example

```
Get-Service wuauserv
```

will show the current status of the update service.

Additionally, this commit modifies several *s* to be lowercase, as the full *display name* of the corresponding service is "Windows Update", not "Windows Update Service" - at least on all the Windows versions I was able to check.

### DIFF
--- a/docs/collector.update.md
+++ b/docs/collector.update.md
@@ -1,19 +1,20 @@
 # update collector
 
-The update collector exposes the Windows Update Service metrics. Note that the Windows Update Service must be running, else metric collection will fail.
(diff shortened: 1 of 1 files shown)


jkroepke:
Thanks! Could you pleas sign off your commit?

ehrenfeu:
Done. Sorry. Apparently I need to check my git client configuration...

jkroepke (approved):
thx!

PR #1702, opened 2024-10-22
docs: Add log.file argument to README

PolygonError:
Adds a line which explains the log.file argument, and explains how the [MSI installer adds an argument which defaults this to eventlog](https://github.com/prometheus-community/windows_exporter/blob/ce2df385a6f994889b70d7ea169df7b04dc83320/installer/files.wxs#L16).

Description taken from log flag setup under [internal/log/flag/flag.go](https://github.com/prometheus-community/windows_exporter/blob/ce2df385a6f994889b70d7ea169df7b04dc83320/internal/log/flag/flag.go#L27)

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,17 +75,18 @@ This can be useful for having different Prometheus servers collect specific metr
 
 windows_exporter accepts flags to configure certain behaviours. The ones configuring the global behaviour of the exporter are listed below, while collector-specific ones are documented in the respective collector documentation above.
 
-| Flag                                 | Description                                                                                                                                         | Default value |
-|--------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------|---------------|
-| `--web.listen-address`               | host:port for exporter.                                                                                                                             | `:9182`       |
-| `--telemetry.path`                   | URL path for surfacing collected metrics.                                                                                                           | `/metrics`    |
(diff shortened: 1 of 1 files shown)


jkroepke:
Thanks a lot!

Background: MSI installer has set it to eventlog, since in a Windows Service context stderr is piped to null.

PolygonError:
> Thanks a lot!
> 
> Background: MSI installer has set it to eventlog, since in a Windows Service context stderr is piped to null.

Thanks for the info, I never realised until I went to look for an issue and found a bunch of events from it which I don't care about, so I just changed it to log to a text file in the installation directory.

Zeeeeta:
> > Thanks a lot!
> > Background: MSI installer has set it to eventlog, since in a Windows Service context stderr is piped to null.
> 
> Thanks for the info, I never realised until I went to look for an issue and found a bunch of events from it which I don't care about, so I just changed it to log to a text file in the installation directory.

May I know how do you set it to a text file?

I use the installer MSI change the setting to 
--log.file "C:\Program Files\windows_exporter\log.txt"
but it stuck there and cant start the service. (everyone has full control on folder, stuck no matter log.txt exists or not)
image

Thanks.

jkroepke:
please create a new issue for that! thank you.

PR #1722, opened 2024-11-08
docs: Update textfile collector defaults inclusion

yangskyboxlabs:
`textfile` collector was disabled by default in #1560 but the collector documentation was not updated to match.


jkroepke:
Thank you!

jkroepke (approved):

PR #1986, opened 2025-04-09
update: fix naming flags (renamed `--collector.updates.*` to `--collector.update.*`)  (click PR number for more information)

neonwhiskers:
# Breaking

| Old | New |
|---|---|
| `--collector.updates.online` | `--collector.update.online` |
| `--collector.updates.scrape-interval` | `--collector.update.scrape-interval` |

## What this PR does

This PR adds support for both the deprecated `--collector.updates.*` flags and the new `--collector.update.*` flags in the update collector.

It ensures:
- Users can still use the old flags without breaking existing setups.
- When both old and new flags are provided, the new ones take precedence.
- A clear warning is printed if deprecated flags are used, guiding users to update.

## Why this is needed

Recent updates standardized the metric prefix to `update`, but the collector flags remained `updates`. To support a cleaner interface while preserving compatibility, this PR introduces new flags and a transition path.

### DIFF
--- a/internal/collector/update/update.go
+++ b/internal/collector/update/update.go
@@ -84,15 +84,57 @@ func NewWithFlags(app *kingpin.Application) *Collector {
 		config: ConfigDefaults,
 	}
 
+	var (
+		online         bool
+		scrapeInterval time.Duration
+	)
+
 	app.Flag(
 		"collector.updates.online",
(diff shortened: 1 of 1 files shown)


jkroepke (approved):
LGTM 👍

PR #2047, opened 2025-05-20
Increase time resolution when possible (collectors: os, system, time)

### LINKED ISSUE #2044: Higher resolution for system time
### Problem Statement

It would be nice if `windows_os_time` had more accuracy, just like node_exporter `node_time_seconds` has.

### Proposed Solution

I know very little of Go, but AFAICT this patch should be enough and would be fully retro-compatible: ([os.go:L292](https://github.com/prometheus-community/windows_exporter/blob/v0.31.0-beta.1/internal/collector/os/os.go#L292C1-L292C3))
```patch
-		float64(time.Now().Unix()),
+		float64(time.Now().UnixMicro()) / 1e6),
```
PS: I noticed it is being deprecated for `windows_time_current_timestamp_seconds` but the `time` module is not enabled in the default build, so ideally it would be nice to have improved resolution even in the "old" method.

lapo-luchini:

Prometheus expects time values in seconds, but since every value is a `float64` is possible (and usually done) to include sub-second precision in values. This should be retro-compatible and break no probe on the user side that I can think of.

#2044 

Time values will be more precise. A user assuming less precision could have unexpected results.
(but they should be mostly positive, I think?)

#### Checklist


lapo-luchini:
This change could also make sense in other collectors, such as `filetime` or `logon`.

jkroepke:
@lapo-luchini should I wait for more commits here?

lapo-luchini:
Oh yes, sorry, I'm doing it shortly.

lapo-luchini:
Okay, I converted all the remaining `.Unix()` to `.UnixMicro()/1e6`.
I wasn't able to test these individually, but the change is straightforward enough.

jkroepke:
Thanks!

PR #2084, opened 2025-06-17
logical_disk: skip unmounted volumes

nicjansma:

The `logical_disk` collector may fail if a volume is recognized by Windows but not mounted.  All `logical_disk` metrics will then not be emitted.

In this case, the error logged is something like:

```
source=collect.go:212 msg="collector logical_disk failed after 42.6338ms, resulting in 0 metrics"
  err="failed to get volumes: GetVolumePathNamesForVolumeName: The system cannot find the file specified."
```

This change makes it so the failure from `GetVolumePathNamesForVolumeName` of `ERROR_FILE_NOT_FOUND` is not a hard-stop (will just be skipped), and the gathering of metrics for other volumes will continue.

This scenario can happen when e.g. an uninitialized volume is plugged into the computer.

image

n/a

n/a

n/a

#### Checklist

### DIFF
--- a/internal/collector/logical_disk/logical_disk.go
+++ b/internal/collector/logical_disk/logical_disk.go
@@ -701,6 +701,11 @@ func getAllMountedVolumes() (map[string]string, error) {
 				break
 			}
 
+			if errors.Is(err, windows.ERROR_FILE_NOT_FOUND) {
+				// the volume is not mounted
+				break
+			}
+
 			if errors.Is(err, windows.ERROR_NO_MORE_FILES) {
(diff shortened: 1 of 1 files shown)


internal/collector/logical_disk/logical_disk.go
```diff
@@ -701,6 +701,11 @@ func getAllMountedVolumes() (map[string]string, error) {
 				break
 			}
 
+			if errors.Is(err, windows.ERROR_FILE_NOT_FOUND) {
```
jkroepke, line 704: can we merge the both ifs together?

jkroepke (approved):
Thanks!

PR #2233, opened 2025-10-17
iis: metric collector fix for align metric types with Prometheus conventions Correct 9 metric type misclassifications in IIS collectors: - Changed 8 metrics from Counter to Gauge (timestamps, maximums, active counts, rates) - Changed 1 metric from Gauge to Counter (cumulative rejection total)

VidhyaSanjeevi:
…tric type misclassifications in IIS collectors: - Changed 8 metrics from Counter to Gauge (timestamps, maximums, active counts, rates) - Changed 1 metric from Gauge to Counter (cumulative rejection total)

### DIFF
--- a/internal/collector/iis/iis_app_pool_was.go
+++ b/internal/collector/iis/iis_app_pool_was.go
@@ -168,6 +168,10 @@ func (c *Collector) buildAppPoolWAS() error {
 }
 
 func (c *Collector) collectAppPoolWAS(ch chan<- prometheus.Metric) error {
+	if c.perfDataCollectorAppPoolWAS == nil {
+		return nil
+	}
+
 	err := c.perfDataCollectorAppPoolWAS.Collect(&c.perfDataObjectAppPoolWAS)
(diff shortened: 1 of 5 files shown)


internal/collector/iis/iis_app_pool_was.go
```diff
@@ -168,6 +168,10 @@ func (c *Collector) buildAppPoolWAS() error {
 }
 
 func (c *Collector) collectAppPoolWAS(ch chan<- prometheus.Metric) error {
+	if c.perfDataCollectorAppPoolWAS == nil {
```
jkroepke: Whats the reason for this?

jkroepke: There is already a null check on function side, with is sufficient. It may not make sense to have the nil check twice

https://github.com/prometheus-community/windows_exporter/blob/370a42b79ad518695a7391b94cc49c12b13dc63f/internal/pdh/collector.go#L279-L281

PR #2288, opened 2026-01-02
hyperv: Add host CPU ratio metric for monitoring CPU allocation

Dominik-esb:

*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*:

N/A - This is a feature enhancement

- The metric uses existing performance counters (`Hyper-V Hypervisor Logical Processor` and `Hyper-V Hypervisor Virtual Processor`) to calculate the ratio
- The host subcollector is enabled by default as part of the standard hyperv collector configuration
- Includes proper hostname labeling for multi-host monitoring scenarios

- New metric: `windows_hyperv_host_cpu_ratio` with `host` label
- New subcollector: `host` (enabled by default in hyperv collector)
- Metric is automatically collected when the hyperv collector is enabled

#### Checklist

### DIFF
--- a/internal/collector/hyperv/hyperv_host.go
+++ b/internal/collector/hyperv/hyperv_host.go
@@ -44,6 +44,8 @@ type collectorHost struct {
 
 type perfDataCounterValuesHost struct {
 	Name string
+
+	HypervisorLogicalProcessorTotalRunTimePercent float64 `perfdata:"% Total Run Time"`
 }
 
 type vmTopology struct {


Dominik-esb:
Hi @jkroepke,

could you have a look at this idea?

I’m trying to calculate the ratio between logical CPU cores in a cluster and the total v-cores assigned to VMs (e.g. 100 physical cores / 300 v-cores = 3:1).

```
round(
  count(count by (vm, core) (windows_hyperv_hypervisor_virtual_processor_time_total{cluster=~"xxx"}))
  / 
  (sum(windows_cpu_logical_processor{cluster=~"xxx"}) / 2)
  * 10
) / 10
```

This isn’t straightforward, as there is no per-VM metric that exposes the configured core count. Currently, I’m approximating the number of virtual cores using:

```
count(
  count by (vm, core) (
    windows_hyperv_hypervisor_virtual_processor_time_total{cluster=~"xxx"}
  )
)
```


This works to some extent, but it doesn’t feel ideal.

While testing, I also noticed that some VMs are missing, causing the v-core count per cluster to differ from what Hyper-V reports. In multiple clusters, 2–6 VMs are missing. This is likely a separate issue, but it directly affects the CPU ratio calculation.

Even if the VM counting issue can be resolved, having a reliable logical-core-to-v-core ratio per host could be useful.

As an alternative it might also be worth considering exposing the configured core count per VM as a dedicated metric

jkroepke:
> This isn’t straightforward, as there is no per-VM metric that exposes the configured core count. 

Sometimes, I run additional Win32 APIs calls to gain such informations. Generally, I like to avoid WMI as much as possible. As I know, the [Host Compute Service API](https://learn.microsoft.com/en-us/virtualization/api/#host-compute-system-apis) can be used to query such information. The Host Compute Service API is used by the container collector.

PR #2296, opened 2026-01-12
mscluster: Add virtual disk metrics sub-collector

Dominik-esb:
### What this PR does / why we need it

This PR adds a new virtual disk sub-collector to the mscluster collector to expose storage metrics for Windows Failover Cluster virtual disks (Storage Spaces).

**New metrics:**
- `windows_mscluster_virtualdisk_info` - Virtual disk information with friendly name label
- `windows_mscluster_virtualdisk_size_bytes` - Total size of the virtual disk in bytes
- `windows_mscluster_virtualdisk_footprint_on_pool_bytes` - Physical storage consumed on the storage pool in bytes
- `windows_mscluster_virtualdisk_storage_efficiency_percent` - Storage efficiency percentage (Size/FootprintOnPool * 100)

The metrics query the `MSFT_VirtualDisk` WMI class from the `root/Microsoft/Windows/Storage` namespace to provide visibility into virtual disk storage usage and efficiency in clustered environments.

### Which issue this PR fixes

*(optional)*: fixes #

### Special notes for your reviewer

**Testing Notes:**
- Requires Windows Server with Failover Clustering and Storage Spaces configured
- The collector only runs when explicitly enabled via `--collector.mscluster.enabled=virtualdisk`

### Particularly user-facing changes

- New sub-collector `virtualdisk` available under the mscluster collector
- Provides storage efficiency metrics for virtual disks in Windows Failover Clusters
- Metrics help identify thin provisioning effectiveness and storage pool capacity planning

### Checklist

- [ ] Added documentation for the new sub-collector
- [x] Tested on Windows Server with Failover Clustering enabled

### DIFF
--- a/internal/collector/mscluster/mscluster.go
+++ b/internal/collector/mscluster/mscluster.go
@@ -161,7 +161,7 @@ func (c *Collector) Build(_ *slog.Logger, miSession *mi.Session) error {
 			errs = append(errs, fmt.Errorf("failed to build resource group collector: %w", err))
 		}
 	}
-  
+
 	if slices.Contains(c.config.CollectorsEnabled, subCollectorSharedVolumes) {
 		if err := c.buildSharedVolumes(); err != nil {
 			errs = append(errs, fmt.Errorf("failed to build shared_volumes collector: %w", err))
@@ -246,7 +246,7 @@ func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
 
 	go func() {
 		defer wg.Done()
-    
+
 		if slices.Contains(c.config.CollectorsEnabled, subCollectorSharedVolumes) {
 			if err := c.collectSharedVolumes(ch); err != nil {
 				errCh <- fmt.Errorf("failed to collect shared_volumes metrics: %w", err)


Dominik-esb:
Hi @jkroepke ,

I'm implementing virtual disk health metrics for the mscluster collector and ran into a strange issue where the exporter hangs during initialization. Would appreciate your insight on this.

### The Problem

When I add `OperationalStatus` and `HealthStatus` fields to the struct, the exporter hangs completely on startup:

```go
type msftVirtualDisk struct {
    FriendlyName      string `mi:"FriendlyName"`
    OperationalStatus string `mi:"OperationalStatus"`  // <- Adding this causes hang
    HealthStatus      string `mi:"HealthStatus"`       // <- Adding this causes hang
    Size              uint64 `mi:"Size"`
    FootprintOnPool   uint64 `mi:"FootprintOnPool"`
}
```

```
.\windows_exporter.exe --log.level=debug --collectors.enabled="mscluster"
time=2026-01-12T00:48:18.967+01:00 level=DEBUG source=main.go:150 msg="logging has Started"
```

### What's Confusing

The fields definitely exist and work fine in PowerShell:

```powershell
Get-CimInstance -Namespace "root\Microsoft\Windows\Storage" -ClassName MSFT_VirtualDisk |
    Select-Object FriendlyName, OperationalStatus, HealthStatus, Size, FootprintOnPool | Format-List

# Output:
FriendlyName      : xxx-FTT2-CSV04
OperationalStatus : OK
HealthStatus      : Healthy
Size              : 16716012716032
FootprintOnPool   : 35827275005952
```

### What I Found

After digging through the code, I discovered that `QueryUnmarshal` in `internal/mi/session.go:258-277` iterates through every struct field with an `mi:` tag and calls `GetElement()` individually:

```go
for i := range elemType.NumField() {
    field := elemValue.Field(i)

    // Check if the field has an `mi` tag
    miTag := elemType.Field(i).Tag.Get("mi")
    if miTag == "" {
        continue
    }

    // This syscall hangs for OperationalStatus/HealthStatus on Storage namespace
    element, err := instance.GetElement(miTag)
    if err != nil {
        if errors.Is(err, MI_RESULT_NO_SUCH_PROPERTY) {
            continue  // Should skip missing properties
        }
        return fmt.Errorf("failed to get element %s: %w", miTag, err)
    }
    // ... unmarshal the value into the field
}
```

**I think the hang happens when `instance.GetElement("OperationalStatus")` or `instance.GetElement("HealthStatus")` is called.** The syscall to `MI_Instance_GetElement` never returns - no error, no timeout, just blocks forever at the kernel/WMI provider level? Since the hang is inside the syscall, it could be the reason no error caught.

### Why Does PowerShell Work Then?

- **PowerShell**: Uses `Get-CimInstance` which retrieves all properties at once in a single operation
- **Go MI Framework**: Retrieves properties individually via separate `GetElement()` syscalls for each tagged field

### What I've Triedon query only - hang still occurs during collection phase
- ✗ Removing fields from SELECT query but keeping in struct → still hangs (GetElement gets called anyway)
- ✗ I tried to add more debugging but my go knowledge is limited. 
- ✗ Removing validation query only → hangs during collection phase too
- ✓ Removing fields from struct entirely → works!


This seems like a Windows Storage Management API bug specific to MI's property-by-property retrieval, but I wanted to check if there's something I'm missing before giving up on these metrics.
Could you have a look at it?


Thanks for any guidance!

jkroepke:
I have the feeling that the type of both fields is is correct.

Ref: 

https://github.com/microsoft/wmi/blob/291d2706b618cbf9ff7d1c3b97195bc1510c23b0/server23h2/root/microsoft/windows/storage/MSFT_VirtualDisk.go#L47

https://github.com/microsoft/wmi/blob/291d2706b618cbf9ff7d1c3b97195bc1510c23b0/server23h2/root/microsoft/windows/storage/MSFT_VirtualDisk.go#L101

Dominik-esb:
Thanks, good to know — I honestly never came across this documentation before.

I tried to find the unit / underlying type for these fields and didn’t get very far, so this is really interesting.
So if I understand this correctly, that the CIM/WMI provider itself doesn’t return the raw value for these properties when accessed this way?

In other words, Get-CimInstance already receives a mapped / interpreted value (e.g. Healthy, OK) instead of the underlying numeric uint, which makes it look like the property is a string even though it isn’t.

One thing I’m still unclear about:
Is there a reliable way in PowerShell to inspect the actual underlying type of these CIM properties before running into issues like this?
For example, something that clearly shows whether a property is an enum / integer vs. a real string, instead of only showing the human-readable representation.

And related to that:
Is there any missing error handling or safeguard on "our" side that could catch this earlier? Even if this only affects development-time changes, a hard hang isn’t ideal — it would be much nicer to fail fast or surface a clearer error instead of blocking indefinitely.

I’ll try switching to the underlying numeric type to see if that avoids the hang.

Thanks again — this was a very helpful hint!

jkroepke:
I have really no idea about WMI and thats why I like to avoid it, if possible. But I'm aware that there is no other MS Cluster API. (?)

According [docs](https://learn.microsoft.com/en-us/windows/win32/api/mi/nf-mi-mi_instance_getelement), the type of the property is exposed as return field.

It could be possible that there is an error on the implementation inside windows_exporter.

Dominik-esb:
I changed the units and tested the exporter again. Sadly that did not fixed the problem.
Could you help out with some debug outputs?
I'm not sure where exactly and how to add them.

It should be around the GetInstance if I'm right? https://github.com/prometheus-community/windows_exporter/blob/master/internal/mi/session.go#L240

Dominik-esb:
You’re right.
I tested this earlier today with both a slice of uint16 and uint16. Neither is compatible with the library, since the API returns a slice. That’s fine for now.

OperationalStatus is only a nice-to-have metric, so I’m happy to drop it.
Thank for the hint. I had no clue what types the lib works with.

Dominik-esb:
Implementation done.
Documentation is updated, Metrics are tested. 

Thanks for the support

jkroepke:
It seems like the Query function only supports a subset of types

https://github.com/prometheus-community/windows_exporter/blob/f0d5fd9ba0a29e8a42e4c175f31d52e4fd807d02/internal/mi/session.go#L276-L297

compared to

https://github.com/prometheus-community/windows_exporter/blob/f0d5fd9ba0a29e8a42e4c175f31d52e4fd807d02/internal/mi/value.go#L73-L126

kind in mind, the mi library is written by myself - without AI. So only a minimal value product. It supports what windows_exporter needs and it not a general purpose library.

Dominik-esb:
Yes,

I remember I dig though your `mi` implementation and also saw the supported types but at this moment I believed the   `OperationalStatus` and `HealthStatus` are strings.

jkroepke:
> OperationalStatus is only a nice-to-have metric, so I’m happy to drop it.

Could you try `[]string` as type?

Dominik-esb:
Exporter is able to start with slice of strings but the query times out with OperationalStatus:
`time=2026-01-15T20:11:35.901+01:00 level=WARN source=collect.go:199 msg="collector mscluster timeouted after 9.5s, resulting in 1004 metrics"`

test-csv.ps1
jkroepke: remove this please

internal/collector/mscluster/mscluster_virtualdisk.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create WMI query: %w", err)
+	}
+
+	c.virtualDiskMIQuery = virtualDiskMIQuery
+
+	c.virtualDiskInfo = prometheus.NewDesc(
+		prometheus.BuildFQName(types.Namespace, nameVirtualDisk, "info"),
+		"Virtual Disk information (value is always 1)",
+		[]string{"name"},
```
jkroepke: Not so happy with `name` as stable identifier. since to virtual disks can have the same name?

internal/collector/mscluster/mscluster_virtualdisk.go
```diff
@@ -0,0 +1,145 @@
...
+
+	virtualDiskInfo              *prometheus.Desc
+	virtualDiskHealthStatus      *prometheus.Desc
+	virtualDiskSize              *prometheus.Desc
+	virtualDiskFootprintOnPool   *prometheus.Desc
+	virtualDiskStorageEfficiency *prometheus.Desc
+}
+
+// msftVirtualDisk represents the MSFT_VirtualDisk WMI class
+type msftVirtualDisk struct {
```
jkroepke: Could you add there the OperationalStatus with the correct type. but commit it out? and link to your comment? That should preserver some historical value.

Dominik-esb:
Example:
```
# HELP windows_mscluster_virtualdisk_footprint_on_pool_bytes Physical storage consumed by the virtual disk on the storage pool in bytes
# TYPE windows_mscluster_virtualdisk_footprint_on_pool_bytes gauge
windows_mscluster_virtualdisk_footprint_on_pool_bytes{name="ClusterPerformanceHistory",unique_id="2B2704D5E42BF148BB190ADDF82F3010"} 7.8920024064e+10
```

Dominik-esb:
Hi @jkroepke,

Same here: anything missing for the merge?

jkroepke:
@Dominik-esb please check if I resolve the Merge conflicts good

Dominik-esb:
All done 
Thanks you

PR #2300, opened 2026-01-20
horizon_blast: Add Omnissa Horizon Blast collector

tpowelldev:
## Summary
  - Add new collector for Omnissa Horizon Blast protocol metrics.
  - Exposes 114 metrics across 16 performance counter sets.
  - Includes session, imaging, audio, CDR, clipboard, printing, USB, and other Blast features.
  - Not enabled by default.
  - Tested on Windows 11 with Omnissa Agent 8.15.0
  - Test included.
  - All documentation updated.
 
Available MI counter sets included:
  - Horizon Blast Audio Counters
  - Horizon Blast CDR Counters
  - Horizon Blast Clipboard Counters
  - Horizon Blast HTML5 MMR Counters
  - Horizon Blast Imaging Counters
  - Horizon Blast Other Feature Counters
  - Horizon Blast Printing Counters
  - Horizon Blast RdeServer Counters
  - Horizon Blast RTAV Counters
  - Horizon Blast SDR Counters
  - Horizon Blast Serial Port and Scanner Counters
  - Horizon Blast Session Counters
  - Horizon Blast Smart Card Counters
  - Horizon Blast USB Counters
  - Horizon Blast View Scanner Counters
  - Horizon Blast Windows Media MMR Counters

This replaces the functionality that was removed in https://github.com/prometheus-community/windows_exporter/pull/1686 (vmware_blast collector) and updates the metric names / paths following the sale of Horizon from VMware to Omnissa

#### Checklist


jkroepke:
Hi, I guess the same could be archived by using the generic [performancecounter](https://github.com/prometheus-community/windows_exporter/blob/master/docs/collector.performancecounter.md) collector? I see no use-case for having a dedicated collector.

tpowelldev:
> Hi, I guess the same could be archived by using the generic [performancecounter](https://github.com/prometheus-community/windows_exporter/blob/master/docs/collector.performancecounter.md) collector? I see no use-case for having a dedicated collector.

Hey - thanks for replying. Whilst I agree that it's technically possible to achieve the same with performancecounter, that requires any users of Omnissa Horizon to write their own 200+ lines of yaml to configure everything they need to monitor their environment. Within that, they then need to make sure that they're configuring their type assignments correctly (counter vs gauge) and are converting units where sensible. 

With this collector, Horizon users can easily discover and monitor their platform metrics by enabling a single collector. Additionally, Grafana dashboards can be written and distributed to use fixed metric names for this.

This is exactly the same pattern as many of the other collectors, vmware, hyperv, cpu, memory, net, etc, so isn't breaking consistency, or hard to maintain (we can even ask Omnissa if they're happy to own this). What are the concerns about adding it? Is the project direction to eventually sunset dedicated collectors in favour of performancecounter? It would be good to understand for future contributions.

jkroepke:
Maintainer bandwidth is limited and test integration is missing. The collector is hard to maintain.

Why? If someone opens a bug, I have no real chance to validate or debug it. I need to craft special debug builds and ask the reporter for feedback. That is hard and slow.

If I need to refactor the exporter, these collectors are always a blind spot. For example, I had serious issues with the VMware collector. The migration from WMI queries to native performance queries was a headache. Documentation about the Performance Counters was basically non existent. After the release, there were still multiple bugs in that collector, which I had to fix without having a system to verify the issues.

My current test setup is a Windows Server 2025 virtual machine running on my gaming box.

I am more comfortable with something like presets, where the community can create YAML configurations. These could be embedded by default into windows_exporter.

In the past, I also rejected additional collectors like the defender collector, because performance counters also fit that situation.

I would like to reduce the number of collectors. The VMware collector is also a candidate. It could be migrated to the performance collectors.

The difference compared to cpu, memory, and net collectors is that they require additional Win32 API syscalls, which cannot be solved via performance collectors.

tpowelldev:
Thanks for the explanation — I do appreciate the maintenance concerns, not least from having looked at the commit history on this repo. I can see you’re managing it largely by yourself. That said, I work for an organisation with 1,000+ Horizon machines, and my employer is explicitly supportive of me maintaining this collector as part of my role. I’m happy to be listed as the owner/maintainer so this doesn’t become an ongoing blind spot for you.

I also hear your preference for presets, and that may well be the right long‑term direction. If the project intent is to eventually sunset dedicated collectors entirely, it would be helpful to state that clearly so contributors can align with it early. Having a framework to support a collection of performancecounter presets could also be useful, although that would still imply maintenance and testing being owned by the project.

If the PR still isn’t something you’re comfortable merging, I understand — but in that case we’ll likely need to maintain a fork of the core project with our own collectors injected to meet our internal needs. It’s not ideal, but it’s the practical fallback.

(For what it’s worth, a Defender collector would actually have been super useful for us too — but we can develop our own easily enough.)

Anyway, happy to discuss collector ownership / support, or equally happy to close this off and go our own way.

Thanks again for taking the time to reply.

jkroepke:
As I mention, currently no plans to merge this. It make more sense to reduce the amount of collectors and keep an focus on the generic one.

PR #2371, opened 2026-03-26
file: fix collector producing no metrics, startup hang, and duplicate metric errors

### LINKED ISSUE #2370: File collector (--collector.file) produces no metrics and hangs on startup
### Current Behavior

The `file` collector produces no metrics and causes the exporter to hang during startup when configured via a YAML config file or with multiple patterns.

Three separate bugs are responsible:

**1. YAML config file: patterns ignored / parse error**

When configuring `file-patterns` in `config.yaml` as a scalar string:

```yaml
collector:
  file:
    file-patterns: C:\\Windows\\System32\\spool\\PRINTERS\\*.spl
```

The exporter fails to start with:

```
Failed to load configuration err="configuration file validation error:
  yaml: unmarshal errors:
    line X: cannot unmarshal !!str into []string"
```

Even when written correctly as a YAML list, all patterns were silently merged into a single comma-joined string and treated as one invalid pattern, so `windows_file_count` always reported `0`.

**2. Exporter hangs on startup with any recursive pattern**

When `--collector.file.file-patterns` is set to a recursive pattern such as `C:\Users\**`, the exporter appears to hang and never finishes starting up. No metrics endpoint becomes available.

**3. Duplicate metric error when patterns overlap**

When using overlapping patterns (e.g. `*.spl` and `*` targeting the same directory), the scrape endpoint returns an HTTP 500 with:

```
error gathering metrics: collected metric "windows_file_mtime_timestamp_seconds"
  { label:{name:"file" value:"C:\\...\\1.spl"} } was collected before
  with the same name and label values
```

Prometheus then drops the entire scrape.

## Expected Behavior

1. YAML list patterns should be parsed correctly and each pattern should be monitored independently, producing one `windows_file_count{pattern="..."}` metric per entry.
2. The exporter should start instantly regardless of pattern scope — pattern validation at startup should not walk the filesystem.
3. Overlapping patterns should not cause duplicate metric errors — per-file metrics (`mtime`, `size`) should be deduplicated while per-pattern counts remain independent.

### Steps To Reproduce

```Markdown
**Bug 1 — YAML parse error:**
1. Create `config.yaml` with `file-patterns: C:\\Windows\\System32\\spool\\PRINTERS\\*.spl` (scalar, not list)
2. Run `.\windows_exporter.exe --config.file=config.yaml`
3. Exporter exits immediately with unmarshal error

**Bug 2 — Startup hang:**
1. Configure any recursive pattern: `--collector.file.file-patterns "C:\Users\**"`
2. Run the exporter
3. Exporter never finishes starting; no metrics endpoint

**Bug 3 — Duplicate metric error:**
1. Configure overlapping patterns:
   
   file-patterns:
     - "C:\\Windows\\System32\\spool\\PRINTERS\\*.spl"
     - "C:\\Windows\\System32\\spool\\PRINTERS\\*"
   
2. Start the exporter and scrape `/metrics`
3. HTTP 500 with duplicate label error
```

### Environment

- windows_exporter Version: latest (`7eb36c1`)
- Windows Server Version: Windows 11 / Windows Server 2019+

### windows_exporter logs

```shell
**Bug 1:**

time=... level=ERROR source=main.go msg="Failed to load configuration"
  err="configuration file validation error: yaml: unmarshal errors:
    line 7: cannot unmarshal !!str 'C:\Windows\...' into []string"


**Bug 2:**

time=... level=INFO source=main.go msg="Enabled collectors: file"
[no further output — exporter hangs]


**Bug 3:**

time=... level=ERROR source=http.go msg="error gathering metrics: 2 error(s) occurred:
* [from Gatherer #2] collected metric \"windows_file_mtime_timestamp_seconds\"
  { label:{name:\"file\" value:\"C:\\Windows\\System32\\spool\\PRINTERS\\1.spl\"}}
  was collected before with the same name and label values"
```

## Root Causes

| # | Location | Root Cause |
|---|----------|------------|
| 1a | `file.go` `NewWithFlags` | Used `StringsVar` instead of `StringVar` + `app.Action` split — the config resolver's comma-joined list was stored as a single element instead of being split |
| 1b | `file.go` `NewWithFlags` | `strings.Split("", ",")` returns `[""]` not `[]` when no patterns are set, leaking an empty pattern into every scrape |
| 1c | `pkg/collector/config.go` | `File` collector was missing from `ConfigDefaults`, so YAML config for the file collector was silently ignored |
| 2 | `file.go` `Build()` | Pattern validation called `doublestar.Glob()` — a full recursive filesystem walk — on every pattern at startup. With broad patterns this walks hundreds of thousands of files before the listener starts |
| 2b | `file.go` `Build()` | `filepath.ToSlash()` was not called before `doublestar.SplitPattern()`, so Windows backslash paths were never split correctly and `Glob` silently matched nothing |
| 3 | `file.go` `collectGlobFilePath` | Each pattern's goroutine independently emits `mtime`/`size` per file with no cross-pattern deduplication |

Dominik-esb:

Fixes multiple bugs in the `file` collector that together caused it to produce no metrics, hang on startup, and return HTTP 500 errors on overlapping patterns. Also adds a new `windows_file_count` metric.

---

**1. `pkg/collector/config.go` — `File` missing from `ConfigDefaults`**

The `File` entry was absent from the global `ConfigDefaults` struct, so YAML config for the file collector was silently ignored at startup.

```go
// before
var ConfigDefaults = Config{
    // ...
    Exchange: exchange.ConfigDefaults,
    // File was missing
    Fsrmquota: fsrmquota.ConfigDefaults,
}

// after
var ConfigDefaults = Config{
    // ...
    Exchange: exchange.ConfigDefaults,
    File:     file.ConfigDefaults,
    Fsrmquota: fsrmquota.ConfigDefaults,
}
```

---

**2. `NewWithFlags` — all patterns from config file merged into one invalid pattern**

`kingpin.StringsVar` was used for the `file-patterns` flag. The config resolver ([`internal/config/flatten.go`](../internal/config/flatten.go)) flattens YAML lists into a single comma-joined string and sets it as the flag default. `StringsVar` stores this verbatim as one element rather than splitting it, so all configured patterns were concatenated into a single invalid path and no files were ever matched.

```yaml
# config.yaml — three patterns
collector:
  file:
    file-patterns:
      - "C:\\Windows\\System32\\spool\\PRINTERS\\*.spl"
      - "C:\\Windows\\System32\\spool\\PRINTERS\\*.shd"
      - "C:\\Windows\\System32\\spool\\PRINTERS\\*"
```

```
# before — all three merged into one label, count always 0
windows_file_count{pattern="C:\...\*.spl,C:\...\*.shd,C:\...\*"} 0

# after — one series per pattern
windows_file_count{pattern="C:\Windows\System32\spool\PRINTERS\*.spl"} 2
windows_file_count{pattern="C:\Windows\System32\spool\PRINTERS\*.shd"} 2
windows_file_count{pattern="C:\Windows\System32\spool\PRINTERS\*"}     4
```

Fixed by switching to `StringVar` + `app.Action` split, consistent with the `net` and `logical_disk` collectors:

```go
// before
).Default(strings.Join(ConfigDefaults.FilePatterns, ",")).StringsVar(&c.config.FilePatterns)

// after
).Default(strings.Join(ConfigDefaults.FilePatterns, ",")).StringVar(&filePatterns)

app.Action(func(*kingpin.ParseContext) error {
    for _, p := range strings.Split(filePatterns, ",") {
        if p != "" {
            c.config.FilePatterns = append(c.config.FilePatterns, p)
        }
    }
    return nil
})
```

Empty strings are filtered to prevent a spurious `windows_file_count{pattern=""}` on every scrape when no patterns are configured (`strings.Split("", ",")` returns `[""]` not `[]`).

---

**3. `Build()` — exporter hangs on startup with recursive patterns**

`Build()` called `doublestar.Glob()` to validate each pattern at startup. This is a full recursive filesystem walk — with a broad pattern like `C:\Users\**` it walks hundreds of thousands of files before the HTTP listener starts, making the exporter appear permanently hung.

```go
// before — walks the entire filesystem subtree at startup
_, err := doublestar.Glob(os.DirFS(basePath), pattern, doublestar.WithFilesOnly())

// after — syntax check only, no filesystem access
if !doublestar.ValidatePattern(filepath.ToSlash(filePattern)) {
    return fmt.Errorf("invalid glob pattern: %s", filePattern)
}
```

`Build()` also lacked a `filepath.ToSlash()` call before `doublestar.SplitPattern()`. On Windows, backslash paths were never split correctly so the base path was always `"."` and the validation glob silently matched nothing. Fixed by adding `filepath.ToSlash()` — already present in `collectGlobFilePath` but missing here.

---

**4. `Collect()` — HTTP 500 on overlapping patterns**

When patterns overlap (e.g. `*.spl` and `*` targeting the same directory), the same file was walked by multiple goroutines and `mtime`/`size` metrics were emitted multiple times with identical labels. Prometheus rejects the entire scrape:

```
error gathering metrics: collected metric "windows_file_mtime_timestamp_seconds"
  { label:{name:"file" value:"C:\\Windows\\System32\\spool\\PRINTERS\\1.spl"}}
  was collected before with the same name and label values
```

Fixed by adding a mutex-protected `seenFiles` map shared across all pattern goroutines. Per-file metrics (`mtime`, `size`) are emitted exactly once regardless of how many patterns match a given file. Per-pattern `count` metrics remain independent.

---

**5. New metric — `windows_file_count`**

Added `windows_file_count{pattern="..."}` gauge reporting the number of files matched per pattern. Only emitted on a successful walk — a failed walk logs an error and skips the emit to avoid reporting a misleading partial count.

```
# HELP windows_file_count Number of files matching the pattern
# TYPE windows_file_count gauge
windows_file_count{pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*.spl"} 2
windows_file_count{pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*.shd"} 2
windows_file_count{pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*"}     4
```

- fixes #2370

- `windows_file_count` counts files for which metrics were actually emitted (unique per pattern, cross-pattern duplicates excluded). A file matched by both `*.spl` and `*` is counted once in the first pattern to emit it; the second pattern still walks it but skips the emit and does not increment its count.
- `doublestar.ValidatePattern` requires forward-slash paths; `filepath.ToSlash` is called before validation on all code paths.
- `seenFiles` is allocated fresh on every `Collect` call — no state leaks between scrapes.

- `windows_file_count{pattern="..."}` is a new metric — one series per configured pattern per scrape.
- The exporter no longer hangs on startup when broad recursive patterns such as `C:\Users\**` are configured.
- Overlapping patterns no longer cause HTTP 500 errors on `/metrics`.
- YAML `file-patterns` list is now correctly parsed; previously all patterns were silently ignored when using a config file.

#### Checklist

- [X] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [X] The PR title has a summary of the changes and the area they affect
- [X] The PR body has a summary to reflect any significant (and particularly user-facing) changes introduced by this PR

### DIFF
--- a/internal/collector/file/file.go
+++ b/internal/collector/file/file.go
@@ -205,7 +205,7 @@ func (c *Collector) collectGlobFilePath(ch chan<- prometheus.Metric, filePattern
 		mu.Lock()
 		_, alreadySeen := seenFiles[filePath]
 		if !alreadySeen {
-			seenFil‚es[filePath] = struct{}{}
+			seenFiles[filePath] = struct{}{}
 		}
 		mu.Unlock()
 
--- a/internal/collector/file/file.go
+++ b/internal/collector/file/file.go
@@ -51,6 +51,7 @@ type Collector struct {
 	logger    *slog.Logger
 	fileMTime *prometheus.Desc
 	fileSize  *prometheus.Desc
+	fileCount *prometheus.Desc
 }
 
 func New(config *Config) *Collector {
@@ -73,12 +74,23 @@ func NewWithFlags(app *kingpin.Application) *Collector {
 	c := &Collector{
 		config: ConfigDefaults,
 	}
-	c.config.FilePatterns = make([]string, 0)
+
+	var filePatterns string
 
 	app.Flag(
 		"collector.file.file-patterns",
 		"Comma-separated list of file patterns. Each pattern is a glob pattern that can contain `*`, `?`, and `**` (recursive). See https://github.com/bmatcuk/doublestar#patterns",
-	).Default(strings.Join(ConfigDefaults.FilePatterns, ",")).StringsVar(&c.config.FilePatterns)
+	).Default(strings.Join(ConfigDefaults.FilePatterns, ",")).StringVar(&filePatterns)
+
+	app.Action(func(*kingpin.ParseContext) error {
+		for _, p := range strings.Split(filePatterns, ",") {
+			if p != "" {
+				c.config.FilePatterns = append(c.config.FilePatterns, p)
+			}
+		}
+
+		return nil
+	})
 
 	return c
 }
@@ -110,12 +122,20 @@ func (c *Collector) Build(logger *slog.Logger, _ *mi.Session) error {
 		nil,
 	)
 
-	for _, filePattern := range c.config.FilePatterns {
-		basePath, pattern := doublestar.SplitPattern(filePattern)
+	c.fileCount = prometheus.NewDesc(
+		prometheus.BuildFQName(types.Namespace, Name, "count"),
+		"Number of files matching the pattern",
+		[]string{"pattern"},
+		nil,
+	)
 
-		_, err := doublestar.Glob(os.DirFS(basePath), pattern, doublestar.WithFilesOnly())
-		if err != nil {
-			return fmt.Errorf("invalid glob pattern: %w", err)
+	for _, filePattern := range c.config.FilePatterns {
+		if filePattern == "" {
+			continue
+		}
+
+		if !doublestar.ValidatePattern(filepath.ToSlash(filePattern)) {
+			return fmt.Errorf("invalid glob pattern: %s", filePattern)
 		}
 	}
 
@@ -127,18 +147,32 @@ func (c *Collector) Build(logger *slog.Logger, _ *mi.Session) error {
 func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
 	wg := sync.WaitGroup{}
 
+	var mu sync.Mutex
+
+	seenFiles := make(map[string]struct{})
+
 	for _, filePattern := range c.config.FilePatterns {
 		wg.Add(1)
 
 		go func(filePattern string) {
 			defer wg.Done()
 
-			if err := c.collectGlobFilePath(ch, filePattern); err != nil {
+			count, err := c.collectGlobFilePath(ch, filePattern, &mu, seenFiles)
+			if err != nil {
 				c.logger.Error("failed collecting metrics for filepath",
 					slog.String("filepath", filePattern),
 					slog.Any("err", err),
 				)
+
+				return
 			}
+
+			ch <- prometheus.MustNewConstMetric(
+				c.fileCount,
+				prometheus.GaugeValue,
+				float64(count),
+				filePattern,
+			)
 		}(filePattern)
 	}
 
@@ -147,9 +181,11 @@ func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
 	return nil
 }
 
-func (c *Collector) collectGlobFilePath(ch chan<- prometheus.Metric, filePattern string) error {
+func (c *Collector) collectGlobFilePath(ch chan<- prometheus.Metric, filePattern string, mu *sync.Mutex, seenFiles map[string]struct{}) (int, error) {
 	basePath, pattern := doublestar.SplitPattern(filepath.ToSlash(filePattern))
 	basePathFS := os.DirFS(basePath)
+
+	var count int
 
 	err := doublestar.GlobWalk(basePathFS, pattern, func(path string, d fs.DirEntry) error {
 		filePath := filepath.Join(basePath, path)
@@ -163,6 +199,19 @@ func (c *Collector) collectGlobFilePath(ch chan<- prometheus.Metric, filePattern
 
 			return nil
 		}
+
+		mu.Lock()
+		_, alreadySeen := seenFiles[filePath]
+		if !alreadySeen {
+			seenFiles[filePath] = struct{}{}
+		}
+		mu.Unlock()
+
+		if alreadySeen {
+			return nil
+		}
+
+		count++
 
 		ch <- prometheus.MustNewConstMetric(
 			c.fileMTime,
@@ -181,8 +230,8 @@ func (c *Collector) collectGlobFilePath(ch chan<- prometheus.Metric, filePattern
 		return nil
 	}, doublestar.WithFilesOnly(), doublestar.WithCaseInsensitive())
 	if err != nil {
-		return fmt.Errorf("failed to glob: %w", err)
+		return count, fmt.Errorf("failed to glob: %w", err)
 	}
 
-	return nil
+	return count, nil
 }
--- a/internal/collector/file/file.go
+++ b/internal/collector/file/file.go
@@ -83,7 +83,7 @@ func NewWithFlags(app *kingpin.Application) *Collector {
 	).Default(strings.Join(ConfigDefaults.FilePatterns, ",")).StringVar(&filePatterns)
 
 	app.Action(func(*kingpin.ParseContext) error {
-		for _, p := range strings.Split(filePatterns, ",") {
+		for p := range strings.SplitSeq(filePatterns, ",") {
 			if p != "" {
 				c.config.FilePatterns = append(c.config.FilePatterns, p)
 			}
@@ -201,6 +201,7 @@ func (c *Collector) collectGlobFilePath(ch chan<- prometheus.Metric, filePattern
 		}
 
 		mu.Lock()
+
 		_, alreadySeen := seenFiles[filePath]
 		if !alreadySeen {
 			seenFiles[filePath] = struct{}{}
--- a/internal/collector/file/file.go
+++ b/internal/collector/file/file.go
@@ -52,7 +52,6 @@ type Collector struct {
 	logger    *slog.Logger
 	fileMTime *prometheus.Desc
 	fileSize  *prometheus.Desc
-	fileCount *prometheus.Desc
 }
 
 func New(config *Config) *Collector {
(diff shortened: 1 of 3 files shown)


internal/collector/file/file.go
```diff
@@ -164,6 +200,20 @@ func (c *Collector) collectGlobFilePath(ch chan<- prometheus.Metric, filePattern
 			return nil
 		}
 
+		mu.Lock()
```
jkroepke: Do we need mutex here?

jkroepke (commented):
See comments

internal/collector/file/file.go
Dominik-esb: I think so,
I had a review session with claude code and it pointed out that the `seenFiles` map **could** be accessed from multiple goroutines. 
And as maps are not [Concurrency-Safe](https://medium.com/@diveshkumarchordia/go-maps-and-concurrency-why-you-need-synchronization-and-how-to-get-it-right-6cf5cae2f26b) I guess it could happen that two Goroutine write to the map and corruption happens

Dominik-esb: I think there is also the possibility to use `sync.Map` witch is also doing `mu.Lock()` under the hood

jkroepke: Why we have to use seenFiles across multiple glob pattern? If multiple glob pattern matches the same files, it should could twice.

Dominik-esb: It does. I added count++ before the alreadySeen check, so each pattern increments its own counter independently of
  every other pattern.

internal/collector/file/file.go
```diff
@@ -127,18 +147,32 @@ func (c *Collector) Build(logger *slog.Logger, _ *mi.Session) error {
 func (c *Collector) Collect(ch chan<- prometheus.Metric) error {
 	wg := sync.WaitGroup{}
 
+	var mu sync.Mutex
+
+	seenFiles := make(map[string]struct{})
```
jkroepke: After re-reading the code and your PR description, now I understand the purpose of `seenFiles` and I personally don't like it. 

What about add a second label, called pattern to make the metric unique, if the pattern overlap? Then, other opportunities like `sum(windows_file_size_bytes) by (pattern)` would work as expected.

Then, we would not longer need the new `windows_file_count` metric, since `count(windows_file_size_bytes) by (pattern)` would fit as well.

jkroepke (commented):
See comment

internal/collector/file/file.go
Dominik-esb: Done 

Example:

```
# HELP windows_file_mtime_timestamp_seconds File modification time
# TYPE windows_file_mtime_timestamp_seconds gauge
windows_file_mtime_timestamp_seconds{file="C:\\Windows\\System32\\spool\\PRINTERS\\1.shd",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*"} 1.774565821967978e+09
windows_file_mtime_timestamp_seconds{file="C:\\Windows\\System32\\spool\\PRINTERS\\1.shd",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*.shd"} 1.774565821967978e+09
windows_file_mtime_timestamp_seconds{file="C:\\Windows\\System32\\spool\\PRINTERS\\11.spl",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*"} 1.774565821967978e+09
windows_file_mtime_timestamp_seconds{file="C:\\Windows\\System32\\spool\\PRINTERS\\11.spl",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*.spl"} 1.774565821967978e+09
windows_file_mtime_timestamp_seconds{file="C:\\Windows\\System32\\spool\\PRINTERS\\12.shd",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*"} 1.774565821967978e+09
windows_file_mtime_timestamp_seconds{file="C:\\Windows\\System32\\spool\\PRINTERS\\12.shd",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*.shd"} 1.774565821967978e+09
windows_file_mtime_timestamp_seconds{file="C:\\Windows\\System32\\spool\\PRINTERS\\test\\12.shd",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\test\\*"} 1.774565821967978e+09
# HELP windows_file_size_bytes File size
# TYPE windows_file_size_bytes gauge
windows_file_size_bytes{file="C:\\Windows\\System32\\spool\\PRINTERS\\1.shd",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*"} 0
windows_file_size_bytes{file="C:\\Windows\\System32\\spool\\PRINTERS\\1.shd",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*.shd"} 0
windows_file_size_bytes{file="C:\\Windows\\System32\\spool\\PRINTERS\\11.spl",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*"} 0
windows_file_size_bytes{file="C:\\Windows\\System32\\spool\\PRINTERS\\11.spl",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*.spl"} 0
windows_file_size_bytes{file="C:\\Windows\\System32\\spool\\PRINTERS\\12.shd",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*"} 0
windows_file_size_bytes{file="C:\\Windows\\System32\\spool\\PRINTERS\\12.shd",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\*.shd"} 0
windows_file_size_bytes{file="C:\\Windows\\System32\\spool\\PRINTERS\\test\\12.shd",pattern="C:\\Windows\\System32\\spool\\PRINTERS\\test\\*"} 0
```

jkroepke:
It might be kill every prometheus TSDB with the huge amount of different labels.. but hey.. 

Thanks!

Dominik-esb:
Yeah hight cardinality...
Best to use it only if really needed

PR #2379, opened 2026-04-01
container: fix deserialize HNS response in GetHNSEndpointStats

### LINKED ISSUE #2378: container: network metrics always report zero (v0.31.x)
### Current Behavior

Container network metrics should report actual byte/packet counters matching HNS endpoint statistics:

```
windows_container_network_receive_bytes_total{...}   33595751
windows_container_network_transmit_bytes_total{...}   766262
windows_container_network_receive_packets_total{...}  28987
windows_container_network_transmit_packets_total{...} 9394
```

### Expected Behavior

All container network metrics are permanently zero, while CPU, memory, and storage metrics work correctly:

```
windows_container_cpu_usage_seconds_total{...}                 13.703125   ← correct ✅
windows_container_memory_usage_commit_bytes{...}               154722304   ← correct ✅
windows_container_storage_read_count_normalized_total{...}     3915        ← correct ✅
windows_container_network_receive_bytes_total{...}             0           ← always zero ❌
windows_container_network_transmit_bytes_total{...}            0           ← always zero ❌
windows_container_network_receive_packets_total{...}           0           ← always zero ❌
windows_container_network_transmit_packets_total{...}          0           ← always zero ❌
```

### Steps To Reproduce

```Markdown
**1. Run a process-isolated Windows container generating traffic:**


docker run -d --restart=always --isolation process --name net-test `
    mcr.microsoft.com/windows/servercore:ltsc2022 `
    powershell -Command "while (`$true) { `$targets = @('http://www.msftconnecttest.com/connecttest.txt','http://www.bing.com','http://detectportal.firefox.com/success.txt'); foreach (`$url in `$targets) { try { `$r = Invoke-WebRequest -Uri `$url -UseBasicParsing -TimeoutSec 5; Write-Host('OUT >> GET {0} -> {1} ({2} bytes)' -f `$url, `$r.StatusCode, `$r.RawContentLength) } catch { Write-Host('OUT >> GET {0} -> ERROR: {1}' -f `$url, `$_.Exception.Message) } }; Start-Sleep -Seconds 2 }"


**2. Run the exporter as Administrator:**


.\windows_exporter.exe --collectors.enabled=container


**3. Observe all network metrics are zero:**


(Invoke-WebRequest -Uri http://localhost:9182/metrics -UseBasicParsing).Content | Select-String "container_network"


**4. Prove the data exists in HNS by calling the raw API:**


$endpoints = Get-HnsEndpoint | Where-Object { $_.SharedContainers.Count -gt 0 }

Add-Type -TypeDefinition @"
using System;
using System.Runtime.InteropServices;
public class HNS {
    [DllImport("vmcompute.dll", CharSet = CharSet.Unicode)]
    public static extern int HNSCall(string method, string path, string body, out string response);
}
"@

foreach ($ep in $endpoints) {
    $response = ""
    [HNS]::HNSCall("GET", "/endpointstats/$($ep.Id)", "", [ref]$response) | Out-Null
    Write-Host "Endpoint: $($ep.Id)"
    Write-Host "Response: $response"
}


Output shows non-zero values wrapped in the `Output` field:

Response: {"Success":true,"Output":{"BytesReceived":33595751,"BytesSent":766262,...}}
```

### Affected versions

- v0.31.0 through v0.31.5 (all versions containing PR #2136)
- Last known working version: **v0.30.9**

Reproduced on both Windows Server 2022 (Kubernetes/containerd) and Windows 11 (Docker Desktop, process isolation).

### windows_exporter logs

```shell
No errors are logged — `collector_success{collector="container"} 1`.
```

### Potential Root cause

`GetHNSEndpointStats()` in `internal/headers/hcn/hcn.go` does not unwrap the HNS response envelope.

The HNS API (`HNSCall GET /endpointstats/<id>`) returns a **wrapped response**:

```json
{"Success":true,"Output":{"BytesReceived":33595751,"BytesSent":766262,...}}
```

But the function deserializes directly into `EndpointStats`, ignoring the `Output` wrapper:

```go
// current code (buggy)
var stats EndpointStats
if err := json.Unmarshal([]byte(result), &stats); err != nil { ... }
return stats, nil
```

Go's `json.Unmarshal` silently ignores the unknown `Success`/`Output` fields and returns a zero-valued struct.

Note that `ListEndpoints()` in the same file **correctly** unwraps the identical envelope:

```go
// ListEndpoints — correct
var endpoints struct {
    Success bool                 `json:"success"`
    Output  []EndpointProperties `json:"output"`
}
```

slab-msft:

 The HNS API (HNSCall GET /endpointstats/<id>) returns a wrapped response:
 {"Success":true,"Output":{"BytesReceived":...}}

 GetHNSEndpointStats() was deserializing directly into EndpointStats,
 missing the Output wrapper. Go's json.Unmarshal silently ignored the
 unknown fields, producing a zero-valued struct. This caused all
 container network metrics to always report 0.

 The fix unwraps the response, matching the pattern already
 used by ListEndpoints() in the same file.
 
 - Fixes: https://github.com/prometheus-community/windows_exporter/issues/2378

#### Checklist

- [ ] [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [ ] The PR body has a summary to reflect any significant (and particularly user-facing) changes introduced by this PR

### DIFF
--- a/internal/headers/hcn/hcn.go
+++ b/internal/headers/hcn/hcn.go
@@ -75,13 +75,21 @@ func GetHNSEndpointStats(endpointID string) (EndpointStats, error) {
 		return EndpointStats{}, err
 	}
 
-	var stats EndpointStats
+	var response struct {
+		Success bool          `json:"Success"`
+		Error   string        `json:"Error"`
+		Output  EndpointStats `json:"Output"`
+	}
 
-	if err := json.Unmarshal([]byte(result), &stats); err != nil {
+	if err := json.Unmarshal([]byte(result), &response); err != nil {
 		return EndpointStats{}, fmt.Errorf("failed to unmarshal JSON %s: %w", result, err)
 	}
 
-	return stats, nil
+	if !response.Success {
+		return EndpointStats{}, fmt.Errorf("HNSCall failed: %s", response.Error)
+	}
+
+	return response.Output, nil
 }
 
 func hnsCall(method, path, body *uint16) (string, error) {


jkroepke:
Thanks! Since I do not have access to Azure Environment/Windows Server environment, this area is really hard to maintain.

PR #2404, opened 2026-05-20
logical_disk: append trailing backslash to drive roots for SHCreateItemFromParsingName

### LINKED ISSUE #2290: BitLocker Sub Collector Questions
### Problem Statement

I have been looking at the BitLocker collector recently and had a few questions.

Can the sample config.yaml under tools and the example_config.yml under docs be updated to include the syntax for using the BitLocker collector?  Searching online, I found a few examples that didn't work, but the one below does.  Both metrics and bitlocker_status can be in single quotes, double quotes, or no quotes and it still works.

collector:
  logical_disk:
    enabled:
      - metrics
      - bitlocker_status

When I look in a browser at what is being published by the exporter, I don't see any results for the OS drive or the root of a drive with mount points.  No C:\ or D:\ even though D:\\MOUNT1, D:\\MOUNT2, etc. do show up.  C: and D: do show up when using get-bitlockervolume, they have a VolumeStatus of FullyEncrypted, and Protection Status of On.   The Get-WmiObject command from another thread (Get-WmiObject -Namespace "Root\CIMV2\Security\MicrosoftVolumeEncryption" -Class "Win32_EncryptableVolume") returns C:\ and D:\ and other windows_logical_disk counters, like windows_logical_disk_free_bytes, include C:\, D:\, ....  Am I missing something in my config?

The metrics coming back under windows_logical_disk_bitlocker_status (decrypting, disabled, encrypting, locked, off, on, suspended, unknown, waiting_for_activation) only have a 1 by disabled even though get-bitlockervolume shows them all FullyEncrypted with a Protection Status of On.

Thanks

### Environment

- windows_exporter Version:
- Windows Server Version:

Dominik-esb:

Fixes two related defects in the `logical_disk` `bitlocker_status` sub-collector.

**1. Trailing backslash for volume root paths**
`SHCreateItemFromParsingName` needs a trailing `\` for volume roots. Without it, the Shell treats `C:` as a relative path instead of the volume root, so the `System.Volume.BitLockerProtection` property isn't populated. This caused encrypted drives to be reported as `disabled` (default value `0`), or dropped from metrics entirely. Mount points like `D:\MOUNT1` had the same issue. Fix: ensure all volume paths end with a trailing backslash.

Reference: [Microsoft — Fully Qualified vs. Relative Paths](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#fully-qualified-vs-relative-paths)

**2. Report the real COM HRESULT instead of stale `GetLastError`**
The COM wrappers in `internal/headers/shell32/shell32.go` reported the syscall's third return value (`GetLastError`), which COM methods don't set — surfacing misleading errors like `"Too many posts were made to a semaphore."` and `"The operation completed successfully."`. They also used `hr != 0`, wrongly treating success codes like `S_FALSE` (1) as failures. Fix: detect failure with `int32(hr) < 0` (`FAILED(hr)`) and report the actual HRESULT via `ole.NewError(hr)`.

- fixes [BitLocker Sub Collector Questions #2290](https://github.com/prometheus-community/windows_exporter/issues/2290)

BitLocker status metrics now correctly report encryption/protection state for drive roots (`C:\`, `D:\`) and mount points — previously missing or wrongly reported as `disabled`. Genuine failures now log the real COM HRESULT. Verified on Windows 11: reports `waiting_for_activation` for C:, matching `manage-bde -status` and `Get-BitLockerVolume`.

#### Checklist

- [DCO](https://github.com/prometheus-community/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- The PR title has a summary of the changes and the area they affect
- The PR body has a summary to reflect any significant (and particularly user-facing) changes introduced by this PR

### DIFF
--- a/internal/collector/logical_disk/logical_disk.go
+++ b/internal/collector/logical_disk/logical_disk.go
@@ -835,6 +835,15 @@ func (c *Collector) workerBitlocker(ctx context.Context, initErrCh chan<- error)
 			}
 
 			status, err := func(path string) (int, error) {
+				// SHCreateItemFromParsingName requires a trailing backslash to correctly
+				// identify volume roots (e.g. "C:\" instead of "C:") and mount point
+				// directories (e.g. "D:\MOUNT1\" instead of "D:\MOUNT1"). Without it,
+				// drive-root paths silently fail and mount-point paths return a default
+				// property value of 0 (reported as "disabled").
+				if !strings.HasSuffix(path, `\`) {
+					path += `\`
+				}
+
 				item, err := shell32.SHCreateItemFromParsingName(path)
 				if err != nil {
 					return -1, fmt.Errorf("SHCreateItemFromParsingName failed: %w", err)
--- a/internal/collector/logical_disk/logical_disk.go
+++ b/internal/collector/logical_disk/logical_disk.go
@@ -834,6 +834,11 @@ func (c *Collector) workerBitlocker(ctx context.Context, initErrCh chan<- error)
 				continue
 			}
 
+			// TEMP DIAGNOSTIC: query both the bare path and the backslashed path
+			// to determine whether the trailing backslash actually changes the
+			// result of SHCreateItemFromParsingName/GetProperty. REMOVE BEFORE MERGE.
+			c.bitlockerProbe(&pkey, path)
+
 			status, err := func(path string) (int, error) {
 				// SHCreateItemFromParsingName requires a trailing backslash to correctly
 				// identify volume roots (e.g. "C:\" instead of "C:") and mount point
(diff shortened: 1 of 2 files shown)


jkroepke:
@Dominik-esb since reporter reports that this wont help, I do not plan to merge this for now.

Dominik-esb:
@jkroepke 
I added a different fix based on you commands. 
I would ask him to check again. If it still does not fix the issue we can postpone this PR.

Dominik-esb:
The new fix is based on you idea of @jkroepke in [#2290](https://github.com/prometheus-community/windows_exporter/issues/2290#issuecomment-4552324946)

With the help of the [MS documentation](https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-shcreateitemfromparsingname) and Claude I identified following:

> **SHCreateItemFromParsingName function**
Return value
Type: HRESULT
If this function succeeds, it returns S_OK. Otherwise, it returns an HRESULT error code.


## COM HRESULT vs. GetLastError mismatch

`GetProperty` (and `SHCreateItemFromParsingName`) call COM methods via
`SyscallN`/`Call`, which return an **HRESULT** as the first value.

The bug: the error message wraps the **third** return value (`err`), which is
`GetLastError()`. COM methods **do not set the last error**, so that value is
stale/unrelated garbage left over from a previous syscall.

This produces misleading messages such as:

- `"The operation completed successfully."` (errno `0`)
- `"Too many posts were made to a semaphore."` (errno `298`)

Fix: detect failure with `int32(hr) < 0` (`FAILED(hr)`) and report the actual HRESULT via `ole.NewError(hr)`.

PR #2416, opened 2026-06-03
Update README to clarify installer behavior wrt firewall

andrechalella:
Clarified that the installer MAY create a Windows Firewall exception, instead of WILL.

Firewall exception was created by default until a couple years ago. Seemingly the README wasn't updated then, and still stated that the installer WILL create an exception. The rest of the document doesn't clear it up definitively, even though there is an example where the appropriate argument is passed -- I refer to it in my patch.


andrechalella:
Today I learned that I have to keep the repo alive.
Edit: recreated in https://github.com/prometheus-community/windows_exporter/pull/2417

PR #2428, opened 2026-06-22
registry: new collector for generating metrics from Windows registry entries

### LINKED ISSUE #2421: Collect Windows registry values as metrics
### Problem Statement

There is no way for windows_exporter to expose numeric values stored in the Windows registry as metrics.

For some Windows software, the registry is the only place certain data exists. We run several third-party applications that are configured via registry values (`MaxConnections`, `Enabled=0/1`) and that also report runtime state there (`RunningInstances=3`) — no performance counters, no WMI classes. Monitoring Windows configuration settings themselves (memory management, feature toggles) is a similar case.

None of the existing collectors can reach this data: `performancecounter` reads from PDH, which can't access ordinary registry values, and no collector offers a configurable WMI query.

### Proposed Solution

A small, config-driven `registry` collector (disabled by default) that reads user-specified `REG_DWORD`/`REG_QWORD` values and exports them as gauges, with the key path and value name as labels. Something like:

```yaml
collector:
  registry:
    keys: |-
      - key: HKLM\SOFTWARE\Vendor\App
        values: [MaxConnections, Enabled, RunningInstances]
```

```
windows_registry_value{key="HKLM\\SOFTWARE\\Vendor\\App",name="RunningInstances"} 3

windows_registry_key_success{key="HKLM\\SOFTWARE\\Vendor\\App"} 1
```

This is just a sketch — I'm very open to feedback on the config format, metric naming, labels, and scope (e.g. whether to support enumerating all numeric values under a key, or only explicitly listed ones).

I'm happy to do the work here: I have a working prototype with tests and docs, and I'd gladly adapt it to whatever direction the maintainers prefer and submit a PR.

### Additional information

I recognize new collectors are not accepted lightly and that `performancecounter` is the preferred generic mechanism — but PDH simply can't read registry data.

There is a workaround, which we use today: a PowerShell script run as a scheduled task writes `.prom` files for the `textfile` collector. It works, but the failure modes (script errors, stale files, broken tasks) are invisible to Prometheus without extra staleness alerting, and the script and task have to be deployed and maintained on every host independent from the exporter configuration.

### Acceptance Criteria

- As a user, I can configure registry keys/values and see each `REG_DWORD`/`REG_QWORD` value exposed as a gauge with identifying labels.
- As a user, I can configure this via command-line flag or the configuration file.
- As a user, I can tell from the metrics when a configured key can't be read, without other keys being affected.
- Defaults are reasonably set: disabled by default, collects nothing until configured.
- New settings are appropriately documented (`docs/collector.registry.md` and the README collector table).

dluxem:
Adds a new `registry` collector that exposes configured Windows registry values as named Prometheus metrics. This enables operators to monitor arbitrary `REG_DWORD`/`REG_QWORD` registry values — such as application configuration — without writing custom scripts.

The collector follows the same per-value-as-metric pattern established by the `performancecounter` collector. Each configured value is exported under its own metric name (explicit or auto-generated), with user-defined labels, help text, and type (`gauge` or `counter`). A `windows_registry_key_success` gauge is emitted per key group so scrape failures are always observable.

Configuration is via the `--collector.registry.keys` flag, which accepts a JSON or YAML array of key descriptors. Validation at startup rejects duplicate key names, duplicate value names within a key, and metric names shared across values with inconsistent `# HELP` or `# TYPE`.

Only `REG_DWORD` and `REG_QWORD` value types are supported.

- fixes #2421

- I have a [sample configuration](https://gist.github.com/dluxem/634eca72c6e4624d911b611597c9f5d1) available for testing
- Auto-generated metric names follow `windows_registry_<group>_<value>` lowercased with non-alphanumeric characters replaced by `_`. Special characters are replaced individually, not collapsed, so names like `Foo (Bar)` → `foo__bar`; the docs recommend using an explicit `metric` for such cases.
- Two values may share a `metric` name if their labels differ

- New collector: `registry` (disabled by default)
- New flag: `--collector.registry.keys`
- New metric: `windows_registry_key_success{name="..."}` — 1 if all configured values under the key were read successfully, 0 otherwise
- User-defined metrics named and typed per configuration
- New documentation: `docs/collector.registry.md`

#### Checklist


jkroepke:
Looks good to me. Copilot finds some interresting finding, e.g. if a Registry key is not present, it whole exporter would not start. 

Not sure, what the best here.

internal/collector/registry/registry.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	return errors.Join(errs...)
+}
+
+func (c *Collector) collectKey(ch chan<- prometheus.Metric, key Key) error {
+	rk, err := winregistry.OpenKey(key.hive, key.subPath, winregistry.QUERY_VALUE)
+	if err != nil {
+		return fmt.Errorf("failed to open registry key %s: %w", key.label, err)
+	}
```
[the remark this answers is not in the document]
dluxem: I changed the error handling so now all ares are of `types.ErrNoData` and added a warning log to help further diagnose missing key vs permission denied, etc.

dluxem:
@jkroepke If I understood the finding correctly, the collect would be marked as failed for a missing or inaccessible key. `Build()` does not call `winregistry.OpenKey()` so we shouldn't have that failure mode on startup.

However, failing the collector because a key is missing or inaccessible seems wrong to me, so I switched all errors to `types.ErrNoData`, then log a warning (plus the _success metrics is set to 0). I think that seems reasonable here.

Keys can be added or removed/re-added by installers or config management solutions, so I don't think that should ever be fatal or consider a collector failure if they are not there.

internal/collector/registry/registry.go
```diff
@@ -0,0 +1,384 @@
...
+	return Name
+}
+
+func (c *Collector) Close() error {
+	return nil
+}
+
+func (c *Collector) Build(logger *slog.Logger, _ *mi.Session) error {
+	c.logger = logger.With(slog.String("collector", Name))
+
```
jkroepke, line 114:
```suggestion

    c.logger.Info("file collector is in an experimental state! It may subject to change.")

```

internal/collector/registry/registry.go
```diff
@@ -0,0 +1,386 @@
...
+    c.logger.Info("file collector is in an experimental state! It may subject to change.")
```
jkroepke:
```suggestion
    c.logger.Info("registry collector is in an experimental state! It may subject to change.")
```

jkroepke:
@dluxem could you take a look at the CI?

dluxem:
> @dluxem could you take a look at the CI?

Formatting fixed 🙂

PR #2490, opened 2026-09-10
mi: add support for uint16 array (UINT16A) properties

Dominik-esb:

The hand-written `mi` library could not read array-valued WMI properties. `Instance.GetElement` passed a single machine word (8 bytes) as the `MI_Value` out-parameter, but `MI_Value` is a union whose largest members (`MI_Datetime` and the array descriptors) are larger than that. When a queried element was an array, the provider wrote past the buffer and corrupted the stack, which manifested as a hang or scrape timeout.

This PR:

- Passes a full-width buffer to `MI_Instance_GetElement` so the provider can never overflow it.
- Captures the array element count alongside the value pointer on `Element`.
- Adds `UINT16A` (`uint16[]`) handling to both `Element.GetValue` and `Session.QueryUnmarshal`, unmarshalling into a Go `[]uint16`.

This unblocks array-valued properties such as `OperationalStatus` and `ThinProvisioningAlertThresholds` on `MSFT_VirtualDisk` / `MSFT_StoragePool`, which previously had to be dropped.

*(optional, in `fixes #<issue number>` format)*:

Follow-up to the discussion in #2296: https://github.com/prometheus-community/windows_exporter/pull/2296#issuecomment-3736584632

- Root cause was the undersized value buffer, not just a missing type case; the buffer fix is what prevents the hang described in #2296.
- Only scalar-word semantics are reused for existing types (word 0), so existing scalar/string queries are unaffected.
- Tests use `Win32_DiskDrive.Capabilities` (a reliably-populated `uint16[]` in `root/cimv2`), which runs on the fast provider and avoids the Storage-provider latency seen in #2296.
- Verified end-to-end on a Storage Spaces Direct cluster: `MSFT_StoragePool.OperationalStatus` now unmarshals correctly (`{2}` = OK) with no hang or timeout.

None directly. This is an internal `mi` capability that enables future collectors to expose `uint16[]` properties.

#### Checklist

- [x] DCO signed

### DIFF
--- a/internal/mi/instance.go
+++ b/internal/mi/instance.go
@@ -98,13 +98,14 @@ func (instance *Instance) GetElement(elementName string) (*Element, error) {
 		return nil, fmt.Errorf("failed to convert element name %s to UTF-16: %w", elementName, err)
 	}
 
-	// MI_Value is a union whose largest members (MI_Datetime, the array
-	// descriptors) exceed a single machine word. Provide a full-width buffer so
-	// the provider cannot write past it and corrupt the stack when the element
-	// is an array or datetime. Word 0 holds the scalar/pointer/array-data; word
-	// 1 holds the array element count for array types.
+	// MI_Value is a union sized to its largest member. On 64-bit MI_Datetime is
+	// 36 bytes and the union rounds up to 40 bytes for 8-byte alignment, so a
+	// full MI_Value must be provided; the previous single-word buffer let the
+	// provider write past it and corrupt the stack for array or datetime
+	// elements. Word 0 holds the scalar/pointer/array-data; word 1 holds the
+	// array element count for array types.
 	var (
-		valueBuf  [4]uint64
+		valueBuf  [5]uint64
 		valueType ValueType
 	)
 
--- a/internal/mi/instance.go
+++ b/internal/mi/instance.go
@@ -98,8 +98,13 @@ func (instance *Instance) GetElement(elementName string) (*Element, error) {
 		return nil, fmt.Errorf("failed to convert element name %s to UTF-16: %w", elementName, err)
 	}
 
+	// MI_Value is a union whose largest members (MI_Datetime, the array
+	// descriptors) exceed a single machine word. Provide a full-width buffer so
+	// the provider cannot write past it and corrupt the stack when the element
+	// is an array or datetime. Word 0 holds the scalar/pointer/array-data; word
+	// 1 holds the array element count for array types.
 	var (
-		value     uintptr
+		valueBuf  [4]uint64
 		valueType ValueType
 	)
 
@@ -107,7 +112,7 @@ func (instance *Instance) GetElement(elementName string) (*Element, error) {
 		instance.ft.GetElement,
 		uintptr(unsafe.Pointer(instance)),
 		uintptr(unsafe.Pointer(elementNameUTF16)),
-		uintptr(unsafe.Pointer(&value)),
+		uintptr(unsafe.Pointer(&valueBuf)),
 		uintptr(unsafe.Pointer(&valueType)),
 		0,
 		0,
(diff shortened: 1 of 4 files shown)


Dominik-esb:
Working example:

```
# HELP windows_mscluster_storagepool_health_status Health status of the storage pool. 0: Healthy, 1: Warning, 2: Unhealthy, 5: Unknown
# TYPE windows_mscluster_storagepool_health_status gauge
windows_mscluster_storagepool_health_status{name="XXXX010006-S2D",unique_id="{b2910a3a-5eab-48cd-9e48-f0d2f05cfbbc}"} 0
windows_mscluster_storagepool_health_status{name="Primordial",unique_id="{c0e1555e-13be-4acb-adcb-c8a3d74ab61d}"} 0
windows_mscluster_storagepool_health_status{name="Primordial",unique_id="{e254e391-001d-11f0-b4b7-806e6f6e6963}"} 0
```

Using [MSFT_StoragePool](https://learn.microsoft.com/en-us/windows-hardware/drivers/storage/msft-storagepool)

```
OperationalStatus

Data type: UInt16 array

Access type: Read-only

Qualifiers: [Required](https://learn.microsoft.com/en-us/windows/win32/wmisdk/standard-qualifiers)

The operational status of the storage pool. Unlike HealthStatus, this property indicates the status of hardware, software, and infrastructure issues related to the storage pool, and can contain multiple values.
```

internal/mi/instance.go
```diff
@@ -98,16 +98,21 @@ func (instance *Instance) GetElement(elementName string) (*Element, error) {
...
 	}
 
+	// MI_Value is a union whose largest members (MI_Datetime, the array
+	// descriptors) exceed a single machine word. Provide a full-width buffer so
+	// the provider cannot write past it and corrupt the stack when the element
+	// is an array or datetime. Word 0 holds the scalar/pointer/array-data; word
+	// 1 holds the array element count for array types.
 	var (
-		value     uintptr
+		valueBuf  [4]uint64
```
jkroepke: Quick review: AI tells me that `valueBuf [4]uint64` is 32 bytes, but the MI header’s MI_Datetime is 36 bytes, and MI_Value rounds to 40 bytes with 8-byte alignment. 

Have to take an eye here.

Dominik-esb: I'll check it

internal/mi/session.go
```diff
@@ -293,6 +293,10 @@ func (s *Session) QueryUnmarshal(dst any,
 				field.SetString(stringValue)
 			case ValueTypeREAL32, ValueTypeREAL64:
 				field.SetFloat(float64(element.value))
+			case ValueTypeUINT16A:
+				if field.Kind() == reflect.Slice {
+					field.Set(reflect.ValueOf(element.getUint16Array()))
+				}
```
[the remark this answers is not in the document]
jkroepke: Even `reflect.TypeOf` could be replaced by `reflect.TypeFor[[]uint16]`

internal/mi/mi_test.go
```diff
@@ -332,3 +338,79 @@ func Test_MI_QueryTimeout(t *testing.T) {
...
+	require.NotEmpty(t, element)
+
+	value, err := element.GetValue()
+	require.NoError(t, err)
+
+	capabilities, ok := value.([]uint16)
+	require.True(t, ok, "Capabilities should unmarshal to []uint16")
+	require.NotEmpty(t, capabilities)
+
+	err = operation.Close()
```
jkroepke, line 376: Seems like, if the OS has more Disks, (gh runner have 2 Disks), then all results must be enumerated/drain first.

internal/mi/instance.go
Dominik-esb: Had a quick check with AI. I missed the 8bytes. 

The fix looks good to me now.

PR #2493, opened 2026-09-10
hyperv: add host sub-collector for CPU allocation ratio

Dominik-esb:

There is currently no way to determine the CPU allocation of a Hyper-V host from the exported metrics, because no metric exposes a virtual machine's configured processor count.

The usual workaround is to count distinct `core` label values on the virtual processor series:

```
count(count by (vm, core) (windows_hyperv_hypervisor_virtual_processor_time_total{cluster=~"..."}))
```

This under-reports. On our clusters, 2–6 VMs are consistently missing from that count, so the resulting ratio does not match what Hyper-V itself reports.

This PR adds a `host` sub-collector that exposes the processor counts directly, per VM and in total, along with the ratio of virtual cores to logical host cores.

None. Supersedes the draft in #2288, which was opened from a branch that no longer applies cleanly.

Processor counts come from the Host Compute Service API rather than WMI, following your suggestion on #2288. The existing container collector uses the same `internal/headers/hcs` package. 

Verified on a Windows Server Hyper-V host: both virtual machines are enumerated, their processor counts are reported correctly, and the ratio matches.

Adds a `host` sub-collector.

| Metric | Type | Labels |
|---|---|---|
| `windows_hyperv_host_logical_processor_count` | gauge | None |
| `windows_hyperv_vm_processor_count` | gauge | `vm_id`, `vm` |
| `windows_hyperv_total_vm_processor_count` | gauge | None |
| `windows_hyperv_host_cpu_ratio` | gauge | None |

#### Checklist

- [x] DCO signed

### DIFF
--- a/internal/collector/hyperv/hyperv_host.go
+++ b/internal/collector/hyperv/hyperv_host.go
@@ -1,0 +1,278 @@
+// SPDX-License-Identifier: Apache-2.0
+//
+// Copyright The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build windows
+
+package hyperv
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"log/slog"
+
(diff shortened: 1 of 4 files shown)


internal/collector/hyperv/hyperv_host.go
```diff
@@ -0,0 +1,278 @@
...
+	if logicalCoreCount > 0 {
+		ratio = totalVirtualCoreCount / logicalCoreCount
+	}
+
+	ch <- prometheus.MustNewConstMetric(
+		c.hostCPURatio,
+		prometheus.GaugeValue,
+		ratio,
+	)
+
```
jkroepke, line 167: I see not the value of a computed value, if it can be done by prometheus.

Dominik-esb: True makes sense. 

Let me remove it. Its even better for the handling about hyper threading cores.

PR #2494, opened 2026-09-10
os: add WMI health metric

Dominik-esb:

WMI can stop responding while the host and the exporter otherwise keep running. 

This is a recurring problem on Hyper-V cluster nodes. When WMI stops responding there, Failover Cluster Manager generally cannot connect to the host either, and recovering usually requires a reboot.

This PR adds `windows_os_wmi_health`, a gauge that reports `1` when a minimal `SELECT CSName FROM Win32_OperatingSystem` query succeeds and `0` when it fails.

None.

#### Special notes for reviewer

This check may be somewhat specific to how we operate our Hyper-V clusters, but I think other Hyper-V admins would benefit from it too.

Adds `windows_os_wmi_health` (gauge, no labels). 

#### Checklist

- [x] DCO signed


Dominik-esb:
Maybe there is a better name for the metrics. It could happen that a different WMI namespace then `root/cimv2` fails but the metric wrongly promises WMI healthy.
This check specifically reports issues for Hyperv WMI issues.

Maybe its also better placed in Hyperv collector?

PR #2495, opened 2026-09-20
docs: add textfile freshness examples

noqt:
Adds the textfile examples the collector page currently requests: a last-success timestamp, queries that distinguish parse errors from successfully parsed but stale output, alerts for stale and missing producer data, and same-directory temporary-file publication so a scrape does not observe a partial payload.

This addresses the failure mode described in #2421 without changing collector behavior. The stale-output distinction was independently exercised in Lumi Eggcracker's Linux monitoring guide: https://github.com/noqt/Lumi-Eggcracker/blob/f35699de1de03564c753f088d569ce1297a2cf8c/docs/monitoring.md#collect-once . Eggcracker is Linux-only; that link is provenance for the monitoring pattern, not a claim of Windows or windows_exporter support.

Testing: documentation-only change; render/read back the Markdown and run any repository documentation checks available without metered credentials.


jkroepke:
Do you think, https://github.com/guhuajun/prometheus_posh_client would help as rederence?

